PR #3, opened 2017-08-22
Dockerfile with Python client and dependencies

JorritvandenBerg:
Using this file, the client can be set up more easily. Having Docker installed, it is just "docker build . -t neo-python" from the main directory of the repository and "docker run -it neo-python" to start the client.


localhuman:
Hi JorritvandenBerg, 

Docker file looks good!  Could you perhaps make a pull request against the development branch instead of master?

Going forward we will have all PRS go through development.

Thanks!

PR #14, opened 2017-09-22
Fixed bug in Transaction_GetReferences, case issues and missing return

hal0x2328:
Uncovered these issues when attempting to execute the mintTokens operation in the ICO template code from the neo-project csharp-examples.


localhuman:
Thanks! 👍

PR #15, opened 2017-09-30
Fix for invoke script parameter passing

hal0x2328:
Made changes to the stack packing operations to replicate neo-gui invoke script output


hal0x2328:
Just found EmitPushBigInteger, will refactor and resubmit

localhuman:
I think there could be different solution to your problem. 

Could you send me a script hash you're trying to invoke, and the params you're trying to invoke with?  That way I can test :)

localhuman:
Alright, I'm looking at the ICO_Template right now, and it looks like the 2nd param is a variable length array/object of more params, in which case your solution makes sense.

However, this approach would break invokes for any invoke that doesn't have variable length params objects.

I can work on a fix for the Invoke method which will accept that type of argument set...

PR #18, opened 2017-10-03
put array length and PACK at end of array

hal0x2328:
neo-gui puts the array length on the end of the array, followed by a PACK opcode. Making this change allows the VM to properly import the array values into the stack for the contract execution.


localhuman:
👍

PR #19, opened 2017-10-04
Fix ReadFloat and ReadDouble in BinaryReader

liato:
ReadFloat should read 4 byte and ReadDouble should read 8 byte.


localhuman:
Looks good, thanks!

PR #20, opened 2017-10-04
reverse order an array object if found inside of params

hal0x2328:
Also needed to reverse any array item when found during iteration of params


localhuman:
👍

PR #29, opened 2017-10-12
completed implementation of take() builtin

hal0x2328:
First foray into trying to implement builtins into compiler, started with take() since it was already fleshed out in builtins.py. Test case seems to work:
`
neo> build boa/tests/src/TakeTest.py test 05 05 False
Saved output to boa/tests/src/TakeTest.avm 

Calling boa/tests/src/TakeTest.py with arguments [] 
Test deploy invoke successful
Used total of 38 operations 
Result b'hello' 
Invoke TX gas cost: 0.001 
`

### DIFF
--- a/boa/code/builtins.py
+++ b/boa/code/builtins.py
@@ -108,15 +108,14 @@ def concat(str1, str2):
 #    pass
 
 
-# @TODO this currently does not work
-#def take(source, count):
-#    """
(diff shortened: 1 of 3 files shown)


localhuman:
Looks good, though I'll have to test is out when I get home from work.

The same functionality can be done with slice notation, so currently `str[:5]` should return the same thing.

PR #34, opened 2017-10-21
Prompt history file persistence, purged password from history during send, build instructions update

hal0x2328:
Changed the prompt to persist the history between sessions using a FileHistory instead of InMemoryHistory, at the same time removing the storage of the password to history during a send. 

Also updated the readme to include updated instructions for overcoming pycrypto module build issues using Python 3.5 on OSX.

Also re-added opcode debugging commits from earlier PR that was auto-closed by the above updates.


localhuman:
This is great.  Thank you!

PR #36, opened 2017-10-30
prompt.py: Allow quitting with CTRL-D. Before it threw an exception.

metachris:
When in the prompt users press CTRL + D, it threw an Exception before. Added an exception handler to quit the prompt.

This solution is recommended in the prompt-toolkit example here: https://github.com/jonathanslenders/python-prompt-toolkit/tree/master/examples/tutorial


localhuman:
👍

PR #37, opened 2017-10-30
info about testnet or mainnet in bottom toolbar

metachris:
Add info about whether we are in TestNet or MainNet to the bottom status bar.


localhuman:
Great!  
One thing that would be nice in addition to this is adding the ability to start the prompt with your preferred net, ie:
`python prompt.py --config=protocol.testnet.json`

PR #39, opened 2017-10-30
implemented multi-sig capability

hal0x2328:
This pull request implements multi-signature capability that is compatible with the latest neo-gui, i.e. you can begin a multi-sig transaction in either neo-python or neo-gui, then import and sign/relay the resulting JSON-serialized transaction in either client. 

This should allow people with private networks to claim all the NEO and GAS from the initial contract without having to use Windows at all.


localhuman:
This looks great!

PR #45, opened 2017-11-02
Tests for UInt160, UInt256, minor bugfix and cleanup

metachris:
* Full test coverage for UInt*
* Bugfix (minor) for UInt256
* Moved comparing methods from UInt160 to UIntBase because they work with all UIntBase implementations anyway


localhuman:
Once again, great work!

PR #47, opened 2017-11-02
fix for multi-sig context returning complete after one signature

hal0x2328:
The added check for self.isMultiSig under ContractParametersContext broke the privnet-claimall script since I wasn't passing isMultiSig=True when creating the context. This fix addresses that and also an issue finding the module paths when running the script from inside the contrib directory.


metachris:
The code works! It still throws an Error though. I think i've solved it, and made a PR against hal's branch: https://github.com/hal0x2328/neo-python/pull/1

metachris:
Can confirm that this works.

👍

PR #48, opened 2017-11-02
Settings: more robust network detection

metachris:


metachris:
👍 

Can confirm that this works.

PR #51, opened 2017-11-02
fix #50, make theme config backward compatible

### LINKED ISSUE #50: PR #49 is not backward compatible
PR #49 is nice but not backward compatible. It requires `ApplicationConfiguration.themes` in the configuration json file. If it is not existing it fails with: 

```
  File "prompt.py", line 823, in <module>
    settings.setup(args.config)
  File "/Users/chris/Projects/chris/neo/neo-python/neo/Settings.py", line 79, in setup
    self.token_style = config['themes'][config['theme']]
KeyError: 'themes'
```

I would suggest to check if this key exists and if not use the old default.

metachris:
Makes the theme config reading from json backward compatible

### DIFF
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -57,6 +57,7 @@ class SettingsHolder:
         with open(config_file) as data_file:
             data = json.load(data_file)
 
+        self.config_file = config_file
         config = data['ProtocolConfiguration']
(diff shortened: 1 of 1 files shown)


localhuman:
Fixed in 8da5e2c

PR #53, opened 2017-11-03
Added protocol.privnet.json

metachris:
Added a `protocol.privnet.json`. Not sure if you want, but I guessed its useful to possibly a large enough part of the user-base to warrant a third file besides mainnet and testnet configs. Any thoughts?


localhuman:
👍

PR #55, opened 2017-11-04
Apply theme to right config when specifying both -c and -t

ixje:
When specifying both `-c`  and `-t` the theming currently gets applied to the pre-loaded config.

Apply the theme after the specified config has been loaded instead of before

manual testing

**Did you add any tests?**
no

yes - autopep8

no

`development`


localhuman:
I think I agree with @metachris, would be easier to maintain a single `theme.json` file

ixje:
How about a `settings.json` or `configuration.json` just to keep it generic and open for any unforeseen configurations we might want to store in there?

localhuman:
those two names are a bit misleading, leading one to think those files would alter how the node functions.  I'd prefer something like `preferences.json`

metachris:
+1 for `preferences.json`

ixje:
Shall we then also move it completely out of Settings.py and into a
Preferences.py? As the former seems to be dedicated to node functionality.

neo/Settings.py
```diff
@@ -78,15 +78,10 @@ def setup(self, config_file):
...
-        try:
-            self.token_style = config['themes'][config['theme']]
-        except Exception as e:
-            self.token_style = {
-                "Command": "#ff0066",
-                "Default": "#00ee00",
-                "Neo": "#0000ee",
-                "Number": "#ffffff"
-            }
+        with open('preferences.json') as data_file:
```
localhuman: for reverse compatibility, could we add a check in the case that this file isn't located or is named something different?

metachris: Just `if os.path.exists(fn):` would do the trick.

ixje: I believe there is a little bit more to it than just checking if the file exists. If we're going to question if the file is there, then I think we should also question if the data inside is valid. Note that these checks also do not exist for non-existing protocol loading. Either way I'm looking into it.

metachris: :+1:

Good idea to load the JSON and if it fails to give a warning or smth. If the keys are not found it should just use the default imo.

Since the existing protocol.json files don't change i guess they are safe as they are.

ixje: the files itself might be safe yes, but `prompt` will throw up. Just try `python prompt.py -c nope`

ixje:
Current behaviour is:
* `preferences.json` is found but has invalid json content -> `sys.exit` to allow the user to make changes (e.g. when they accidentally removed a `"` or similar). If they completely borked the file or have thrown it away they'll just have to download it again. I don't think we should be keeping shadow copies.
* `preferences.json` is found but is missing the `theme` or `themes` key -> restore defaults and `sys.exit`. User has to run the command again. Given the way `Settings.py` is bootstrapped this to me seemed like the easiest and still reasonable approach. Many programs do some recovery and then ask you to try again.

PR #57, opened 2017-11-05
cli arg --version, added to Travis CI tests

metachris:
* Added cli argument `--version`
* Added running `python prompt.py --version` to TravisCI, to catch errors that show up if imports are missing or broken.


localhuman:
👍

PR #66, opened 2017-11-06
NeoNode bugfix: maximum recursion depth (fixes #59)

### LINKED ISSUE #59: Max Recursion depth issue
`Could not check message data maximum recursion depth exceeded while calling a Python object `

Seems like a race condition that happens when a node has been running for some time.

This bug has a _10 NEO_ bounty to anyone who resolves it!

metachris:
Do not merge yet, long-running tests are currently in progress.

Edit: Several instances of this code (on mainnet and testnet) have been running for about 8 hours now without problems. I'll leave it running overnight and report back tomorrow morning.

It may be easier to get an overview of the changes by looking at the updated `CheckDataReceived` method: https://github.com/metachris/neo-python/blob/2d5023732ef219f6ecd30b50e4f1d2e5f22b5db8/neo/Network/NeoNode.py#L107

NeoNode ran into an issue which caused the maximum recursion depth do be reached (`Could not check message data maximum recursion depth exceeded while calling a Python object`). See #59 

The reason was a ping-pong style recursion buildup between `NeoNode.CheckDataReceived` and `NeoNode.CheckMessageData`. This commit reduced the complexity of the code and avoids this recursion buildup.

Currently being tested, by leaving prompt.py run on several machines. One is syncing the whole mainnet from scratch, the other one just normal operations. Final results should be available in a few hours.

**Did you add any tests?**

No

Yes

No

development

### DIFF
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -164,7 +164,7 @@ class NeoNode(Protocol):
 
         # Finally, after a message has been fully deserialized and propagated,
         # check if another message can be extracted with the current buffer:
-        if len(self.buffer_in) > 24:
+        if len(self.buffer_in) >= 24:
             self.CheckDataReceived()
 
     def MessageReceived(self, m):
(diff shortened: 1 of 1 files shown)


metachris:
**Seems to work!** Several instances have now been running (both on TestNet and MainNet) for about 20 hours.  Before the problem always occurred when running overnight, so this timerange should cover it.

I'm now positive that this PR solves the issue.

PR #68, opened 2017-11-06
Generic and robust UserPreferences

metachris:
Separated UserPreferences from Settings and made user-specific overrides easy and generic

Preferences were mangled with Settings, not easily extensible and had bugs with missing prefs json file.

* Using defaults, only overrides are stored in JSON file
* Graceful fallbacks
* Easily extensible for other preferences
* Separate from Settings
* Common fixed filenames are defined in Settings
* 100% test coverage

Tests, both unit and manual

**Did you add any tests?**

Yes

yes

No. Just that new user preferences can easily be included into `UserPreferences.py`

development


localhuman:
This is great, if there was an emoji for user preferences I would use it!

PR #69, opened 2017-11-06
fixture download: improved error handling

metachris:

When the fixture download failed, the partially downloaded fixture file is left over, but the tests cannot run on them and we got many confusing error messages from the tests.

This commit just aborts running the tests if there is a defect fixtures file, and gives a hint on how you might solve it.

Raising an Exception when the fixtures file is corrupt.

Manual tests

**Did you add any tests?**

No

Yes

No

development


localhuman:
👍

PR #72, opened 2017-11-07
Settings: fix relative path references by making them absolute

metachris:

When importing neo from outside the root directory, some imports complained that the settings JSON files could not be found.

Making them absolute

Manual tests

**Did you add any tests?**

No

Yes

No

development


localhuman:
👍

PR #73, opened 2017-11-08
NEP2 Support for KeyPair (fixes #65)

### LINKED ISSUE #65: [Feature] Implement NEP 2 for wallet
Implement NEP2 standard for passphrase protection outlined here:

https://github.com/neo-project/proposals/blob/master/nep-2.mediawiki

metachris:

Added NEP2 encrypted key support to KeyPair. See #65

Adding NEP2 encrypted key support to KeyPair. @ixje helped with testing and finalizing the solution.

Tests

**Did you add any tests?**

Yes

Yes

No

development

### DIFF
--- a/neo/Cryptography/Helper.py
+++ b/neo/Cryptography/Helper.py
@@ -342,3 +342,13 @@ def base256_encode(n, minwidth=0):  # int/long to byte array
     b.reverse()
 
     return b
+
+
+def xor_bytes(a, b):
+    """ XOR on two bytes objects """
+    assert isinstance(a, bytes)
+    assert isinstance(b, bytes)
+    res = bytearray()
+    for i in range(len(a)):
+        res.append(a[i] ^ b[i])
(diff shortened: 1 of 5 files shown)


metachris:
Good point about the docstrings @ixje, thx! I'll add more to this PR.

metachris:
@ixje Added some docstrings, formatted like this: http://www.sphinx-doc.org/en/stable/ext/example_google.html

localhuman:
Great work @metachris @ixje! 

My only feedback would be to prefer [rST doc format](https://www.python.org/dev/peps/pep-0287/), mostly to keep it in line with the docs for [neo-boa](http://neo-boa.readthedocs.io/en/latest/_modules/boa/compiler.html#Compiler)

localhuman:
I will merge this now so we can merge in the `feature-claim` changes.  Though we'll have to address our choice of docstring format :)

Issue #63, opened 2017-11-06
[Feature] Add docs


localhuman:
Add actual documentation to the project.

localhuman:
- Add in docstrings to all methods/modules etc
- Fill in docstrings with more than boilerplate
- Create sphinx documentation
- Publish to [readthedocs.io](url)

ixje:
Any good document that shows how the docstrings should look for sphinx compatibility? I find there website a bit messy. Perhaps some samples how to document function parameters, return type would be nice.

metachris:
Here is an example:

* Docstrings: https://github.com/metachris/logzero/blob/master/logzero/__init__.py#L379
* Generated docs: https://logzero.readthedocs.io/en/latest/#i-logzero-logfile

ixje:
I started converting #74 to [rST format](https://www.python.org/dev/peps/pep-0287/) as requested by @localhuman [here](https://github.com/CityOfZion/neo-python/pull/73#issuecomment-342985048). Unfortunately also that standard leaves room for interpretation / free will. Before having to edit everything a 3rd time I'd like to get to an agreement on how to document the following:

- multiple return types
- exceptions
- notes

### multiple return types

Example with current [Sphinx Napoleon format](https://pypi.python.org/pypi/sphinxcontrib-napoleon):
```
"""
Change the password used to protect the private key.

    Args:
        password_old (str): the current password used to encrypt the private key.
        password_new (str): the new to be used password to encrypt the private key.

    Returns:
        bool: False, if the old password does not match the current saved password.
        None: If successfully changed the password.

"""

```

**Attempt 1**:
```
        :return: False, if the old password does not match the current saved password. None, If succesfully changed the password.
        :rtype: bool, None
```

**Attempt 2** (suggested [here](https://stackoverflow.com/a/39761518/3869515)):
Note: this drops the `:rtype:`
```
    :return: 
        bool False if the old password does not match the current saved password. 
        None None If successfully changed the password.
```

### Exception documentation
Example:
```
        """
        Retrieve the script_hash from an address.

        Args:
            address (str): a base58 encoded address.

        Raises:
            ValuesError: if an invalid address is supplied or the coin version is incorrect.
            Exception: if the address checksum fails.

        Returns:
            UInt160: script hash

        """
```

**PyCharm format:**
> :raises: :exc:`ExceptionType`
```
   :raises: :exc: `ValueError` if an invalid address is supplied or the coin version is incorrect.
[...]

PR #74, opened 2017-11-08
Document and cleanup Wallet.py

ixje:
Contributes to addressing #63 

write write write

-

**Did you add any tests?**
-

yes

just a small update to UserWallet.CreateKey / Wallet.Createkey()  nothing that breaks others calling it

development


ixje:
@localhuman I believe you made most commits in the past in this file, so I think you'd be best suited to check if the docsstrings make sense/are correct. I tried to verify the arguments and return types as much as possible and add a sensible description.

localhuman:
Looks like the test build failed.  Also please refer to this comment for doc string format https://github.com/CityOfZion/neo-python/pull/73#issuecomment-342985048

metachris:
fyi there is no `void` in Python, methods return `None` by default.

ixje:
Thanks for pointing out. I wasn't sure what to do with that and you guys were all asleep / at work when I was writing. I have to update it to the rST format anyway, so I'll fix that as well.

ixje:
@localhuman as discussed

Functions that require a description:
FindCoinsByVins
FindUnspentCoins
FindUnspentCoinsByAsset
FindUnspentCoinsByAssetAndTotal
ProcessBlocks
ProcessNewBlock
Rebuild
MakeTransaction

Functions that have an argument/return type I'm not 100% sure about (see the TODO in the docstring):
DeleteAddress
Sign
SaveTransaction

neo/Wallets/Wallet.py
```diff
@@ -122,38 +121,76 @@ def __init__(self, path, passwordKey, create):
...
     def AddWatchOnly(self, script_hash):
+        """
+        Add a watch only address to the wallet.
 
+        Args:
+            script_hash (UInt160): a bytearray (len 20) representing the public key.
+
+        Note:
+            Prints a warning to the console if the address already exists in the wallet.
+
```
metachris: Could you please remove the empty line at the end of each docstring?

neo/Wallets/Wallet.py
```diff
@@ -165,42 +202,124 @@ def ChangePassword(self, password_old, password_new):
...
+        Determine if the wallet contains the address.
+
+        Args:
+            address (str): a string representing the public key.
+
+        Returns:
+            bool: True, if the address is present in the wallet. False otherwise.
+
+        """
+
```
metachris: Also I'd say don't add empty lines after the docstring, start with the code straight away.

neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -87,21 +84,14 @@ def Create(path, password):
         return wallet
 
     def CreateKey(self, prikey=None):
-        if prikey:
-            private_key = prikey
-        else:
-            private_key = bytes(Random.get_random_bytes(32))
-
-        account = WalletKeyPair(priv_key=private_key)
+        account = super(UserWallet, self).CreateKey(private_key=prikey)
```
metachris: I'm not 100% sure about this. What if the `Wallet.CreateKey` method changes to do something more? You could just simplify the block on top with `private_key = prikey or bytes(Random.get_random_bytes(32))`

metachris:
Good job, a lot of initial much needed docstrings! I've just got a few minor suggestions (see above): (no newline at end of docstring and after docstring, and one possible code issue). 👍

neo/Implementations/Wallets/peewee/UserWallet.py
ixje: my changes actually make it match the original `neo-project` more closely
See: https://github.com/neo-project/neo/blob/1bbebb1db8fa7e5a72f372e933f65f4815a1911b/neo/Implementations/Wallets/EntityFramework/UserWallet.cs#L102

metachris: Thanks, good to have this context!

localhuman:
Great work!  Will work on filling in what is missing-

PR #78, opened 2017-11-09
NEP2 import and export in prompt

metachris:

Importing and exporting NEP2 keys from prompt.py

Adding `import nep2 {key}` and `export nep2 {address}`.

Manual testing

**Did you add any tests?**

No

Yes

I fixed a small thing in `export wif`, so that it requires an address. It would export the wif for all addresses, but without information for which address this is. In `help` it's stated that `export wif {address}` needs an address and probably that's good default behaviour.

Development

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -81,11 +81,13 @@ class PromptInterface(object):
                 'config log {on/off}',
                 'build {path/to/file.py} (test {params} {returntype} {needs_storage} {test_params})',
                 'import wif {wif}',
+                'import nep2 {nep2_encrypted_key}',
                 'import contract {path/to/file.avm} {params} {returntype} {needs_storage}',
                 'import contract_addr {contract_hash} {pubkey}',
                 'import watch_addr {address}',
                 'import token {token_contract_hash}',
                 'export wif {address}',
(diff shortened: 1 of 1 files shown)


prompt.py
```diff
@@ -250,47 +252,75 @@ def do_close_wallet(self):
     def do_import(self, arguments):
         item = get_arg(arguments)
 
-        if item:
```
metachris: It looks like much changes here, but I only put the `if not item` test on top and then could de-indent the section. perhaps easier see by looking at the current code here: https://github.com/metachris/neo-python/blob/nep2-prompt/prompt.py#L255

prompt.py
```diff
@@ -302,20 +332,49 @@ def do_export(self, arguments):
         item = get_arg(arguments)
 
         if item == 'wif':
+            if not self.Wallet:
```
metachris: Here again might be easier to see whats going on by looking at the new version here: https://github.com/metachris/neo-python/blob/nep2-prompt/prompt.py#L334

I've also re-added the check for the user-supplied `address` as stated in `help`.

localhuman:
👍

PR #81, opened 2017-11-10
Remove white spaces from docstring endings 

ixje:
as per @metachris review's request

This was still supposed to be part of #74 but my push was still pending in SourceTree.


metachris:
👍

PR #86, opened 2017-11-14
Feature events localhuman

localhuman:
- Added `execution_success` to `SmartContractEvent`, so that `Notify` events can have this context for when they are being processed
- Moved dispatch of `Notify` to occur after execution of the contract, and include the `execution_success` in the event

Notes: Ideally, when dispatching the events we'd be passing around `UInt160` or `UInt256` for tx hash or for contract hashes, that way any service consuming the events will have a flexible data type to use

### DIFF
--- a/neo/EventHub.py
+++ b/neo/EventHub.py
@@ -11,7 +11,7 @@
 #       print("handler1 called with args", args)
 #
 from collections import namedtuple
-
+import pdb
 # pymitter manages the event bus: https://github.com/riga/pymitter#examples
 from pymitter import EventEmitter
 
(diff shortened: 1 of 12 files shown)


neo/EventHub.py
```diff
@@ -31,39 +42,72 @@ class SmartContractEvent:
...
-@events.on("*")
-@events.on("*.*")
-def on_any_event(*args):
-    print("")
-    print("=EVENT: %s" % args)
-    print("")
+
+
+@events.on(SmartContractEvent.RUNTIME_LOG)
+def on_sc_log(*args):
```
metachris: if you know these events always include a SmartContractEvent as argument, we could change the method signature from this:

    def on_sc_log(*args):
        if len(args) > 0 and type(args[0]) is SmartContractEvent:
            sc_event = args[0]

to

    def on_sc_log(sc_event):

neo/SmartContract/StateMachine.py
```diff
@@ -85,6 +83,42 @@ def CheckStorageContext(self, context):
...
+        if success:
+            self.Commit()
+
+        height = Blockchain.Default().Height
+        tx_hash = engine.ScriptContainer.Hash
+
+        # dispatch all notify events, along with the success of the contract execution
+        for notify_event_args in self.notifications:
+
+            dispatch_smart_contract_event(SmartContractEvent.RUNTIME_NOTIFY, notify_event_args.State, notify_event_args.ScriptHash.ToString(), height, tx_hash.ToString(), success)
```
metachris: What data type is `notify_event_args.State` here? Would be great for usability if Runtime.Notify events always have a list of python objects as payload.

neo/VM/InteropService.py
```diff
@@ -357,5 +355,11 @@ def stack_item_to_py(stack_item):
     elif isinstance(stack_item, ByteArray):
         return stack_item.GetBigInteger()
 
+    elif isinstance(stack_item, InteropInterface):
+        return stack_item.GetInterface()
+
+    elif isinstance(stack_item, Struct):
+        return stack_item.GetArray()
```
metachris: Not sure, but should we convert the array items like for `Array` above?

    return [stack_item_to_py(item) for item in stack_item.GetArray()]

metachris:
Commented at a few points inline. Great work, love it! 👍

metachris:
PS: Also I'm all for your suggestion re "Ideally, when dispatching the events we'd be passing around UInt160 or UInt256 for tx hash or for contract hashes"

neo/VM/InteropService.py
localhuman: I haven't done any testing with `Struct` yet, but that should work

PR #89, opened 2017-11-18
Logging: configure through settings, enable (rotating) logfile

metachris:

Flexible logging in general, and smart-contract events in particular, with configuration through `neo.Settings`. Implements [logzero](https://logzero.readthedocs.io/en/latest/#example-usage) to simplify logging from anywhere in the project.

In Settings you can now:
* enable/disable logging smart contract events with `settings.log_smart_contract_events(..)`
* Setup a (rotating) logfile with `settings.logfile(..)`

Anywhere in the code you can do `from logzero import logger` and then use it like `logger.info(..)`, `logger.debug(..)`, `logger.exception(..)`, etc. If a logfile was setup through the settings, all those log statements will also be written to the logfile.

Note: with this approach, we could replace the various `print` and `self.__log` statements with `logger.xxx(..)` calls, and it would automatically be printed correctly (and potentially with color support, if we want this) and written to a previously configured logfile.

Added logging setup to neo.Settings, added SC logs to EventHub and implemented logzero.

Manual testing

**Did you add any tests?**

No

Yes

No

feature-events

### DIFF
--- a/neo/EventHub.py
+++ b/neo/EventHub.py
@@ -10,15 +10,17 @@
 #   def handler1(*args):
 #       print("handler1 called with args", args)
 #
-from collections import namedtuple
+from neo.UInt160 import UInt160
+from neo.Cryptography.Crypto import Crypto
+from neo.Settings import settings
+
+# See https://logzero.readthedocs.io/en/latest/#example-usage
+from logzero import logger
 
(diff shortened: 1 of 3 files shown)


localhuman:
This is awesome! Thanks!

PR #90, opened 2017-11-18
Feature logging

metachris:

This is a big one. Basically replaced all `print` and `__log` statements with `logger.info`, and the internal `NeoNode` and a few other very verbose and often occuring ones with `logger.debug`. Just in `neo.Prompt.*` the original print statements were left, because they are often used for user interaction.

In prompy.py you can toggle:

* debug logging from prompt with `config debug on/off`
* smart contract events with `config sc-events on/off`

Note: if you set the disable debug logging with `config debug off`, this also effects the logfile, since we are configuring and using the same logger for console and file handler.

In Settings there is now: `set_log_smart_contract_events(..)`, `set_logfile(..)` and `set_loglevel(..)` with the default loglevel set to `logging.INFO` initially.

I also tried to change all error output and exceptions to `logger.error` and `logger.exception`.

Lastly, I fixed an error that often occured for smart contracts here: https://github.com/CityOfZion/neo-python/compare/feature-events...metachris:feature-logging?expand=1#diff-af1e060838849446615e1242da849889R263

Lots of search and replace, and manual testing, and automatic testing.

The tests still work, and manual testing too.

**Did you add any tests?**

No

Yes

...

Any log statement that should always be presented to the user should use at least the INFO loglevel (eg. `logger.info(...)`). Pure debug output that the user has to first enable should use `logger.debug(..)`.

feature-events

### DIFF
--- a/neo/BigInteger.py
+++ b/neo/BigInteger.py
@@ -1,3 +1,4 @@
+from logzero import logger
 
 from neo.Cryptography.Helper import base256_encode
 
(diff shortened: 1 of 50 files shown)


localhuman:
This is, as always, a great addition!

PR #91, opened 2017-11-18
Feature events

metachris:
EventHub and logging.

Many commits, many changes. Will add some documentation to the README / docs in a later step. A prompt user doesn't really notice much, except a bit more log messages, and the option to use `config debug on/off` and `config sc-events on/off`.

For devs that use neo-python as library, this is a great many essential changes. It allows to send all the log messages into logfiles, and to easily react to smart contract events.


localhuman:
Just wanted to leave a comment here about the new way to subscribe to SC events:
```
smart_contract = SmartContract("6537b4bd100e514119e3a7ab49d520d20ef2c2a4")
 
@smart_contract.on_notify
def sc_notify(event):
  print("SmartContract Runtime.Notify event:", event)
  if len(event.event_payload):
    print(event.event_payload[0].decode("utf-8"))
```

Great work by @metachris !

PR #92, opened 2017-11-19
Cleanup in create wallet flow (error handling,add docstrings, deadcode)

ixje:
- remove some unnecessary code
- adds/updated docstrings as per #63 
- add some return value checks/decisions

see above

>  python -m unittest discover neo

**Did you add any tests?**
no

yes

no

development


neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -79,19 +79,43 @@ def Open(path, password):
...
+        Create a KeyPair and store it encrypted in the database.
+
+        Args:
+            private_key (iterable_of_ints): (optional) 32 byte private key.
+
+        Returns:
+            KeyPair: a KeyPair instance.
+        """
         account = super(UserWallet, self).CreateKey(private_key=prikey)
-        self._keys[account.PublicKeyHash.ToBytes()] = account
```
metachris: I don't think you can delete this line. `Wallet.py` uses `self._keys`

neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -102,18 +126,23 @@ def OnCreateAccount(self, account):
 
         db_account, created = Account.get_or_create(
             PrivateKeyEncrypted=encrypted_pk, PublicKeyHash=account.PublicKeyHash.ToBytes())
-        db_account.PrivateKeyEncrypted = encrypted_pk
+        if not created:
```
metachris: Could you please explain the change here? Not quite clear to me what is the advantage now.

neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -102,18 +126,23 @@ def OnCreateAccount(self, account):
...
+
+        Args:
+            contract(neo.SmartContract.Contract): a Contract instance.
+        """
         super(UserWallet, self).AddContract(contract)
 
         db_contract = None
         try:
             db_contract = Contract.get(ScriptHash=contract.ScriptHash.ToBytes())
-            db_contract.delete_instance()
```
metachris: @localhuman please feedback: has there been a reason to delete the old contract here? Not sure this change is compatible.

neo/Wallets/Wallet.py
```diff
@@ -67,9 +67,9 @@ def __init__(self, path, passwordKey, create):
         """
 
         Args:
-            path: (str) A path indicating where to create or open the wallet
-            passwordKey: (str) A password to use in creating or opening the wallet
-            create: (bool) Whether to create the wallet or simply open
+            path: (str) A path indicating where to create or open the wallet.
```
metachris: The data type comes before the colon (so `path (str):`)

prompt.py
```diff
@@ -222,11 +222,13 @@ def do_create(self, arguments):
...
                     return
 
-                contract = self.Wallet.GetDefaultContract()
-                key = self.Wallet.GetKey(contract.PublicKeyHash)
-
                 print("Wallet %s " % json.dumps(self.Wallet.ToJson(), indent=4))
-                print("pubkey %s " % key.PublicKey.encode_point(True))
+
+                contract = self.Wallet.GetDefaultContract()
+                if contract is not None:
```
metachris: Can this ever be None?

neo/Implementations/Wallets/peewee/UserWallet.py
ixje: the super class does the exact same thing from which you can access it:
https://github.com/CityOfZion/neo-python/blob/master/neo/Wallets/Wallet.py#L274

the `neo` C# actually does the same thing as well
https://github.com/neo-project/neo/blob/master/neo/Implementations/Wallets/EntityFramework/UserWallet.cs#L102-L107

neo/Implementations/Wallets/peewee/UserWallet.py
ixje: My initial change looked like
```
        db_account, created = Account.get_or_create(
            PrivateKeyEncrypted=encrypted_pk, PublicKeyHash=account.PublicKeyHash.ToBytes())
        if not created:
            db_account.PrivateKeyEncrypted = encrypted_pk
            db_account.save()
```
but then I realised that `Account.get_or_create` does not automatically save and that I had to change the indentation on `db_account.save()`. 

Now when thinking about it I believe 
```
        if not created:
            db_account.PrivateKeyEncrypted = encrypted_pk
```
should actually be removed completely, because it is always redundant. `PrivateKeyEncrypted` and `PublicKeyHash` are supplied as [filters](http://docs.peewee-orm.com/en/latest/peewee/api.html#Model.get_or_create). If it finds a record that satisfies them, then assigning the `encrypted_pk` is redundant. If no record exist that matches the filters, then it will use the already supplied filter/arguments to create a new record.

neo/Implementations/Wallets/peewee/UserWallet.py
ixje: for what it is worth, the new code matches closer to https://github.com/neo-project/neo/blob/master/neo/Implementations/Wallets/EntityFramework/UserWallet.cs#L37-L41
(also note https://github.com/neo-project/neo/blob/master/neo/Implementations/Wallets/EntityFramework/UserWallet.cs#L59 where they actually save the context regardless if it took the if or else branch, which I'm doing now as well).  Oh and the tests didn't fail! ;)

neo/Wallets/Wallet.py
ixje: /me looks @localhuman ;)
I didn't write those comments, but I can of course update them

prompt.py
ixje: Apparently so: https://github.com/CityOfZion/neo-python/blob/master/neo/Wallets/Wallet.py#L892

neo/Implementations/Wallets/peewee/UserWallet.py
metachris: Right! 👍

neo/Implementations/Wallets/peewee/UserWallet.py
metachris: Thanks, great explanation. Yeah, this line should probably just be removed.

neo/Wallets/Wallet.py
metachris: Plz 👍

prompt.py
metachris: If it can be None then we should raise an Exception / show an error here instead! 

The suggested code would make it seem that everything worked, when in fact it didn't. Before, it would throw an exception because it would access properties of None.

neo/Implementations/Wallets/peewee/UserWallet.py
ixje: will do

prompt.py
ixje: I agree, before an error would be thrown in `self.Wallet.GetKey` when supplying `None` as argument. For its purpose I believe the solution is valid until there's a proper update on how to notify `GetDefaultContract()` failure.

neo/Wallets/Wallet.py
ixje: will do

prompt.py
metachris: My preference would be to leave the old code here. It would be clear to the to the user that things broke, with this change it wouldn't. Perhaps we should just have `GetDefaultContract` raise an exception instead of returning `None`

neo/Implementations/Wallets/peewee/UserWallet.py
metachris: Thanks for the explanation. I'd still be interested if @localhuman knows of a reason this code was here in the first place, and if this change works.

prompt.py
localhuman: I think it should raise an exception if no default contract is found

neo/Implementations/Wallets/peewee/UserWallet.py
localhuman: I like deleting the existing instance so that any FKs get cleaned out, but if the new version works that's ok too!

neo/Wallets/Wallet.py
```diff
@@ -894,16 +894,21 @@ def GetDefaultContract(self):
...
         """
         try:
-            return self.GetContracts()[0]
-        except Exception as e:
-            logger.error("Could not find default contract")
-        return None
+            default = self.GetContracts()[0]
+        except Exception:
+            error_msg = "Could not find default contract"
+            logger.error(error_msg)
```
metachris: Please include the original exception, and re-raise the original exception. Like this:

    except Exception as e:
        logger.error("Could not find default contract")
        logger.exception(e)
        raise

neo/Wallets/Wallet.py
```diff
@@ -894,16 +894,21 @@ def GetDefaultContract(self):
...
 
         Note:
             Prints a warning to the console if the default contract could not be found.
         """
         try:
-            return self.GetContracts()[0]
-        except Exception as e:
-            logger.error("Could not find default contract")
-        return None
+            default = self.GetContracts()[0]
```
metachris: here you could get rid of the `default` variable and just straight `return self.GetContracts...`

prompt.py
```diff
@@ -224,11 +224,13 @@ def do_create(self, arguments):
 
                 print("Wallet %s " % json.dumps(self.Wallet.ToJson(), indent=4))
 
-                contract = self.Wallet.GetDefaultContract()
-                if contract is not None:
+                try:
+                    contract = self.Wallet.GetDefaultContract()
                     key = self.Wallet.GetKey(contract.PublicKeyHash)
                     if key is not None:
```
metachris: you could also get rid of this line (`if key is not None`). shouldn't (and hopefully can't) happen, but if it would, the user should now something fucked up

neo/Wallets/Wallet.py
ixje: yeah, I guess it's a habit to try and not have multiple returns in a function.

metachris: It's actually a good pattern to return early, when you're done early! 

Makes it easier to keep track of what's happening and what has already ended, and often times you need less indentation.

prompt.py
ixje: the `GetKey()` _can_ return None. If the function can return it, shouldn't we then not at least handle it. It's feel wrong to let it raise another exception because `print("pubkey %s " % key.PublicKey.encode_point(True))` will fail to access `PublicKey.encode_point(True)` on `None type`

prompt.py
```diff
@@ -224,11 +224,13 @@ def do_create(self, arguments):
...
                         print("pubkey %s " % key.PublicKey.encode_point(True))
+                except Exception as e:
+                    print("Exception getting default contract: %s " % e)
```
metachris: Actually, please remove the whole try..except block. If it breaks it raises an exception anyway, and if it breaks it should not continue like this to the code below.

prompt.py
metachris: True, it can return it. But then an error occured and we should not pretend to the user that things worked. Perhaps it would be better for `GetKey` to also raise an exception if it can't find the key.

prompt.py
ixje: I'm all good with stopping further execution, but same as above: why remove the exception handling? It can also at least catch it, notify instead of printing ~20 lines of stacktrace (they can consult the log file for that) and cleanup + `return`. If you don't catch it, it will leave the wallet in a messed up state i.e. it still prints the keys, address etc, but it will not sync. You can also not `wallet close` it at this point because it will throw
>  File "prompt.py", line 245, in do_close_wallet
>    self._walletdb_loop.stop()
> AttributeError: 'NoneType' object has no attribute 'stop'

metachris: > It can also at least catch it, notify instead of printing ~20 lines of stacktrace (they can consult the log file for that) and cleanup + return

Yeah, that's also ok!

neo/Implementations/Wallets/peewee/UserWallet.py
metachris: @ixje please restore the line here.

localhuman:
This PR still says 'changes requested', but it looks like that has been resolved. @metachris is this ok to merge?

metachris:
We're working on it! 3 things I would like to still improve:

* https://github.com/CityOfZion/neo-python/pull/92/files#diff-f1c26fe160bdce08834e51e99d427cefR909
* https://github.com/CityOfZion/neo-python/pull/92/files#diff-8acf03fcba903c22586f9efc6a394aa0L111
* https://github.com/CityOfZion/neo-python/pull/92/files#diff-51c9e721b63b4d4727f5d7f5423a5017R226 -- there is an ongoing discussion about handling this

Should soon be there.

.coveragerc
```diff
@@ -1,5 +1,6 @@
 [run]
-source = neo
```
metachris: please make this a separate PR

neo/Wallets/Wallet.py
```diff
@@ -888,16 +894,19 @@ def GetDefaultContract(self):
...
 
         Note:
             Prints a warning to the console if the default contract could not be found.
         """
         try:
             return self.GetContracts()[0]
-        except Exception as e:
-            logger.error("Could not find default contract")
-        return None
+        except Exception:
```
metachris: you need `except Exception as e:` for this to work

metachris: and it should re-raise, as discussed in a previous thread

metachris:
Looks good to me now 👍

localhuman:
This is a great collaboration, thanks @ixje and @metachris !

PR #105, opened 2017-11-29
Added Wallet.GetAssetAmount(asset_symbol)

metachris:

Added `Wallet.GetAssetAmount(asset_symbol)`, which returns the amount of this asset in this wallet's synced balances.

It's often needed for dapps to query whether a wallet has a certain balance, for instance `NEOGas`. 

Not sure if you want this to be part of `Wallet`, but I thought it might be a good place to start. We could also give it a different name, I started with `GetAssetAmount` because it seemed in tune with the other method names. Should we name it `GetAsset` or something else?

implemented it

Testing

**Did you add any tests?**

Yes

No


metachris:
Rather use `FindUnspentCoinsByAsset` method https://github.com/metachris/neo-python/blob/0ba208abcd2696dac174eea81a9a8f3c985e7395/neo/Wallets/Wallet.py#L407

PR #107, opened 2017-11-30
Fix Wallet.ContainsKey

ixje:
Wallet.ContainsKey was not working

Change the `unhex` parameter to match [this](https://github.com/CityOfZion/neo-python/blob/master/neo/Wallets/KeyPair.py#L81)
```self.PublicKeyHash = Crypto.ToScriptHash(self.PublicKey.encode_point(True), unhex=True)```

Because `Wallet._keys` stores that same `PublicKeyHash` (see [here](https://github.com/CityOfZion/neo-python/blob/master/neo/Wallets/Wallet.py#L278))

- wrote 2 tests
- ran `python -m unittest discover neo` and did not break anything else.

**Did you add any tests?**
yes

no

### DIFF
--- a/neo/Wallets/Wallet.py
+++ b/neo/Wallets/Wallet.py
@@ -220,7 +220,7 @@ class Wallet(object):
         Returns:
             bool: True if exists, False otherwise.
         """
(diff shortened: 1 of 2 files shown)


localhuman:
👍

PR #108, opened 2017-12-02
neo.Network.api.decorators + example with rest api

metachris:

Added a node example which integrates a REST API. Uses https://github.com/twisted/klein (official Twisted repo) which makes it possible to write Flask-like API routes and handlers, and integrates with the Twisted reactor.

This PR is meant to start a discussion and gather feedback in first place.

Added `neo.Network.api.decorators`:

* `@json_response`
* `@authenticated`
* `@catch_exceptions`

Added `examples/smart-contract-rest-api.py`

n/a

Manual tests

**Did you add any tests?**

n/a

No


localhuman:
Looks great 🥇

localhuman:
looks like the tests are looking for the no longer existing `neo.contrib.api`

metachris:
Sloppy me ^^

PR #113, opened 2017-12-10
Feature sc apis

localhuman:

Adds additional functionality covered in these features for neo core:
- Dynamic Invoke: https://github.com/neo-project/neo/pull/120
- Neo.Runtime.GetTime: https://github.com/neo-project/neo/pull/122
- Neo.Transaction.GetUnspentCoins: https://github.com/neo-project/neo/pull/123
- Neo.Header.GetIndex: https://github.com/neo-project/neo/pull/121

These work in conjunction with changes to currently in the `neo-boa` development branch

Main change is that when building a contract or importing one, you now need to specify whether a contract needs dynamic invoke or not, same as you do with storage.  So this:
```
> build ../path/to/file.py test 0510 05 True name []
```
Would now be this:
```
> build ../path/to/file.py test 0510 05 True False name []
```
Unless you want your contract to be able to invoke other contracts defined at runtime, in which case it would be this: 
```
> build ../path/to/file.py test 0510 05 True True name []
```

Same changes apply to the `import contract` command:
```
> import contract path/to/file.avm 0510 05 True
```
Should now be this:
```
> import contract path/to/file.avm 0510 05 True False
```


metachris:
👍

PR #116, opened 2017-12-11
Fix comments to show correct location for privnet-claim-neo-and-gas.py

AshRolls:
Incorrect path in comments for `privnet-claim-neo-and-gas.py`

Updated comment

Just a comment

**Did you add any tests?**
N/a

Nope


metachris:
Thanks! 👍

PR #122, opened 2017-12-19
Add warning message for invalid wallet parameter

rockacola:

Display an error message indicating invalid parameter(s) for `wallet` command.

By checking adding a catch all for `item` matching and print out a console message.

Example:

```
neo> wallet sync
wallet: 'sync' is an invalid parameter
```

Basic smoke test on prompt.py against privatenet.

**Did you add any tests?**

Not Applicable.

No.


metachris:
Looks good to me! 👍

rockacola:
Corrected the target branch to be `development` branch instead.

PR #123, opened 2017-12-20
Add VerificationContract + tests

ixje:
Add IO Helper utility needed in more classes to come


neo/Wallets/VerificationContract.py
```diff
@@ -0,0 +1,114 @@
...
+        super(VerificationContract, self).__init__(
+            redeem_script=redeem_script,
+            param_list=param_list,
+            pubkey_hash=pubkey_hash
+        )
+
+    @property
+    def IsStandard(self):
+        """
+        XXXUnknown.
```
ixje: It seems to do sanity checking on the Script, which makes me think it's more like a `IsValid` check. Eitherway, I don't know what could generate a 'non-standard'. If you have a good description please let me know

neo/Wallets/VerificationContract.py
```diff
@@ -0,0 +1,114 @@
...
+class VerificationContract(Contract):
+    def __init__(self, redeem_script=None, param_list=None, pubkey_hash=None):
+        super(VerificationContract, self).__init__(
+            redeem_script=redeem_script,
+            param_list=param_list,
+            pubkey_hash=pubkey_hash
+        )
+
+    @property
+    def IsStandard(self):
```
ixje: #overwriting Contract.IsStandard because Script should not need to be unhexlified

neo/Wallets/VerificationContract.py
```diff
@@ -0,0 +1,114 @@
...
+        if len(scp) != 35:
+            return False
+
+        if scp[0] != 33 or scp[34] != int.from_bytes(CHECKSIG, 'little'):
+            return False
+
+        return True
+
+    @property
+    def ScriptHash(self):
```
ixje: note I'm overriding this method from the `VerificationCode`  class because `self.Script` should not be a hex string. I'm being strict on data types to prevent mismatches in the future (which i've already encountered quite some).

neo/Wallets/test_VerificationContract.py
```diff
@@ -0,0 +1,82 @@
...
+from neo.Wallets.KeyPair import KeyPair
+
+
+class VerificationContractTest(NeoTestCase):
+    # Captured data from C# client 2.4.1.0
+    raw_verification_contract = binascii.unhexlify(
+        '27fd098f9ffc3309fbc310cccfde9a9fb445a8790100232102103a7f7dd016558597f7960d27c516a4394fd968b9e65155eb4b013e4040406eac')
+    expected_address = 'AWLYWXB8C9Lt1nHdDZJnC5cpYJjgRDLk17'
+    expected_isStandard = True
+    # should we make this a proper ENUM like C#?
```
ixje: Need an opinion on the ContractParameter type. It's currently a class, but I think a proper enum could do as well.

neo/Wallets/VerificationContract.py
localhuman: Any smart contract associated with an address is 'non-standard`

neo/Wallets/test_VerificationContract.py
```diff
@@ -0,0 +1,82 @@
...
+    expected_ParameterList = [bytes([ContractParameterType.Signature])]
```
ixje: Need an opinion on the ContractParameter type. It's currently a class, but I think a proper enum could do as well.

localhuman:
Does `VerificationContract` do anything different or more than `neo.SmartContract.Contract` ?

ixje:
Yes, 
`self.Script` are actual bytes instead of a hexstring encapsulted in a bytes() object. This matches matches the C# reference closer and prevents future issues with data type mismatches.
`self.ParameterList` is actually a list now instead of just a bytes object. This also required changing the serialization parts. Again reflects C# closer.
`Size` is implemented

neo/Wallets/test_VerificationContract.py
ixje: FYI; it's actually an enum in C#

localhuman:
If you plan on replacing the existing `neo.SmartContract.Contract` implementation ( Which is how the c# was originally implemented :) ) please be aware that there is a lot of complexity involved with contract invocations, multi sig contracts, etc. 
This is double the case for changing `ContractParameterType`

ixje:
I hope to not touch `neo.SmartContract.Contract`, and as far as I can see now it's not needed. I simply operate on the fact that in the past when I asked for refactoring you said you want to stay as close to the C# implementation as possible. In the end it's not my call, so if you think the current implementation is wrong and should be changed then just let me know what I should do and I'll try to update and work around it.

localhuman:
Looks good.  I just worry about having very similar functionality implemented twice, once in `VerificationContract` and the other in `Contract`.  If we can get rid of the `neo.SmartContract.Contract` implementation and follow the C# implementation closer that would be ideal.  My only hesitation is that this has implications throughout the codebase.

If possible for now, we should keep `ContractParameterType` as it is.

PR #127, opened 2017-12-21
Add Witness tests + docstring

ixje:
increase testing coverage
write test
tests?
**Did you add any tests?**
yes
yes, there is 1 test not implemented because the purpose of the exception handling in the constructor is unclear


neo/Core/Witness.py
```diff
@@ -12,37 +7,67 @@ class Witness(SerializableMixin):
...
+            invocation_script (bytearray): the invocation script
+            verification_script (bytearray): the verification script
+
+        Throws:
+            Exception: if verification_script is a string
+        """
         try:
             self.InvocationScript = binascii.unhexlify(invocation_script)
         except Exception as e:
             self.InvocationScript = invocation_script
```
ixje: What is a valid case for exception handling? If we submit a list `[0xaa,0xbb]` it catch it as expected, create the Witness instance but that will not be serializable

ixje:
test fails because it depends on the `neo.IO.Helper` update submitted in #123

neo/Core/Witness.py
localhuman: The exception it catches is a `binascii` error when trying to `unhexlify` something that is already a raw byte string like `b'\x01\x02\xAF'` etc.  This could be made to catch a `binascii` error instead of a generic exception.

neo/Core/Witness.py
```diff
@@ -12,37 +7,67 @@ class Witness(SerializableMixin):
...
 
     def Size(self):
-        return sys.getsizeof(self.InvocationScript) + sys.getsizeof(self.VerificationScript)
+        """
+        Get the total size in bytes of the object.
+
+        Returns:
+            int: size.
+        """
+        from neo.IO.Helper import Helper as IOHelper
```
metachris: don't do inline imports. import at top of files.

neo/VM/ExecutionEngine.py
```diff
@@ -747,6 +747,30 @@ def ExecuteOp(self, opcode, context):
...
+                if not arrItem.IsArray:
+                    self._VMState |= VMState.FAULT
+                    return
+
+                arr = arrItem.GetArray()
+
+                arr.append(newItem)
+
+            elif opcode == REVERSE:
+
```
metachris: no new line after if

neo/Core/Witness.py
```diff
@@ -12,37 +7,67 @@ class Witness(SerializableMixin):
     InvocationScript = None
     VerificationScript = None
 
-    def __init__(self, invocation_script=None, verification_script=None):
+    def __init__(self, invocation_script=bytearray(), verification_script=bytearray()):
```
metachris: why make invocation_script use an empty bytearray as default?

prompt.py
```diff
@@ -437,6 +437,8 @@ def show_wallet(self, arguments):
                 AddAlias(self.Wallet, arguments[1], arguments[2])
             else:
                 print("Please supply an address and title")
+        else:
```
metachris: did you merge development into you branch before pushing this PR? i think that's already in there

neo/Core/Witness.py
ixje: I'd like to but I can't in this case because it will create a 'circular import' error.
`Witness` imports `neo.IO.Helper`, which imports `Transaction` which tries to import `Witness`. 

A possible solution is to make `neo.IO.Helper` no longer a class, but make it a simple a module from which you call the methods after importing (`from neo.IO import Helper` -> `Helper.<method x>`. Otherwise I don't know how to solve it. This was the one with the way with least impact I think. Ideas are welcome

neo/VM/ExecutionEngine.py
ixje: I'd say this does not belong to this issue because this commit is from merging the development branch into this branch..

neo/Core/Witness.py
ixje: because `ToJson` and `Serialize` will fail without it when you initialise an empty `Witness`   while it should return 2 bytes representing 1) 0 length of invocation script and 2) 0 length of verification script

prompt.py
ixje: I committed my local changes, merged coz/development/ into my local branch, pushed to my own remote, PR'ed from there to here. I don't have control over keeping `coz/feature-consensus` up to date with `coz/development` and @localhuman said on slack "make sure to merge early and often from development" so that's what I did?

neo/Core/Witness.py
ixje: How about we simplify it and just force the arguments to be of type `bytes`or `bytearray`?
e.g.
```
def __init__(self, invocation_script=bytearray(), verification_script=bytearray()):
    if not isinstance(invocation_script, (bytearray, bytes)):
        raise ValueError('Invalid invocation_script parameter type: {} is not of type bytearray or bytes'.format(type(invocation_script)))

    if not isinstance(verification_script, (bytearray, bytes)):
        raise ValueError('Invalid verification_script parameter type: {} is not of type bytearray or bytes '.format(type(verification_script)))

    self.InvocationScript = invocation_script
    self.VerificationScript = verification_script
```
All existing calls to `Witness` already adhere to these data types and `Deserialize()` also already returns `bytes`. This way we don't need to abuse any exception handling for type checking.

localhuman: both `bytes` and `bytearrays` can be in hexlified and unhexlfied format, so enforcing it to be one or the other won't help.

ixje: Good point. I think instead of direct assignment we should still do the try/catch on `binascii.unhexlify` then, but also keep the instance testing. 

Why? Because a `string` is a valid argument type for `unhexlify` according to the documentation. However, it can throw the same `binascii.Error` exception when the string is invalid as an already raw bytes input. E.g.
1) `binascii.unhexlify(b'\x01\x02\xAF)` throws `binascii.Error('Odd-length string')
2) `binascii.unhexlify('aabbd')` throws the same error as 1.

The first value you still want to keep as it's actually valid but requires no unhexlifying, the second is really faulty. 

So excluding non-`bytes`/`bytearray` types would make it so that the only `binascii.Error` thrown by `unhexlify` is because of an already raw bytes value.

```
def __init__(self, invocation_script=bytearray(), verification_script=bytearray()):
        if not isinstance(invocation_script, (bytearray, bytes)):
            raise ValueError('Invalid invocation_script parameter type: {} is not of type bytearray or bytes'.format(type(invocation_script)))

        if not isinstance(verification_script, (bytearray, bytes)):
            raise ValueError('Invalid verification_script parameter type: {} is not of type bytearray or bytes '.format(type(verification_script)))

        try:
            self.InvocationScript = binascii.unhexlify(invocation_script)
        except binascii.Error:
            self.InvocationScript = invocation_script
            
        try:
            self.VerificationScript= binascii.unhexlify(verification_script)
        except binascii.Error:
            self.VerificationScript = verification_script
```

ixje:
@metachris are there any further actions you want me to take on your feedback or are my explanations acceptable?

neo/Core/Witness.py
ixje: @localhuman ping. I'd like to wrap this PR up so I can submit the next one that depends on this one :)

neo/Core/Witness.py
metachris: I see. If possible it is typically better to resolve circular important at the root instead of workarounds. Don't have much time atm to dive into it. Both ok for me I guess, although preferred resolve circular.

neo/Core/Witness.py
metachris: The question is: does it work with that default args of empty bytearray? Else better check if none and then error out.

neo/Core/Witness.py
localhuman: That looks good! 👍

localhuman:
You'll need to fix the code style for the CI to pass

neo/Core/Witness.py
ixje: Yeah it works with those default args. The `Witness` class is covered 100% in the test file and once I merge the commits from #123 it will show.

ixje:
@localhuman it should be good now. Unfortunately the circular import solution didn't work (read above at the review). tldr; kept importing inline

PR #129, opened 2017-12-23
 Fix LevelDBBlockchain.GetBlockHash, add tests + docstrings for affected methods.

ixje:
`LevelDBBlockchain.GetBlockHash()` returned invalid values on check failures causing [GetBlock()](https://github.com/CityOfZion/neo-python/blob/aff3e6ca415b45ccb4f6bff9350a60a40e54538a/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py#L508) and [GetBlockByHeight()](https://github.com/CityOfZion/neo-python/blob/aff3e6ca415b45ccb4f6bff9350a60a40e54538a/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py#L490) (who rely on the return values) to behave incorrectly.

It now matches the [C# implementation](https://github.com/neo-project/neo/blob/8e1e4f2b91e75c11d35cff3e2607a5fcbf0747ab/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.cs#L206) as well 

- It also addresses a typo in `AddHeaders()`.

Correct the return values from `bool` to `None`

wrote test cases

**Did you add any tests?**
yes

no

### DIFF
--- a/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py
+++ b/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py
@@ -37,7 +37,6 @@ from neo.BigInteger import BigInteger
 
 
 class LevelDBBlockchain(Blockchain):
-
     _path = None
     _db = None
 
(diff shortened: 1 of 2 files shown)


localhuman:
👍

PR #130, opened 2017-12-24
Consensus payload

ixje:
Add `ConsensusPayload` class + merge latest dev into `feature-consensus` branch.

add the class

write tests

**Did you add any tests?**
yes

1. I had to use patching to return a fixed list of validators for the `Blockchain.GetValidators()` call, as this function is not yet implemented.
2. @localhuman can you add this fixture file on the coz host? https://www.dropbox.com/s/bitp6s82h1i3y6d/fixtures_consensus.tar.gz?dl=0
otherwise the tests will not work.


localhuman:
Added fixture file here:

https://s3.us-east-2.amazonaws.com/cityofzion/fixtures/fixtures_consensus.tar.gz

PR #132, opened 2017-12-27
Normalize and correct some console log messages

rockacola:
Normalize and correct some console log messages.

Spell correction and minor rewording on some console log messages.

Basic smoke test on prompt.py against privatenet.

**Did you add any tests?**
Not Applicable.

No.


metachris:
:+1:

PR #136, opened 2017-12-28
Update all dependencies from requirements.txt

metachris:

Outdated dependencies

Update them

Run the tests

**Did you add any tests?**

no

no


localhuman:
🆙 📅

PR #139, opened 2017-12-29
Document remainder of /neo/Wallets/

ixje:
A step forward in addressing #63 

write docs

n/a

**Did you add any tests?**
no

No

### DIFF
--- a/neo/Wallets/NEP5Token.py
+++ b/neo/Wallets/NEP5Token.py
@@ -1,7 +1,7 @@
 import binascii
 import traceback
 
-from decimal import getcontext, Decimal
+from decimal import Decimal
 from logzero import logger
 
 from neo.Core.VerificationCode import VerificationCode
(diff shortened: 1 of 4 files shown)


neo/Wallets/NEP5Token.py
```diff
@@ -130,7 +230,20 @@ def Approve(self, wallet, owner_addr, requestor_addr, amount):
...
 
     def Mint(self, wallet, mint_to_addr, attachment_args):
-
+        """
+        Call the "mintTokens" function of the smart contract.
+
+        Args:
+            wallet (neo.Wallets.Wallet): a wallet instance.
+            mint_to_addr (str): public address of the account to mint the tokens to.
+            attachment_args: ???
```
ixje: @localhuman you probably know what this parameter type + description should be.

localhuman: should be `(str): passed from prompt, should be '--attach-neo=X' or '--attach-gas=X'`

PR #140, opened 2017-12-29
Document all of /neo/IO/

ixje:
A step forward in addressing #63 

write docstrings

n/a

**Did you add any tests?**
no

no


localhuman:
This is excellent work.

PR #143, opened 2018-01-02
Document /neo/Core/State/

ixje:
#63 

write docs

not, only added docstrings. Did not refactor anything.

**Did you add any tests?**
n/a

no


localhuman:
You are on a roll!

PR #144, opened 2018-01-02
Document /neo/Core/TX/

ixje:
#63 

write docs

not. only added docstrings. didn't refactor anything

**Did you add any tests?**
no, n/a

yes, see my comment below about a reference that was required.

### DIFF
--- a/neo/Core/TX/ClaimTransaction.py
+++ b/neo/Core/TX/ClaimTransaction.py
@@ -1,31 +1,52 @@
-import sys
-from itertools import groupby
-from logzero import logger
-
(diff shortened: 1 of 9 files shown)


ixje:
It's worth pointing out that this branch was branched from the same point as #143, yet this did not pass the building because of a missing reference while #143 did. Looking at [the code](https://github.com/CityOfZion/neo-python/blob/master/neo/Core/TX/ClaimTransaction.py#L37) it's clear there's a reference needed, so this makes me wonder how it could have ever passed tests in the past?

localhuman:
Also looks great :)

PR #145, opened 2018-01-02
fix bugs about binary writer and reader

fl0wjacky:
1.when writing var int to stream, if the INT's byte length is 4,should write 'fe' first, not 'fd'.

2.when converting data from stream to bytearray,don't need unhexlify.

### DIFF
--- a/neo/IO/BinaryReader.py
+++ b/neo/IO/BinaryReader.py
@@ -139,7 +139,7 @@ class BinaryReader(object):
         items = []
         for i in range(0, 2000):
             data = self.ReadBytes(64)
-            ba = bytearray(binascii.unhexlify(data))
+            ba = bytearray(data)
             ba.reverse()
             items.append(ba.hex().encode('utf-8'))
         return items
--- a/neo/IO/BinaryWriter.py
+++ b/neo/IO/BinaryWriter.py
@@ -121,7 +121,7 @@ class BinaryWriter(object):
             return self.WriteUInt16(value, endian)
 
         elif value <= 0xFFFFFFFF:
-            self.WriteByte(0xfd)
+            self.WriteByte(0xfe)
             return self.WriteUInt32(value, endian)
 
         else:


neo/IO/BinaryReader.py
```diff
@@ -139,7 +139,7 @@ def Read2000256List(self):
         items = []
         for i in range(0, 2000):
             data = self.ReadBytes(64)
-            ba = bytearray(binascii.unhexlify(data))
```
ixje: The unhexlify is actually needed in this function. The data read from the stream, in the particular case that this function is used in, is non-raw data (as indicated by reading 64 bytes, as opposed to reading 32 bytes as in `ReadHashes()` below) i.e. `b'ef1f8f66a16fba100ed760f4ac6aa5a0d0bb8f4a0e92705b106761ef181718b3'`

When calling the `reverse()` function on this you'll get:
`b'3b817181fe167601b50729e0a4f8bb0d0a5aa6ca4f067de001abf61a66f8f1fe'`

while what actually is needed is:
`b'b3181718ef6167105b70920e4a8fbbd0a0a56aacf460d70e10ba6fa1668f1fef'`

Note that in the latter we reverse whole bytes, not just a single ascii char at a time.

Please revert the changes.

neo/IO/BinaryWriter.py
```diff
@@ -121,7 +121,7 @@ def WriteVarInt(self, value, endian="<"):
             return self.WriteUInt16(value, endian)
 
         elif value <= 0xFFFFFFFF:
-            self.WriteByte(0xfd)
```
ixje: I agree ([ref](http://docs.neo.org/en-us/node/network-protocol.html#convention))

neo/IO/BinaryReader.py
metachris: Thanks for the feedback @ixje. I'm just now moving this to `neocore` and will also take a look later tonight.

metachris:
Can you please split this up into 2 separate PRs? Would help, also for seeing the test implications.

fl0wjacky:
@metachris OK,I will split this up into 2.

metachris:
Please open the new pull requests to this repo: https://travis-ci.org/CityOfZion/neo-python-core, we're in the process of moving a few things over there, and Binary* is the first step already in there now.

fl0wjacky:
@metachris  Do I directly make PR to repo: https://github.com/CityOfZion/neo-python-core/ OK?

metachris:
yes, please 👍

PR #146, opened 2018-01-02
Move base datatypes to neocore

metachris:
* Added `neocore` dependency
* Removed base datatypes
* Changed the imports to `neocore`


metachris:
Too bad coverage decreased, guess this moved a couple of files with high coverage to neopython which lowers the overall average coverage.

localhuman:
Great work!

PR #150, opened 2018-01-04
Clarify config error message

ixje:
Currently when you try to configure an non existing event type, or simply mistype it you'll get a partially confusing error message.

> neo> config sc-evnt on
cannot configure sc-evnt
Try 'config log on/off'
neo> config log on
cannot configure log
Try 'config log on/off'

The "cannot configure <type>" message makes sense, but the one following that suggests you can type `config log on` or `config log off` which you cannot.

I changed the error message to:
`print("Invalid config target: %s", what)`

I believe that without giving a suggestion on what to do next, users will likely try `help` first and see the list of available config commands.
ran unit tests. no errors

**Did you add any tests?**
no

no

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -735,7 +735,8 @@ class PromptInterface(object):
                 print("cannot configure log.  Please specify on or off")
 
         else:
-            print("Invalid config target: %s", what)
+            print("cannot configure %s", what)
+            print("Try 'config sc-events on|off' or 'config debug on|off'")
 
     def parse_result(self, result):
         if len(result):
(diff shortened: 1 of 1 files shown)


metachris:
I would keep the error message, and just change it to what users can actually use: `config sc-events on|off` and `config debug on|off`.

Also we've disabled E501 (max line length) in setup.cfg, so you wouldn't need to change the formatting on those lines.

ixje:
I specifically choose to keep the error message generic so you don't have to add a (potentially in the future big) list of events. Any change of opinion based on this thought?

As for the line endings this is automatically enforced  by my IDE according to the PEP8 rules. I'll see if I can override it.

metachris:
I still think it would be better to give the user immediate help instead of having them go through the help command.

PR #154, opened 2018-01-08
Minor improvements on 'Wallet Rebuild' functionality

rockacola:
Lack of feedback after a valid `wallet rebuild`.

* Add a feedback message when `wallet rebuild` is complete.
* While at it, I've also create a method wrapper for it, keeping the command if-statement bit tidier.

Smoke test.

**Did you add any tests?**
Not applicable.

No.

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -265,6 +265,17 @@ class PromptInterface(object):
             self.Wallet = None
             print("closed wallet %s " % path)
 
+    def do_rebuild_wallet(self, arguments):
+        self.Wallet.Rebuild()
+        try:
+            item2 = int(get_arg(arguments, 1))
+            if item2 and item2 > 0:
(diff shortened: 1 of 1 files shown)


metachris:
Alternatively you could change the message to `Wallet rebuild in progress, this may take a while. You can check the progress with the 'wallet' command.)

rockacola:
Thanks for the education @metachris . I'll prefer to have both: direct feedback (saying that it is in progress) and result of the rebuild (when it completes, success or fail).

Is it possible to listen for rebuild completion without hardcode `print` into components? Alternative I'll do a bit of code dive later on to investigate.

metachris:
There is no automatic notification for this. Closing this PR for now, if you want to work on this again, let's talk in the #neopython channel or just create an issue here and we can help you find the right spots to dock on.

PR #156, opened 2018-01-09
neo.Cryptography and neo.Wallets.KeyPair -> neocore

metachris:

Second part of moving `neo.Cryptography` and `neo.Wallets.KeyPair` to `neocore`. Removing of files and replacing of imports.

n/a

Tests

**Did you add any tests?**

No

Update to neocore v0.3.1 is required.


localhuman:
looks great! `neocore` to the moon!

PR #163, opened 2018-01-11
correct wallet output for script_hash and address

ixje:
the `wallet` command output showed the Address of a wallet in the `script_hash` key instead of the actual scripthash (see below)

```
"addresses": [
        {
            "votes": [],
            "script_hash": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
            "version": 0,
```
as discussed added both `script_hash` and `address` as key/value pairs

run the test suite + manual testing
 
**Did you add any tests?**
no

no


localhuman:
👍

PR #166, opened 2018-01-14
Allow address generation to respect address version from settings

ambethia:

Address version is hard coded to 23, but is intended to be configurable in protocol.json.

See CityOfZion/neo-python-core#31

**Did you add any tests?**

Not sure if needed here, see tests in CityOfZion/neo-python-core#31


metachris:
👍

PR #167, opened 2018-01-14
Respect address version from settings

ambethia:

Address version is hard coded to 23, but is intended to be configurable in protocol.json.

I'm working with a neo fork that uses a different address version.

Use `settings.ADDRESS_VERSION` as default address version rather than hard-coded `23`

Tests

**Did you add any tests?**

Yep

I also submitted changes to `dev` branch and `neo-python-core`, but I had prepared this initially since the project I'm working on is using the master branch. So here's a backport.


metachris:
I'll close this because the related code on the master branch is deprecated, and this functionality is implemented with other PRs across neo-python-core and neo-python

PR #170, opened 2018-01-16
Update all the dependencies

metachris:

Updated the dependencies

n/a

tests

**Did you add any tests?**

no

no


metachris:
Keeping this open as reminder to myself. Will try to sort out the leveldb issues, or if anyone else wants to take a stab, feel free ;)

metachris:
plyvel not updated due to breaking changes. see #172

PR #171, opened 2018-01-16
Added makefile with a few useful commands

metachris:

Adds a Makefile with a few useful commands:

* make lint
* make test
* make coverage
* make docs

n/a

Manual tests

**Did you add any tests?**

no

no


localhuman:
PS, these are great!

PR #176, opened 2018-01-17
Final fix of plyvel issues, docs update

metachris:

Python LevelDB bindings now work out of the box on OSX. Also reported and helped solve an error with plyvel 1.0.3 which lacked snappy support and couldn't load the chain db.

n/a

tests

**Did you add any tests?**

no

no


localhuman:
Nice work!  And thanks to @wbolster for the quick releases of plyvel 🥇

PR #177, opened 2018-01-18
neo/api/REST/NotificationRestApi.py

metachris:

Moved `neo/Implementations/Notifications/REST/NotificationServer.py` to `neo/api/REST/NotificationRestApi.py`

moving

test

**Did you add any tests?**

no

no

### DIFF
--- a/examples/notification-rest-api-server.py
+++ b/examples/notification-rest-api-server.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 """
-This example provides a REST API to query blockchain data, implementing `neo.api.RESTAPI.RestApi`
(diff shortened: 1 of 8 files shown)


localhuman:
Though it is a REST api, my idea is that it should be solely for notifications and not a generic rest API.  So I guess I agree with moving to the new module, but not renaming `NotificationServer` to `RestApi`.  Perhaps could be renamed to `NotificationRestApi`?

metachris:
Sure we can do that! 

The current implementation has 3 api calls: getting info about a block, address and transaction. I'm wondering what you mean with "notifications" in particular -- could you elaborate?

metachris:
better like this? open to any suggestions ;)

localhuman:
👍

PR #178, opened 2018-01-21
Getting Started update

mjakl:

The current Getting Started documentation was a bit confusing for beginners (especially to me as one). The Python 3.5 part was very thin, plus figuring out exactly what it is and how it compares to the other projects in the ecosystem took some time.

I revised the whole Getting Started documentation, hopefully making it clearer and more concrete.

I followed the steps to setup my own environment.

**Did you add any tests?**

Unfortunately, no tests for documentation.

I refactored the documentation of "how to release", that's probably something you should double check if it still makes sense (that's hard to check from the outside).

Windows is still not covered well, but I assume that Windows users would opt for the C# implementation anyways... .

I did not update the documentation in the RST format, please let me know if you'd like to have it aligned.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,10 @@
 
 ## Overview
 
-neo-python is an alternative implementation of the original NEO VM and can be run un
-*nix and Windows environments.
+neo-python is an alternative implementation of the original NEO VM and is easy
+to install/run on *nix environments. For Windows, installing a dependency
+(LevelDB) is more involved, for serious activities, either use a Linux virtual
+machine or run the original NEO VM on Windows.
 
 It comes with batteries included, ie. it runs a full node (syncing and running contracts),
 plus, it comes with a CLI client (through the `prompt.py`) allowing you to develop/run/test
@@ -67,6 +69,11 @@ found at [Read The Docs](https://neo-python.readthedocs.io/en/latest/).
 
 ## Getting started
 
+We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get
+you started. There are monay more videos under the
+[CityOfZion](https://www.youtube.com/channel/UCzlQUNLrRa8qJkz40G91iJg) Youtube
+channel, check them out.
+
 neo-python has two System dependencies (everything else is covered with `pip`):
 
 - LevelDB
@@ -74,12 +81,7 @@ neo-python has two System dependencies (everything else is covered with `pip`):
 
 ### LevelDB
 
-We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get
-you started with this library. There are other videos under the
-[CityOfZion](https://www.youtube.com/channel/UCzlQUNLrRa8qJkz40G91iJg) Youtube
-channel.
-
-#### OSX:
+#### OSX
 
 ```
 brew install leveldb
@@ -115,24 +117,10 @@ support tied to Visual Studio and libraries.
 
 neo-python is currently only compatible with **Python 3.5** (due to its `byteplay3` dependency).
 
-On *nix systems, we can easily install more Python versions on one computer
-using [pyenv](https://github.com/pyenv/pyenv).
-
-After installing `pyenv` (with `brew`, `pip`, or any old package manager), add Python 3.5:
-
-```
-pyenv install 3.5.4
-```
-
-In our `neo-python` directory, make Python 3.5.4 our preferred version:
-
-```
-pyenv local 3.5.4
-```
-
-If we installed `pyenv` properly (especially the `eval "$(pyenv init -)"` part), it
-will automatically switch to Python 3.5.4 whenever we are in the
-neo-python directory.
+On *nix systems, install Python 3.5 via some package manager. It should not
+conflict with the existing System Python (though, if that concerns you, you can
+use third-party tools like [pyenv](https://github.com/pyenv/pyenv) to manage
+multiple versions of Python local to you user-account).
 
 #### Virtual Environment
 
@@ -147,6 +135,7 @@ source venv/bin/activate
 Now let's install neo-python's dependencies:
 
 ```
+pip install -U setuptools pip wheel
 pip install -e .
 ```
 
@@ -155,7 +144,6 @@ automatically is a nice little time-saver:
 
 Example `neo-project/.env`:
 ```
-eval "$(pyenv init -)"
 source venv/bin/activate
 ```
 
@@ -268,11 +256,23 @@ connection), but happens only once.
 
 This is a checklist for releasing a new version:
 
-- [ ] Increase the version number again (eg. scope changed from patch to minor): `bumpversion --no-tag patch|minor|major`
-- [ ] Update `CHANGELOG.md` and `docs/source/changelog.rst` with the new version number and the changes and commit this
-- [ ] Set the release version number and create the tag: `bumpversion release`
-- [ ] Increase patch number and add `-dev`: `bumpversion --no-tag patch`
-- [ ] Push to GitHub, which also updates the PyPI package: `git push && git push --tags`
+.. code-block:: console
+
+    # In case you want to increase the version number again (eg. scope changed from patch to minor):
+    bumpversion --no-tag patch|minor|major
+
+    # Update ``CHANGELOG.md`` and ``docs/source/changelog.rst`` with the new version number and the changes and commit this
+    vi CHANGELOG.md docs/source/changelog.rst
+    git commit -m "Updated Changelogs" CHANGELOG.md docs/source/changelog.rst
+
+    # Set the release version number and create the tag
+    bumpversion release
+
+    # Increase patch number and add `-dev`
+    bumpversion --no-tag patch
+
+    # Push to GitHub, which also updates the PyPI package
+    git push && git push --tags
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -276,12 +276,17 @@ This is a checklist for releasing a new version:
 
 ## Troubleshooting
 
-If you encounter any problems, please take a look at the [installation
-section](https://neo-python.readthedocs.io/en/latest/install.html#further-install-notes)
-in the docs, and if that doesn't help open an issue. We'll try to help.
+If you run into problems, check these things before ripping out your hair:
 
-You can reach us on the [NEO Discord](https://discord.gg/R8v48YA), or simply file
-a [GitHub issue](https://github.com/CityOfZion/neo-python/issues/new).
+* Double-check that you are using Python 3.5.x
+* Update the project dependencies (`pip install -e .`)
+* Compiling contracts with `neo-boa` can cause issues with `pyenv`,
+  try using Python without changed runtime-paths
+* If you encounter any problems, please take a look at the [installation
+  section](https://neo-python.readthedocs.io/en/latest/install.html#further-install-notes)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,18 +25,26 @@
 
 ## Overview
 
+neo-python is an alternative implementation of the original NEO VM and can be run un
+*nix and Windows environments.
```
metachris: Currently it can't easily be run on Windows (because of LevelDB bindings not being easy to install on Windows).

README.md
```diff
@@ -46,34 +54,44 @@
...
+  or documentation, or on any other feature you deem awesome. All successful
+  pull requests will be rewarded with one photo of a cat or kitten.
 
 ## Getting started
 
-You will need to install the libleveldb library. Install [Python 3.5](https://www.python.org/downloads/release/python-354/) to make sure you don't run into any issues with your version of Python being different than the current maintainer's version. Note that Python 3.6 is not currently supported due to incompatibilities with the byteplay module.
+neo-python has two System dependencies (everything else is covered with `pip`):
 
-We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get you started with this library. There are other videos under the [CityOfZion](https://www.youtube.com/channel/UCzlQUNLrRa8qJkz40G91iJg) Youtube channel.
+- LevelDB
```
metachris: Add the links here

README.md
```diff
@@ -46,34 +54,44 @@
...
+- Python **3.5** (3.6 and up are currently not supported)
 
-##### OSX:
+### LevelDB
+
+We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get
```
metachris: the videos should be where they were in "Getting started" and not in LevelDB

README.md
```diff
@@ -88,34 +106,68 @@ yum install -y epel-release
...
 
-Not sure. Installing the Python package plyvel seems to require C++ compiler support tied to Visual Studio and libraries.
+### Python 3.5
 
--------------------
+neo-python is currently only compatible with **Python 3.5** (due to its `byteplay3` dependency).
 
-### Virtual Environment
+On *nix systems, we can easily install more Python versions on one computer
+using [pyenv](https://github.com/pyenv/pyenv).
```
metachris: Not sure we gonna recommend `pyenv` as default tool. Perhaps we talk about this in a separate thread. Installing Python 3.5 on *nix is easy enough with the system package manager?

README.md
```diff
@@ -88,34 +106,68 @@ yum install -y epel-release
...
 ```
 python3.5 -m venv venv
 source venv/bin/activate
 ```
 
-Then install the requirements:
+Now let's install neo-python's dependencies:
 
 ```
-pip install -U setuptools pip wheel
```
metachris: we need this line, else it won't always install.

README.md
```diff
@@ -88,34 +106,68 @@ yum install -y epel-release
...
 pip install -e .
 ```
 
+Tip: using [autoenv](https://github.com/kennethreitz/autoenv) to load the configuration
+automatically is a nice little time-saver:
+
+Example `neo-project/.env`:
+```
+eval "$(pyenv init -)"
```
metachris: Personally I don't want to introduce yet another tool to get things running right now (`pyenv`).

README.md
```diff
@@ -188,72 +246,48 @@ optional arguments:
...
     make docs
 
-
 ## Updating the version number and releasing new versions of neo-python
 
-(Only for admins)
-
 This is a checklist for releasing a new version:
 
-.. code-block:: console
```
metachris: Leave this block as it was

localhuman:
Agree with what @metachris has to say.  Also I would add that `pyenv` causes issues on some platforms when compiling `neo-boa` contracts, since it messes with runtime import paths

README.md
mjakl: I've changed the wording, I think it's a good tip for people who don't want to mess up their system, but I agree, that way it sounds like a precondition which is not the case.

README.md
mjakl: The mentioning of `to help get you started with this library` (I thought it refers to LevelDB) was confusing me towards that interpretation, please check whether it fits now.

mjakl:
@localhuman I've added the `neo-boa` issues with `pyenv` in the troubleshooting section (as I assume that some people run neo-pythin with `pyenv` if we mention it in this guide or not).

I hope that's a satisfying solution?

metachris:
I'd rathe not add pyenv to the mix at this point in time. So don't even mention it in the readme for now. This and neo-boa compatibility is another discussion.

mjakl:
Ok, it's gone (replaced by a link to the official download page, which is one of two ways I found to get Python 3.5 on OS X without pyenv - the other being MacPorts according to StackExchange)

README.md
```diff
@@ -25,18 +25,28 @@
 
 ## Overview
 
+neo-python is an alternative implementation of the original NEO VM and is easy
+to install/run on *nix environments. For Windows, installing a dependency
+(LevelDB) is more involved, for serious activities, either use a Linux virtual
+machine or run the original NEO VM on Windows.
+
+It comes with batteries included, ie. it runs a full node (syncing and running contracts),
```
metachris: This part is kind of inaccurate -- you can do a bit more and can also test smart contracts written in any language (but not real debugging). I would just get rid of this full paragraph, and perhaps even the first one too.

README.md
```diff
@@ -46,34 +56,44 @@
...
-- Pull requests welcome. You can help with wallet functionality, writing tests or documentation, or on any other feature you deem awesome. All successful pull requests will be rewarded with one photo of a cat or kitten.
-
+- Pull requests welcome. You can help with wallet functionality, writing tests
+  or documentation, or on any other feature you deem awesome. All successful
+  pull requests will be rewarded with one photo of a cat or kitten.
 
 ## Getting started
 
-You will need to install the libleveldb library. Install [Python 3.5](https://www.python.org/downloads/release/python-354/) to make sure you don't run into any issues with your version of Python being different than the current maintainer's version. Note that Python 3.6 is not currently supported due to incompatibilities with the byteplay module.
+We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get
```
metachris: Add the videos section below the system dependencies

README.md
```diff
@@ -46,34 +56,44 @@
...
+you started. There are monay more videos under the
+[CityOfZion](https://www.youtube.com/channel/UCzlQUNLrRa8qJkz40G91iJg) Youtube
+channel, check them out.
+
+neo-python has two System dependencies (everything else is covered with `pip`):
 
-We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get you started with this library. There are other videos under the [CityOfZion](https://www.youtube.com/channel/UCzlQUNLrRa8qJkz40G91iJg) Youtube channel.
+- LevelDB
+- Python **3.5** (3.6 and up are currently not supported)
```
metachris: Leave the link to py3.5 as previously, and add one to leveldb too

metachris: Also add the reason why py3.6 is not supported as it was previously ("Note that Python 3.6 is not currently supported due to incompatibilities with the byteplay module.")

README.md
```diff
@@ -88,34 +108,52 @@ yum install -y epel-release
...
--------------------
+neo-python is currently only compatible with **Python 3.5** (due to its `byteplay3` dependency).
 
-### Virtual Environment
+On *nix systems, install Python 3.5 via some package manager, or download an installation package
+from the [official homepage](https://www.python.org/downloads/release/python-354).
 
-Now navigate into the project, make a Python 3 virtual environment and activate
-it via
+#### Virtual Environment
```
metachris: Don't change this to level 4 header, but keep it at level 3

README.md
```diff
@@ -88,34 +108,52 @@ yum install -y epel-release
...
+support tied to Visual Studio and libraries.
 
-Not sure. Installing the Python package plyvel seems to require C++ compiler support tied to Visual Studio and libraries.
+### Python 3.5
 
--------------------
+neo-python is currently only compatible with **Python 3.5** (due to its `byteplay3` dependency).
 
-### Virtual Environment
+On *nix systems, install Python 3.5 via some package manager, or download an installation package
```
metachris: say "via your package manager" (or homebrew on OSX).

README.md
```diff
@@ -88,34 +108,52 @@ yum install -y epel-release
...
 
-Then install the requirements:
+Now let's install neo-python's dependencies:
 
 ```
 pip install -U setuptools pip wheel
 pip install -e .
 ```
 
+Tip: using [autoenv](https://github.com/kennethreitz/autoenv) to load the configuration
```
metachris: autoenv is inactive and he says on the github page to try another project. i would not mention it here. people working with this should be able to run `python3.5 -m venv venv`

README.md
```diff
@@ -46,34 +56,44 @@
...
-
+- Pull requests welcome. You can help with wallet functionality, writing tests
+  or documentation, or on any other feature you deem awesome. All successful
+  pull requests will be rewarded with one photo of a cat or kitten.
 
 ## Getting started
 
-You will need to install the libleveldb library. Install [Python 3.5](https://www.python.org/downloads/release/python-354/) to make sure you don't run into any issues with your version of Python being different than the current maintainer's version. Note that Python 3.6 is not currently supported due to incompatibilities with the byteplay module.
+We have published a Youtube [video](https://youtu.be/oy6Z_zd42-4) to help get
+you started. There are monay more videos under the
```
metachris: "monay" more videos

metachris:
Alright. So lots of good changes to the readme imo, still needs a few more changes. Left the comments, please address them and we can probably merge this quickly.

README.md
mjakl: According to https://discourse.brew.sh/t/how-do-i-install-python-3-5-instead-of-3-6/675 Homebrew is not an option (I didn't manage to get it installed via brew)

mjakl: Correction, not an **easy** option.

metachris: Alright, good point. leave homebrew out then ;) 

Get on Discord and say hello btw.

mjakl:
Included almost all suggestions, just did not include Hombrew (according to my experience it's not possible to easily install Python 3.5 using `brew` - hence the initial `pyenv` discussion).

README.md
```diff
@@ -88,34 +105,44 @@ yum install -y epel-release
...
+### Python 3.5
 
--------------------
+neo-python is currently only compatible with **Python 3.5** (due to its `byteplay3` dependency).
+
+On *nix systems, install Python 3.5 via your package manager, or download an installation package
+from the [official homepage](https://www.python.org/downloads/release/python-354).
 
-### Virtual Environment
+##### Virtual Environment
```
metachris: wrong way -- i think it should be level 3 as it was, the last commit had it at level 4, now it's at level 5. just make it as it was with `### Virtual ...`

README.md
```diff
@@ -25,18 +25,24 @@
 
 ## Overview
 
+neo-python is an alternative implementation of the original NEO environment. It
```
metachris: I don't think this paragraph really helps. Perhaps remove it and just add "Build, deploy and run smart contracts" to the bullet list below?

metachris:
👍 

almost there. only 2 last comments for today ;)

mjakl:
Agreed, finished it off and committed. Thanks for the close check :)

localhuman:
This all good with you @metachris? Looks good to me!

metachris:
👍

PR #179, opened 2018-01-21
prompt.py style improvements, plus a more robust CLI Command Parser

mjakl:

**Did you add any tests?**

Besides the functional change (commit b64d721), I made some style changes (consistent use of single quotes and capital letters, closer adherence to PEP8)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 All notable changes to this project following the `0.4.0` release will be documented in this file.
 
+## [0.4.6-dev] Unreleased
+- implement a more robust CLI command parser
+
 ## [0.4.5] 2018-01-18
 - updated `neo-boa` to `0.2.2`, added support for array `REMOVE` VM opcodes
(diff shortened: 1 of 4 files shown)


mjakl:
Yes, Vienna Meetup, great presentation BTW!

So, `"` is the common style? in `prompt.py` it was a mixture of both. Will fix... .

metachris:
Thanks! And great that you're already contributing! 👍 

Yeah it's more typically doublequotes but sometimes a mix. Just don't fix that everywhere, makes the PR harder to understand. It's always better to keep the PRs simple and focused on one thing. If at all, changing quotes would be another separate PR ;-)

mjakl:
Ok, I've split this PR into two separate ones: #180 and #181. #181 will be easier to read when #180 is applied, but it was not (easily) possible to completely separate them. I reverted some of the `'`/`"` changes (kept only the ones that are very out of style) I hope that's better?

Closing this in favour of the other two.

mjakl:
#182 is a version of the parser without any style changes.

PR #180, opened 2018-01-21
Improve prompt.py style

mjakl:

The `prompt.py` responses were a bit inconsistent (casing, spacing etc.). This PR makes it a bit more consistent.

Adhere to PEP8 and the CoZ coding style used in other files.

No functional change.

**Did you add any tests?**

No.

No.

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 
-
 import argparse
 import datetime
 import json
@@ -9,7 +8,6 @@ import resource
 import traceback
 import logging
 
-import logzero
 from prompt_toolkit import prompt
 from prompt_toolkit.contrib.completers import WordCompleter
 from prompt_toolkit.history import FileHistory
@@ -186,7 +184,6 @@ class PromptInterface(object):
         print_tokens(tokens, self.token_style)
 
     def do_open(self, arguments):
-
         if self.Wallet:
             self.do_close_wallet()
 
@@ -437,7 +434,6 @@ class PromptInterface(object):
             WithdrawOne(self.Wallet)
 
     def do_notifications(self, arguments):
-
         if NotificationDB.instance() is None:
             print("No notification DB Configured")
             return
@@ -465,7 +461,6 @@ class PromptInterface(object):
             print("No events found for %s " % item)
 
     def show_wallet(self, arguments):
-
         if not self.Wallet:
             print("please open a wallet")
             return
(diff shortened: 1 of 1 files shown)


prompt.py
```diff
@@ -352,13 +349,13 @@ def do_export(self, arguments):
...
             if not self.Wallet:
-                return print("please open a wallet")
+                return print("Please open a wallet")
 
             address = get_arg(arguments, 1)
             if not address:
                 return print("Please specify an address")
 
-            passwd = prompt("[Wallet Password]> ", is_password=True)
+            passwd = prompt("[wallet password]> ", is_password=True)
```
metachris: Since most prompts now start capitalized, `wallet password` here should be `Wallet password`?

prompt.py
```diff
@@ -791,50 +784,48 @@ def configure(self, args):
...
     def run(self):
-
         dbloop = task.LoopingCall(Blockchain.Default().PersistBlocks)
         dbloop.start(.1)
 
         Blockchain.Default().PersistBlocks()
 
         tokens = [(Token.Neo, 'NEO'), (Token.Default, ' cli. Type '),
-                  (Token.Command, "'help' "), (Token.Default, 'to get started')]
+                  (Token.Command, "\'help\' "), (Token.Default, "to get started")]
```
metachris: this has no unintended consequences?

prompt.py
```diff
@@ -791,50 +784,48 @@ def configure(self, args):
...
 
         print_tokens(tokens, self.token_style)
-        print("\n")
+        print('\n')
```
metachris: why?

metachris:
Nice cleanups 👍  

Some minor comments inline.

prompt.py
mjakl: All prompts start with lower-case characters, only the responses are capitalised sentences.

Would you rather have the prompts capitalised as well? That's somewhat unusual.

prompt.py
mjakl: Sorry, a leftover from the back/forth with the quotes.

prompt.py
mjakl: It's the only `print("\n")` (1 out of 7), for the sake of consistency...

prompt.py
mjakl: Since we mix the usage of `'` and `"` in the same array, I've used `'` consistently in the next commit.

metachris:
Looks good to me, need to manually test it a bit more. @localhuman needs to sign off on this.

localhuman:
Looks great, thanks for the contributions 😸

PR #181, opened 2018-01-21
A more robust CLI Command Parser

mjakl:
Note, this PR depends on #180. For an independent version of this PR, please see #182 (in case #180 is a no-go).

**Did you add any tests?**

No.


mjakl:
Note, #182 is the same functionality, but without the style changes introduced with #180.

metachris:
Closing this because #182

PR #182, opened 2018-01-21
Improved prompt user-input parser

mjakl:
Note, this PR is an alternative to the #181 combination, but does not depend on #180.

**Did you add any tests?**

No.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ All notable changes to this project following the ``0.4.0`` release will be docu
 [0.4.7-dev] work in progress
 ----------------------------
 
-- ...
+- implement a more robust CLI command parser
 
 [0.4.6] 2018-01-24
 ------------------
--- a/neo/Prompt/InputParser.py
+++ b/neo/Prompt/InputParser.py
@@ -1,0 +1,11 @@
+from pyparsing import ZeroOrMore, Regex
+
+
+class InputParser(object):
+    parser = ZeroOrMore(Regex(r'\[[^]]*\]') | Regex(r'"[^"]*"') | Regex(r'\'[^\']*\'') | Regex(r'[^ ]+'))
+
+    def parse_input(self, user_input):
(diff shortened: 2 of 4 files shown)


localhuman:
Thanks for the PR!  One thing to make sure to check is that commands with python objects ( especially `bytearray` and `bytes` objects ) inside lists ( or outside of them ) still work ok... for example:
`testinvoke f8d448b227991cf07cb96a6f9c0322437f1599b9 transfer [bytearray(b'S\xefB\xc8\xdf!^\xbeZ|z\xe8\x01\xcb\xc3\xac/\xacI)'),bytearray(b'\xaf\x12\xa8h{\x14\x94\x8b\xc4\xa0\x08\x12\x8aU\nci[\xc1\xa5'),1000]`

mjakl:
Sure, will add tests later this day.

mjakl:
I added a test showing that bytearrays work outside and inside lists (with whitespaces in the outer list). And adapted the `CHANGELOG.md` to be more in line with the other entries.

metachris:
That's a nice set of tests 👍

metachris:
Tested it with invoke calls with the NEX ICO template and things work so far from a quick look. Like it. @localhuman can you give this a quick look and test?

mjakl:
Rebased the PR to `development` (adapted CHANGELOG.rst).

localhuman:
Hi @mjakl just wanted to say that I'm not ignoring this PR, I think it is a great changeset!  Unfortunately we've got a lot going on with the upcoming NeoDevCon thingy.  Anyways, I hope to get around to fully testing this after that.  Just wanted to let you know this hasn't been forgotten.

ixje:
This seems to solve exactly the problem I was running in to. I'll report back if I hit any issues.

localhuman:
This is great work! Thanks @mjakl 🎉

mjakl:
Awesome, thank you very much!

PR #184, opened 2018-01-22
Feature send attrs

localhuman:
Added ability to include transaction attributes with the `send` command.  example:
`send neo APRgMZHZubii29UXF9uFa6sohrsYupNAvx 10 --tx-attr={'usage':241,'data':'My Remark'}`

-added a parser that looks for the `--tx-attr` flag

- manual and automated testing

**Did you add any tests?**
- added new tests in `neo/Prompt/Command/tests/test_send_command`

- Transaction attributes are awesome, good way to add meta-data to transactions.  but don't overuse them :)


metachris:
Very nice! We need to document this somewhere too ;)

PR #186, opened 2018-01-22
Fix params parsing & getblockcount method

ixje:
Fix params parsing & getblockcount method
add getblockhash example

parsed `params` with `ast`.
write tests

tests

**Did you add any tests?**
yes

no

### DIFF
--- a/neo/api/JSONRPC/JsonRpcApi.py
+++ b/neo/api/JSONRPC/JsonRpcApi.py
@@ -14,7 +14,6 @@ from neo.Implementations.Notifications.LevelDB.NotificationDB import Notificatio
 from logzero import logger
 from neo.Core.Blockchain import Blockchain
 from neo.api.utils import json_response
-import ast
 
 
 class JsonRpcErrors:
@@ -66,7 +65,7 @@ class JsonRpcApi(object):
             if "method" not in body:
                 raise JsonRpcError("Field 'method' is missing", JsonRpcErrors.INVALID_REQUEST)
 
-            params = ast.literal_eval(body["params"]) if "params" in body else None
+            params = body["params"] if "params" in body else None
             result = self.json_rpc_method_handler(body["method"], params)
             return self.get_response(body["id"], result)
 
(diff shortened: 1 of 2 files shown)


neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -65,7 +66,7 @@ def home(self, request):
             if "method" not in body:
                 raise JsonRpcError("Field 'method' is missing", JsonRpcErrors.INVALID_REQUEST)
 
-            params = body["params"] if "params" in body else None
+            params = ast.literal_eval(body["params"]) if "params" in body else None
```
metachris: wut, whats going on here? json does correct parsing of the string already, what's the use-case for this?

neo/api/JSONRPC/test_json_rpc_api.py
```diff
@@ -120,9 +67,16 @@ def test_invalid_method(self):
...
-        # TODO: currently not returning a correct header height
-        ndb = NotificationDB.instance()
         req = self._gen_rpc_req("getblockcount")
         mock_req = mock_request(json.dumps(req).encode("utf-8"))
         res = json.loads(self.app.home(mock_req))
-        self.assertEqual(res["result"], 123)
+        self.assertEqual(756620, res["result"])
+
+    def test_getblockhash(self):
+        req = self._gen_rpc_req("getblockhash", params='[2]')
```
metachris: here you should do `params=[2]`, then you don't need that `ast` stuff

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -106,5 +106,10 @@ def get_error_payload(self, request_id, error, data=None, code=None):
     def json_rpc_method_handler(self, method, params):
         # print("method", method, params)
         if method == "getblockcount":
-            return str(Blockchain.Default().HeaderHeight)
+            return Blockchain.Default().HeaderHeight
+        if method == "getblockhash":
+            height = params[0]
+            if height >= 0 and height <= Blockchain.Default().Height:
+                return Blockchain.Default().GetBlockHash(height).decode('utf-8')
```
metachris: Please add an `else` block here, and throw a meaningful error ("block number too high" or smth). Like this it would return the "Method not found" error

PR #187, opened 2018-01-23
Implementing a few methods

localhuman:
- added a few method implementations to the JSONRPC server
- added in placeholders for all other methods listed in C# implementation

code

tests

**Did you add any tests?**
yes

no


metachris:
very nice! 👍

PR #188, opened 2018-01-23
fix getaccountstate for invalid address

ixje:
fix getaccountstate for invalid address

add exception handler + extra test

test case

**Did you add any tests?**
yes

no


metachris:
Cherry-picked the relevant commit manually.

PR #190, opened 2018-01-23
Define VersionName in settings, removed from jsons

metachris:
`"VersionName": "/NEO-PYTHON:2.3.4/"` was hardcoded in the protocol jsons, i removed it from the jsons and just define it once in settings.py, which will always use the correct current version of neo-python.

Added this to the JsonRpcApi getversion response too.

### DIFF
--- a/examples/json-rpc-api-server.py
+++ b/examples/json-rpc-api-server.py
@@ -72,15 +72,16 @@ def main():
     ndb = NotificationDB.instance()
     ndb.start()
 
-    api_server = JsonRpcApi(settings.NODE_PORT)
-
     # Run
     reactor.suggestThreadPoolSize(15)
     NodeLeader.Instance().Start()
 
     host = "0.0.0.0"
-    print("Starting server on %s:%s" % (host, settings.NODE_PORT))
-    api_server.app.run(host, settings.NODE_PORT)
+    port = settings.RPC_PORT
+    print("Starting json-rpc api server on http://%s:%s" % (host, port))
+
+    api_server = JsonRpcApi(port)
+    api_server.app.run(host, port)
 
 
 if __name__ == "__main__":
--- a/examples/json-rpc-api-server.py
+++ b/examples/json-rpc-api-server.py
@@ -35,7 +35,6 @@ def main():
     parser.add_argument("-p", "--privnet", action="store_true", default=False,
                         help="Use PrivNet instead of the default TestNet")
     parser.add_argument("-c", "--config", action="store", help="Use a specific config file")
-    parser.add_argument("--port", action="store", help="Port to use", default=8000)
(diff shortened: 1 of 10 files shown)


metachris:
Happy about improvements to the `nonce` thing, if anyone has an idea. Else we can leave it like this for now and improve that later on.

examples/json-rpc-api-server.py
```diff
@@ -73,13 +72,15 @@ def main():
     ndb = NotificationDB.instance()
     ndb.start()
 
-    api_server = JsonRpcApi(args.port)
+    api_server = JsonRpcApi(settings.NODE_PORT)
```
metachris: I want to keep this for now, since a user might want to start the json-rpc api on a non-standard port, therefore I don't want to hardcode `settings.NODE_PORT` within `JsonRpcApi` for the `getversion` response.

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -65,10 +67,16 @@ def internalError(message=None):
...
     port = None
+    nonce = None
 
     def __init__(self, port):
         self.port = port
 
+        # `nonce` in neo-cli this comes from https://github.com/neo-project/neo/blob/46c635adda6cf61090a67a3bef0f3bb1bc85dd81/neo/Network/LocalNode.cs#L68
+        # In neo-python the closest thing to it is https://github.com/CityOfZion/neo-python/blob/master/neo/Network/NeoNode.py#L34
+        # since we don't get direct access to this NeoNode, we generate a new nonce here
+        self.nonce = random.randint(1294967200, 4294967200)
```
ixje: FYI; 
`NeoNode.py` is a remote node, this `Nonce` we require should ideally be from the equivalent of `LocalNode.cs` which is `NodeLeader.py`. 

`NodeLeader.Instance().NodeId` (you might even consider to rename `NodeId` to `Nonce`, but that's minor)

ixje:
I'll leave the `Version` location discussion up to @localhuman. I'm fine with either solution (settings or json)

examples/json-rpc-api-server.py
localhuman: RPC server shouldn't use the node port, as the base network API will be using that.

neo/api/JSONRPC/JsonRpcApi.py
metachris: 👍

examples/json-rpc-api-server.py
localhuman: So I guess we should probably add an `RPC` port to the settings file

localhuman:
I think its good to use the actual neo-python version

examples/json-rpc-api-server.py
metachris: Lol, of course ^^

PR #191, opened 2018-01-23
add `getblocksysfee` error test cases

ixje:
add edge case tests

write tests

tests

**Did you add any tests?**
yes
no


metachris:
:+1:

PR #192, opened 2018-01-23
Localhuman rpc2

localhuman:
- upgrade to `neocore` `v.3.2`
- add hex string specifier to all UInt represented in JSON
- fixed how blocks are serialized so that non-verbose blocks can be retrieved from api
- added full transactions into output for blocks

- code
- tests
**Did you add any tests?**
- tests
- no


metachris:
👍

Issue #169, opened 2018-01-15
[Feature] Implement new voting changes


localhuman:
Recent changes in NEO core added a new Transaction type along with a number of other large changes around voting and consensus.  These changes should be implemented in `neo-python`.

New changes in NEO:
https://github.com/neo-project/neo/pull/130

localhuman:
I've pushed a new branch for these changes here: https://github.com/CityOfZion/neo-python/tree/feature-state-transaction

These are the non-consensus related changes we will be needing for sure:
- add `neo.Core.StateDescriptor`
- add `neo.Core.TX.StateTransaction`
- update `neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain.PersistBlocks` to handle the new `StateTransaction`

There are probably additional changes needed as well, but these will provide a start.

ixje:
reminder for #67

PR #193, opened 2018-01-24
Added support for StateTransaction and StateDescriptors

localhuman:
- adds support for `StateTransactions` and `StateDescriptors`, which are now appearing on testnet due to changes in `v2.6.0`
- read more here: https://github.com/CityOfZion/neo-python/issues/169

- added `StateTransaction` and `StateDescriptor`

- tests

**Did you add any tests?**
- yes

- this changeset allows `neo-python` to function as normally when `v2.6.0` nodes are on the network
- this is not a full implementation of the new functionality, as a full implementation is not needed unless we are performing consensus
- with that in mind, I will leave the issue open as these current changes do not fully solve the issue


metachris:
👍 Awesome!

For reference, it fixes this error: 
wallet-terminal

PR #194, opened 2018-01-24
Prints each result type conveniently (raw,int,str,bool)

nickazg:
Smart contract results are always printed as a bytearray, But its nice to be able to seee what the output represents as different types. eg its currently difficult to see the value of a BigInterger, A bool is easy to tell but its also handy to see it as a quick True/False. 

Because neo-python is awesome there was already conversion methods provided with the ByteArray object. All I did was duplicate and indent the existing "results" print statement

I have been using it for a while, and I am using preexisting methods and just simply printing them.

**Did you add any tests?**
no, I am using preexisting methods and just simply printing them

Not really, everything else is the same


metachris:
Hi! Could you please make `development` the target branch, and resolve the conflict?

nickazg:
@metachris yep sorry, will do!

PR #196, opened 2018-01-24
Prints each result type conveniently (raw,int,str,bool)

nickazg:
Smart contract results are always printed as a bytearray, But its nice to be able what the output represents as different types. eg its currently difficult to see the value of a BigInterger, while a bool is easy to tell, its also handy to see it as a quick True/False. 

Because neo-python is awesome there was already conversion methods provided with the ByteArray object. All I did was duplicate and indent the existing "results" print statement

I have been using it for a while, and I am using preexisting methods and just simply printing them.

**Did you add any tests?**
no, I am using preexisting methods and just simply printing them

Not really, everything else is the same


metachris:
Oh, you could just have edited the PR -- at the top you can click the pencil and change the target branch. Nvm. There is still a conflict with `prompt.py` though.

metachris:
I'm afraid the conflict is not really resolved -- take a look at the changes here: https://github.com/CityOfZion/neo-python/pull/196/files

nickazg:
Yeah I forgot to update my fork to the latest so figured it was easier starting a new request.. That all looks good now?

nickazg:
Ahh annoying, my local repo isnt complaining when I pull from here.. Bare with me for a sec haha.

nickazg:
yeah my local prompt file wasn't updating in my IDE when I pulled, reopening gave me the conflict. Should be all good now, sorry!

prompt.py
```diff
@@ -736,10 +736,14 @@ def load_smart_contract(self, args):
...
-                    print("Results %s " % [str(item) for item in results])
-                    print("Deploy Invoke TX GAS cost: %s" % (tx.Gas.value / Fixed8.D))
-                    print("Deploy Invoke TX fee: %s" % (fee.value / Fixed8.D))
+                    print("Total operations executed: %s " % num_ops)
+                    print("Results:")
+                    print("\t(Raw) %s" % [str(item) for item in results])
+                    print("\t(Int) %s" % [item.GetBigInteger() for item in results])
+                    print("\t(Str) %s" % [item.GetString() for item in results])
+                    print("\t(Bool) %s" % [item.GetBoolean() for item in results])
+                    print("Deploy Invoke TX gas cost: %s " % (tx.Gas.value / Fixed8.D))
```
metachris: keep `GAS` all caps please

metachris:
Interesting contribution, first of all thanks for taking the time 👍 

I'm not 100% sure we want this as default behavious, would like to hear the thoughts of @localhuman on this.

Alternatively, we might put this behind a feature-flag or setting (`config verbose-deploy on/off` and persist this in the userprefs), for instance.

metachris:
Really not 100% sure what to do with this. I somehow like it, actually, but not sure it should be default. We can leave it open for now and think a little more about it. For new developers, this info is actually quite nice to have :)

nickazg:
Yeah its only really a visual convenience thing. Especially for new developers who have so much to take in already, it can be somewhat confusing/unintuitive if they input an Int/bool/string and get returned a b'd' when they were expecting 100 (they should still understand why though) . If they haven't implemented their own node yet its just another step, even if you understand why a byte array is returned. Just food for thought.

PR #197, opened 2018-01-25
add getstorage RPC call

ixje:
add `getstorage` rpc call (thanks to @localhuman for spotting the reversed hash requirement)

-

tests

**Did you add any tests?**
yes

no

yes
feature

### DIFF
--- a/neo/api/JSONRPC/JsonRpcApi.py
+++ b/neo/api/JSONRPC/JsonRpcApi.py
@@ -136,6 +136,22 @@ class JsonRpcApi(object):
         shash_reversed.reverse()
         return UInt160(data=shash_reversed)
 
+    def getstorage(self, params):
+        reverse_hash = bytearray.fromhex(params[0])
+        reverse_hash.reverse()
+        script_hash = UInt160(data=reverse_hash)
+
+        if len(params) < 2:  # note this is not in the original RPC server,I assume they forgot this as it throws an exception
+            raise JsonRpcError(-100, "Invalid key")
+        key = bytearray.fromhex(params[1])
+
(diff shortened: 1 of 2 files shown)


neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -216,7 +224,20 @@ def json_rpc_method_handler(self, method, params):
             raise NotImplementedError()
 
         elif method == "getstorage":
-            raise NotImplementedError()
+            reverse_hash = bytearray.fromhex(params[0])
+            reverse_hash.reverse()
+            script_hash = UInt160(data=reverse_hash)
+
+            if len(params) < 2:  # note this is not in the original RPC server,I assume they forgot this as it throws an exception
+                raise JsonRpcError(-100, "Invalid key")
```
metachris: better error would be "missing second argument 'key'"

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -216,7 +224,20 @@ def json_rpc_method_handler(self, method, params):
             raise NotImplementedError()
 
         elif method == "getstorage":
-            raise NotImplementedError()
+            reverse_hash = bytearray.fromhex(params[0])
```
metachris: quite a few lines of code, i'd prefer to move this into it's own method

metachris:
👍 

Looks good to me

localhuman:
This looks good, but I had already added in my branch with some other changes in it.

ixje:
ah didn't know. Will close this then

Issue #189, opened 2018-01-23
JSON RPC API Functionality


metachris:
Work in progress, now already in dev. See also https://github.com/CityOfZion/neo-python/issues/273 for tracking issues.

Links:

* http://docs.neo.org/en-us/node/api.html
* https://github.com/neo-project/neo/blob/master/neo/Network/RPC/RpcServer.cs
* http://www.jsonrpc.org/specification
* [dry coded rpc server bodies by @ixje](https://gist.github.com/ixje/eb8e6fef5011d0232e6b44b7097da584)
* [example calls at neo-cli-json-rpc-docs.md](https://github.com/CityOfZion/neo-python/blob/jsonrpc1/neo/api/JSONRPC/neo-cli-json-rpc-docs.md)

Initial calls:

- [X] getaccountstate
- [X] getbestblockhash
- [X] getblockcount
- [X] getblockhash
- [X] getblocksysfee
- [X] getconnectioncount
- [X] getrawmempool
- [x] getpeers
- [X] getrawtransaction
- [X] getstorage
- [X] getversion
- [X] sendrawtransaction
- [X] validateaddress
- [X] invoke
- [X] invokefunction
- [X] invokescript
- [X] getassetstate
- [X] getblock (int)
- [X] getblock (hash)
- [X] gettxout (the C# side is broken [#156](https://github.com/neo-project/neo/issues/156))
- [X] getcontractstate (TODO: can't successfully run this call against the live nodes)
- [x] getbalance
- [x] getnewaddress

Additional calls

- [x] listaddress ([more info](https://github.com/neo-project/neo-cli/issues/87#event-1443067788))
- [ ] getapplicationlog ([more info](https://github.com/neo-project/neo-cli/pull/102/files))
- [ ] dumpprivkey
- [x] sendtoaddress
- [x] sendmany

See also:

* https://github.com/CityOfZion/neo-python/issues/233 (JsonRpcApi input & output differs from C#)

ixje:
fyi; I'm working on `gettxout`, but the C# side is broken [#156](https://github.com/neo-project/neo/issues/156)

metachris:
added 2 new calls to the list: 

* listaddress ([see this](https://github.com/neo-project/neo-cli/issues/87#event-1443067788))
* getapplicationlog ([see this](https://github.com/neo-project/neo-cli/pull/102/files))

ixje:
The list can be updated as follows:

**initial calls**
* check `getstorage`

[...]

PR #198, opened 2018-01-25
add validate address rpc call + tests

ixje:
add validate address rpc call + tests

address part of #189 

implement call + tests

tests

**Did you add any tests?**
yes

no

yes, no

feature


neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -128,6 +129,21 @@ def parse_uint_str(self, param):
             return param[2:]
         return param
 
+    def validateaddress(self, params):
+        # check for [] parameter or [""]
+        if params is None or params[0] == '':
```
metachris: this code doesn't yet check for an empty list (it is not None). perhaps upgrade to `if params is None or len(params) == 0 or params[0] == '':`

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -237,8 +253,7 @@ def json_rpc_method_handler(self, method, params):
             raise NotImplementedError()
 
         elif method == "validateaddress":
-            raise NotImplementedError()
-
+            return self.validateaddress(params)
```
metachris: one line whitespace before the next elif would be nice

metachris:
very nice 👍 
just 2 minor comments

neo/api/JSONRPC/JsonRpcApi.py
ixje: I should have anticipated this as I initially was surprised as well.  However, this code is correct given the current `gen_rpc_req` implementation. Here's why

```
req = self._gen_rpc_req("validateaddress", params=[])
```
returns a `dict` without the `params` keyword, therefor `if params is None` actually tests for the empty list. I'm assuming an actual request will be parsed similarly.

```
In [3]: def test_list(params=None):
   ...:     if params:
   ...:         print("yes params")
   ...:     else:
   ...:         print("no params")
   ...:

In [4]: test_list([])
no params
```

metachris: I understand what you are saying. but this is not 100% correct. `if not params` matches an empty list, `if params is None` doesn't. you can just replace this if here with `if not params or params[0] == '':` for the intended effect.

localhuman:
The way neon-js and c# checks valid address is to convert to scripthash and then back.  if it comes back the same it is valid

ixje:
@localhuman yeah I'm aware there are much better ways to do it, but I just followed the `RpcServer.cs` code [from here](https://github.com/neo-project/neo/blob/master/neo/Network/RPC/RpcServer.cs#L240) to [here](https://github.com/neo-project/neo/blob/master/neo/Wallets/Wallet.cs#L399-L406). The parts you talk about is in the actual Wallet code (I do the same in my [mobile wallet](https://github.com/CityOfZion/neon-wallet-react-native/blob/master/app/api/crypto/index.js#L44))

ixje:
@localhuman I'd like to keep the current validation test until we've implemented `isValidPublicAddress` discussed [here](https://github.com/CityOfZion/neo-python-core/issues/20) in `neo-python-core`. That can then be the extensive version that also does the checksum validation you mentioned. Would that work for you?

localhuman:
That works for me.  Would you mind fixing conflicts :)

localhuman:
best commit message ever 🤣

PR #199, opened 2018-01-26
Localhuman rpc3

localhuman:
- Adds support for `getrawtransaction`
- Adds more robust `ContractParameter` implementation for parsing to/from json
- Adds support for `getstorage`, `invoke`,`invokefunction`, and `invokescript`
- updates to `neocore=0.3.3`
- uses `ToJsonString` rather than `ToString` for `Fixed8` representation in json

- fun

- tests

**Did you add any tests?**
- yes

- hopefully not.  There were a lot of underlying changes that needed to be made to support the RPC Invoke related functionality, so there is a fair amount of code changes.

- yes

- yes


prompt.py
```diff
@@ -243,6 +243,8 @@ def do_create(self, arguments):
                     print("Pubkey %s" % key.PublicKey.encode_point(True))
                 except Exception as e:
                     print("Exception creating wallet: %s" % e)
+                    import traceback
```
metachris: I'd prefer the import on top of the file ;)

metachris:
Very nice! Waiting for input from @ixje, but looks good to me 👍

neo/SmartContract/ContractParameter.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @staticmethod
+    def ParamToUInt160(param):
+        shash_reversed = bytearray(binascii.unhexlify(ContractParameter.Parse0x(param)))
+        shash_reversed.reverse()
+        return UInt160(data=shash_reversed)
+
+    @staticmethod
+    def ParamToUInt256(param):
+        shash_reversed = bytearray(binascii.unhexlify(ContractParameter.Parse0x(param)))
```
ixje: Should this perhaps be a utility function in a more generic class or maybe even `neocore`? `UInt256` would be used for `Transaction`s beyond smart contracts (e.g. reversing the assetid for an transaction output entry, or handling the transaction id for an input entry)

localhuman:
Might be a good idea to add a `Parse` method to `UInt` in `neocore` that will take a string argument and return either a 160 or 256, similar to how c# does it.

Will add doc strings and clean up `getrawtransaction` and `invokefunction` :)

neo/SmartContract/ContractParameter.py
```diff
@@ -8,35 +8,33 @@
...
 
     Type = None
     Value = None
 
     def __init__(self, type, value):
+        """
+
+        Args:
+            type:
+            value:
```
ixje: what are the types of `self.Type` and `self.Value`?

neo/SmartContract/ContractParameter.py
```diff
@@ -8,35 +8,33 @@
...
+        """
+        Convert a StackItem to a ContractParameter object
 
-    @staticmethod
-    def ParamToUInt256(param):
-        shash_reversed = bytearray(binascii.unhexlify(ContractParameter.Parse0x(param)))
-        shash_reversed.reverse()
-        return UInt256(data=shash_reversed)
+        Args:
+            item (neo.VM.InteropService.StackItem) The item to convert to a ContractParameter object
```
ixje: minor, missing `:` after type

neo/SmartContract/ContractParameter.py
```diff
@@ -55,6 +53,12 @@ def ToParameter(item: StackItem):
             return ContractParameter(type=ContractParameterType.InteropInterface, value=item.GetInterface())
 
     def ToJson(self):
+        """
+        Converts a ContractParameter instance to a json representation
+
+        Returns:
+            json object
```
ixje: The convention I used in the past was:
```
        """
        Convert object members to a dictionary that can be parsed as JSON.

        Returns:
             dict:
        """
```
The description isn't that important, but I believe the `return` type could be more explicit (as there's no json object type in python afaik)

neo/SmartContract/ContractParameter.py
```diff
@@ -96,6 +106,16 @@ def ToVM(self):
 
     @staticmethod
     def FromJson(json):
+        """
+        Convert a json object to a ContractParameter object
+
+        Args:
+            item (dict) The item to convert to a ContractParameter object
```
ixje: missing `:` between type and description in docstring

neo/SmartContract/ContractParameterType.py
```diff
@@ -27,14 +30,30 @@ def __str__(self):
 
     @staticmethod
     def FromString(val):
+        """
+        Create a ContractParameterType object from a str
+
+        Args:
+            val (str) the value to be converted to a ContractParameterType
```
ixje: same as previous comment

neo/SmartContract/ContractParameterType.py
```diff
@@ -27,14 +30,30 @@ def __str__(self):
...
 
 
 import inspect
 
 
 def ToName(param_type):
+    """
+    Gets the name of a ContractParameterType based on its value
+    Args:
+        param_type (ContractParameterType) type to get the name of
```
ixje: see previous comment

neo/Core/TX/Transaction.py
```diff
@@ -140,17 +140,18 @@ def Deserialize(self, reader):
         if self.ScriptHash is None:
             raise Exception("Script hash is required from deserialize!!!!!!!!")
 
-    def ToJson(self):
+    def ToJson(self, index):
         """
         Convert object members to a dictionary that can be parsed as JSON.
```
ixje: `Args` missing from docstring

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -196,8 +168,8 @@ def json_rpc_method_handler(self, method, params):
             return len(NodeLeader.Instance().Peers)
 
         elif method == "getcontractstate":
-            script_hash = self.parse_uint_str(params[0])
-            contract = Blockchain.Default().GetContract(script_hash)
+            script_hash = UInt160.ParseString(params[0])
+            contract = Blockchain.Default().GetContract(script_hash.ToBytes())
             if contract is None:
                 raise JsonRpcError(-100, "Unknown contract")
```
ixje: Sorry one last item (was waiting for the new fixtures to download). There's no test covering this exception branch. Would be great to have that, because currently we're near 100% coverage (just need to add the unimplemented methods and cover this path) 💯

PR #201, opened 2018-01-27
Add `gettxout` RPC call + tests

ixje:
add `gettxout` RPC call for #189 

add + write tests

make test

**Did you add any tests?**
yes

changed storage prefix in `GetUnspent()` call from `DBPrefix.ST_SpentCoin` to `DBPrefix.ST_Coin` to make the call functional

yes

feature


localhuman:
🥇

PR #206, opened 2018-02-01
Refactor typo for the word cloning

melvin0008:
There is a typo when we print cloning from live under implementation/leveldb

Correct the spelling

Ran neo wallet on a private net

**Did you add any tests?**
No

No

No

development


metachris:
thanks 👍

PR #214, opened 2018-02-08
Fix wallet encryption

dethos:
This pull request addresses a wallet security issue present in the current master branch. The issue regards the encryption method being used, as it does not match the one used on the reference implementation. At this moment the master key is encrypted using the `passwordHash` which is also stored on the wallet (_the key is right next to the lock_).

The problem was solved using the same method that is used in the reference (C#) implementation. The software should calculate the `sha256` hash of the password and then calculate the `sha256` hash of the hash, and call it `encryption key`. This `key` should be used to encrypt the master key (and never stored, only the `sha256` hash of it).

Manually. Initially I could open all my wallets without needing their passwords (just by removing 2/3 lines of code). Now I can't open the wallets without a valid password.

**Did you add any tests?**
No

After this pull request is merged, users will not be able to open the wallets created by previous versions. Because of this and as a PoC, a small script (`reencrypt_wallet.py`) was made to reset the wallet's current password in a way that it could be used in this new version. 
The script duplicates the wallet to avoid losing any data. Perhaps it should be recommended that after "migrating" and making sure the new wallet can be opened, to create a new wallet and transfer everything to it, since this script does not change the master key, it only re-encrypts it with a new key. 

Yes, `make test`

This PR is being made into a feature-branch as indicated by the repository maintainer, at the time he was informed of the issue. 

cc: @localhuman @metachris


metachris:
Thanks, looks important. Can we detect if the user is trying to open an old wallet and print some info?

Btw, the tests are failing, please run `make lint` and `make test`.

@localhuman what do you think?

localhuman:
@metachris this is an important changeset, thanks very much to @dethos for bringing up the issue and solution 🥇.

We'll need to fix the tests before we can get much further, we can go from there

dethos:
Correct, I will try to fix the tests as soon as possible, so you can move forward with this.

dethos:
@localhuman the tests were fixed and the test wallets "migrated" using the `reencrypt_wallet.py` script.

PR #218, opened 2018-02-12
Windows support

iNomaD:
It adds installation instructions and a small change in prompt.py

There was a bunch of problems and I included the solutions to all of them

I tested by myself

**Did you add any tests?**
No

Not really

No

To development

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -4,7 +4,7 @@ import argparse
 import datetime
 import json
 import os
-import sys
+import psutil
 import traceback
 import logging
 
@@ -757,13 +757,10 @@ class PromptInterface(object):
                     return
 
     def show_mem(self):
-        if sys.platform == 'win32':
-            totalmb = "unknown "  # not supported
-        else:
-            import resource
-            total = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
-            totalmb = total / 1000000
-        out = "Total: %sMB\n" % totalmb
+        process = psutil.Process(os.getpid())
+        total = process.memory_info().rss
+        totalmb = total / (1024 * 1024)
+        out = "Total: %s MB\n" % totalmb
         out += "Total buffers: %s\n" % StreamManager.TotalBuffers()
         print_tokens([(Token.Number, out)], self.token_style)
 
--- a/prompt.py
+++ b/prompt.py
@@ -4,7 +4,7 @@ import argparse
 import datetime
 import json
 import os
-import resource
+import sys
 import traceback
 import logging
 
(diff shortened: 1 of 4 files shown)


prompt.py
```diff
@@ -757,8 +757,12 @@ def load_smart_contract(self, args):
                     return
 
     def show_mem(self):
-        total = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
-        totalmb = total / 1000000
+        if sys.platform == 'win32':
+            totalmb = "unknown "  # not supported
+        else:
+            import resource
```
metachris: keep this import at top of the file please, we try to avoid inline imports.

prompt.py
```diff
@@ -757,8 +757,12 @@ def load_smart_contract(self, args):
                     return
 
     def show_mem(self):
-        total = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
-        totalmb = total / 1000000
+        if sys.platform == 'win32':
+            totalmb = "unknown "  # not supported
```
metachris: is there no way to get the memory usage on windows?

prompt.py
iNomaD: It will not run then :(

metachris: Oh really? Well, then let's keep it like this ;)

prompt.py
iNomaD: I am not an expert of Python. The help of someone is appreciated!

metachris:
Thanks for the contribution, nice docs! Does it actually work well like this, I suppose you manually tested it? 

Can you please also run `make test` to see if the tests pass in Windows too?

iNomaD:
Unfortunately, they don't pass in Windows. Just tested manually.

prompt.py
ixje: can you try this?
https://stackoverflow.com/a/2017608/3869515

iNomaD: @ixje, it's not the same, it indicates the total memory of OS.

iNomaD: What exactly should the function `show_mem` return?
If the goal is memory currently used by the process then we can use a cross-platform solution instead of `resource`.
https://stackoverflow.com/a/21632554/5644693

ixje: oh I missed the specifics. One of these 2 should probably do then:
[still using WMI](https://stackoverflow.com/a/938800/3869515) 
[possible better alternative](https://stackoverflow.com/a/21632554/3869515)

iNomaD: Btw, I've just tested it on Ubuntu and it seems like `resource.getrusage(resource.RUSAGE_SELF).ru_maxrss` is the maximum value over all previous `process.memory_info().rss`. So what do we really need: the maximum or the current?

metachris:
Can you post the output in a separate gist somewhere? Would be interested where it fails.

prompt.py
metachris: We want to show as close as possible the *current* memory consumption.

iNomaD:
@metachris That's it. https://gist.github.com/iNomaD/a73eed3d93f919522647627ba56bb3ed

metachris:
@localhuman for it too?

localhuman:
thanks @iNomaD, great contribution! 🎉

localhuman:
Would be great if we can eventually get tests to pass on windows :)

metachris:
I think they'll only pass inside the Linux subsystem, too many things not supported natively.

PR #220, opened 2018-02-14
Fix gas costs

localhuman:
Fixes error in Gas cost calculation for `Storage.Put`

- cast calculation to int in the same way a certain part of the c# implicitly does

- wrote tests for gas costs, compared to results of operations performed via JSON-RPC invoke on the C# nodes

**Did you add any tests?**
- yes

- no

- yes

- yes


shargon:
Nice work! 👏👏

metachris:
Great! 👍

PR #227, opened 2018-02-16
Add migration flag to new wallets

dethos:
After the fix of wallet encryption, users will not be able to open old wallets with the new version of `neo-python`. This pull request adds a warning to the user explaining that the wallet needs to be "migrated" before continuing.

Used `SaveStoredData` and `LoadStoredData`

By testing it manually with old and new wallets.

**Did you add any tests?**
No

No

Yes

To a feature branch


localhuman:
Perfect!

PR #230, opened 2018-02-16
PR template update: please add to changelog

metachris:
Added a line to the PR template about adding something to CHANGELOG

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,5 @@
 
-**What current issue(s) does this address?, or what feature is it adding?**
-
-**How did you solve this problem?**
+**What current issue(s) does this address, or what feature is it adding? How did you solve this problem?**
 
 **How did you make sure your solution works?**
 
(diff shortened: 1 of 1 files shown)


metachris:
Good point, updated the PR.

localhuman:
Would it make sense to add a checklist to the template instead of a series of Y/N questions?

metachris:
I think it would. Updated the PR

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,12 +5,12 @@
...
-**Did you run `make lint` and `make test`?**
-
-**Are you making a PR to a feature branch or development rather than master?**
+**Please check the following, if applicable:**
 
-**Did you add an entry to `CHANGELOG.rst`? (if not, please do)**
+- [ ] Did you add any tests?
+- [ ] Did you run `make lint`?
+- [ ] Did you run `make test`?
+- [ ] Are you making a PR to a feature branch or development rather than master?
```
ixje: this question cannot be handled by a check box

metachris: yes, it can, it still is a yes/no question.

ixje: After reading it again I guess we do not differentiate between Dev vs feature?

ixje: ok nm

PR #232, opened 2018-02-16
Feature wallet changes

localhuman:

- fixes security issue with current wallet implementation
- all existing wallets will need to be run through the command `python reencrypt_wallet.py {wallet_file}`
- if you open a wallet that has not been migrated, you'll receive a warning.

See this PR for more information https://github.com/CityOfZion/neo-python/pull/214

Thanks to @dethos for identifying the issue and creating the fix!


metachris:
When opening an old wallet, it currently prints this message:

    neo> open wallet neo-privnet.wallet
    [password]> ***
    [E 180216 18:01:01 UserWallet:330] Could not get key Instance matching query does not exist:
        SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS t1 WHERE ("t1"."Name" = ?)
        PARAMS: ['MigrationState']
    Could not open wallet: This wallet is currently vulnerable. Please execute the "reencrypt_wallet.py" script on this wallet before continuing

I think it might be helpful to change or hide the first error message (Could not get key Instance matching query does not exist:), because it's not helpful, and only leave the latter.

metachris:
@dethos can you do this to finish implementation of this feature?

If noone has time for it, it would be okay for me to merge this as is too. Just think usability could be improved a little bit.

dethos:
@metachris those lines are produced by the `LoadStoredData` call ([here](https://github.com/CityOfZion/neo-python/blob/master/neo/Implementations/Wallets/peewee/UserWallet.py#L330)). Since they are "printed" by the exception handling, there isn't a way we can try to catch the exception outside the call. 

To avoid printing them on the console, I see 2 possibilities (neither of those are "pretty"). We can edit the logging and affect every call or we "duplicate" the method's code just for this case. 

If you think it is worthy, I can prepare a commit with the changes (will push it tomorrow 20-02-2018)

metachris:
I think it's probably not worth the time if we have to build a workaround that's annoying ;)

I am okay with merging this as is. 👍

ixje:
I seem to be getting the following error (note: I'm still on commit): https://github.com/CityOfZion/neo-python/commit/19d18f5d56a620317cf640ca030e1e9075dbcae0

```
[E 180220 21:12:53 UserWallet:330] Could not get key Instance matching query does not exist:
    SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS t1 WHERE ("t1"."Name" = ?)
    PARAMS: ['PasswordHash'] 
[E 180220 21:12:53 InvokeQueue:131] Password hash not found in database
      File "/home/middleware/InvokeQueue.py", line 148, in open_wallet
        self.wallet = UserWallet.Open(self.path, self.pw)
      File "/home/middleware/neo/Implementations/Wallets/peewee/UserWallet.py", line 124, in Open
        return UserWallet(path=path, passwordKey=password, create=False)
      File "/home/middleware/neo/Implementations/Wallets/peewee/UserWallet.py", line 43, in __init__
        super(UserWallet, self).__init__(path, passwordKey=passwordKey, create=create)
      File "/home/middleware/neo/Wallets/Wallet.py", line 104, in __init__
        raise Exception("Password ha
```

It seems related to this PR, but as far as i can see nothing from this PR made it into `Development` yet or did I miss anything?

PR #236, opened 2018-02-18
Fixed bug with SmartContract event handler sharing

brianlenz:

Event handlers were not being properly isolated between multiple SmartContract instances.

Initialize the event_handlers dictionary in the constructor instead of sharing the dictionary statically across all SmartContract instances.

Tested it. Events were firing multiple times with multiple SmartContracts previously. Now, they only fire for the proper script hash.

**Did you add any tests?**

No.

No.

Yes, but tests appear to be failing presently on `development`.

`development`

### DIFF
--- a/neo/contrib/smartcontract.py
+++ b/neo/contrib/smartcontract.py
@@ -40,11 +40,12 @@ class SmartContract:
     sent in the smart contract.
     """
     contract_hash = None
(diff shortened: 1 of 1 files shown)


localhuman:
👍

metachris:
👍

PR #241, opened 2018-02-20
Missing transaction tx, or None, as return of SC Invokes

vncoelho:
InvokeContract was not returning tx object or False status

The possibility of easy accessing the tx result of a "testinvoke" or "import contract" in the prompt.
Several users may use prompt.py as example.

Quite of. It is just an idea for better template for being used.

Not sure

**Did you add any tests?**
Manual and auto tests

no

no

no

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -716,10 +716,10 @@ class PromptInterface(object):
 
                 result = InvokeContract(self.Wallet, tx, fee)
 
-                return
+                return tx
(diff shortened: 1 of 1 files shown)


vncoelho:
Hey @metachris , thanks for checking with carefully.

I should be None and an extra verification should also verifies if False was returned when broadcasting to the Blockchain.

PR #245, opened 2018-02-21
Bug wallet sync bug password error

localhuman:
- newly created wallets had an issue syncing, so nothing would show up in `synced balances`.  this is fixed
- with new wallet encryption, some commands in prompt were not properly encoding the password and giving an 'incorrect password' error. this is fixed.
- also removed 'try .. catch' blocks for `BuildAndRun` command so that exceptions aren't swallowed and missed

- code

- running tests

**Did you add any tests?**
- fixed tests

- not really

- yes

- yes


metachris:
monster :)

PR #252, opened 2018-02-23
Improved Docker setup and documentation

metachris:
* created `/docker` directory with all Docker stuff inside
* 2 Dockerfiles now, one for dev and one for master
* docker-compose file to setup neoscan with a private network
* Documentation

Manual tests

**Did you add any tests?**

No

No

Yes

dev


docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+# to a private network, make sure the privatenet container is already running.
+# https://hub.docker.com/r/cityofzion/neo-privatenet
+#
+# Start a container interactively, opening a bash in `/neo-python`, and mounting the current directory as `/neo-python/sc`:
+#
+#    $ docker run --rm -it -v $(pwd):/neo-python/sc --net=host -h neopython --name neopython neopython /bin/bash
+#
+# Once you are inside the container, you can start neo-python with `python3 prompt.py -p` (using -p to connect to a private net).
+# To update neo-python, just run `git pull` and `pip install -e .`
+FROM ubuntu:16.04
```
revett: Why not use `python:3.5` as the base image? 

Benefits:

- Smaller image size
- No need to install Python dependencies
- Faster image build time

metachris: The reason is that the Python image need more space! I did test the builds for both `python:3.5` and `ubuntu:16.04` and the Python image uses 200MB more compared to Ubuntu.

revett:
@metachris Is it worth having a README within the `docker/` directory instead of adding documentation to the files themselves.

To me it hides the documentation by having them in the file - SHOW OFF the documentation 😊

docker/Dockerfile
revett: Ignore me, I thought it was the other way round 😂

docker/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+# Install dependencies
+RUN apt-get update && apt-get install -y \
+    wget \
+    git-core \
+    python3-dev \
```
revett: @metachris Do we need to define this as Python 3.5?

metachris: yes

metachris:
Yes, I think having a README is worth it -- it's context for people that click into the `docker` folder on GitHub, and not much effort to maintain. Makes it overall just a little easier for new people.

But perhaps we can make it shorter, perhaps even without the examples.

I do like having the documentation in the Dockerfiles, because then you can link to the files directly and have everything explained there.

revett:
@metachris Is it worth adding the neo-python Dockerfile to the Docker Compose stack?

metachris:
Sorry, not sure exactly what you mean -- which Docker Compose stack // what's the question in particular?

metachris:
Actually, not sure the Docker compose file belongs here.

localhuman:
@metachris you think this is good to go?

metachris:
I think so. We can delete the docker compose file later anytime too.

PR #254, opened 2018-02-24
Development

JorritvandenBerg:
It adds a command line arg to skip the confirmation step when using bootstrap.py. This comes handy when calling the script in a non-interactive fashion. 

Add the -s or --skipconfirm arg to bootstrap.py, which sets require_confirm to False if it is passed in the command line. Also I moved the confirmation warning to the code block where require_confirm is True. 

Tested at each code addition and did a final check by pulling and testing the code from my fork remote.

**Did you add any tests?**
No, the change is pretty straightforward and I used the existing arg parsing as an example.

Apart from printing a confirmation warning only when a confirmation is required, no changes.

'make lint' outputs 'pycodestyle neo examples prompt.py'
'make test' seems to get stuck at downloading the fixtures

The development branch. The master branch does not contain the require_confirm option yet.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -9,13 +9,22 @@ def main():
                         help="use MainNet instead of the default TestNet")
     parser.add_argument("-c", "--config", action="store", help="Use a specific config file")
 
-    parser.add_argument("-n", "--notifications", action="store_true", default=False, help="Bootstrap notification dataase")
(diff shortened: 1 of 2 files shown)


localhuman:
@JorritvandenBerg downloading the test fixtures is a 1GB file, so it takes quite a while to do.  Either way, this looks good.  Thanks!

metachris:
👍

PR #256, opened 2018-02-25
Support Synchronous Wallet Rebuilds

brianlenz:

There are situations where it can be useful to synchronously wait for the wallet rebuild process to complete before returning. We have a script that processes transactions in bulk. It needs to use a wallet to sign the transactions, but sometimes the wallet can become corrupted and result in "insufficient funds" errors. When this happens, we need to rebuild the wallet from a local copy. It's useful to have that rebuild process block until the wallet is fully rebuilt.

https://github.com/NarrativeNetwork/tokensale-neo-smartcontract/blob/master/util/bulk-process-tx.py#L295

Added an optional argument to allow wallet rebuilds be performed synchronously so that `ProcessBlocks` does not return until the wallet is fully rebuilt. We could alternatively handle this blocking behavior externally, but I think it may be useful to other devs integrating in a similar fashion to be able to leverage this optional parameter to synchronously block on the rebuild, as well.

Tested that it has no impact on existing behavior and only changes behavior when the new, optional parameter is used.

No.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Wallets/Wallet.py
+++ b/neo/Wallets/Wallet.py
@@ -655,7 +655,7 @@ class Wallet(object):
             block_limit (int): the number of blocks to process synchronously. defaults to 1000. set to 0 to block until the wallet is fully rebuilt.
         """
         blockcount = 0
-        while self._current_height <= Blockchain.Default().Height and (block_limit==0 or blockcount < block_limit):
+        while self._current_height <= Blockchain.Default().Height and (block_limit == 0 or blockcount < block_limit):
 
             block = Blockchain.Default().GetBlockByHeight(self._current_height)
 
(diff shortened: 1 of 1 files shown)


metachris:
Very nice, I really like this! Just wondering if 500 was a magic number by @localhuman, and whether changing that to 1000 by default has any unintended consequences.

brianlenz:
I didn't change that part; it was recently changed from 500 to 1000 here:

https://github.com/CityOfZion/neo-python/commit/3173f030def0629a850f80e4020ed52616c1be79#diff-f1c26fe160bdce08834e51e99d427cefR654

localhuman:
changed to 1000 to speed up wallet rebuilds, shouldn't have any adverse effects elsewhere.  

Ideally we'd have some tests surrounding this functionality, but I understand they may be a bit difficult to write.

PR #259, opened 2018-02-26
Development

JorritvandenBerg:
Informing the user that an insufficient funds error is possibly caused by an unsynced wallet and how to resolve that.

Extended the error message

Tested

No, just a string that has changed

- [ ] Did you add any tests?
- [x] Are you making a PR to a feature branch or development rather than master? <- development


metachris:
I think we could improve this PR by checking if the wallet is not synced, and only showing the error message then. What do you think about this?

JorritvandenBerg:
Updated, I added GetSyncState to the Wallet class to verify this and throw a warning if the wallet is not fully synced. Sorry, commits became a bit messy. I had to fiddle a bit to reproduce the case, which is not easy on a private, but on test net it often occurs. Below is a screenshot of the warning.

screen shot 2018-02-26 at 23 37 04

neo/Wallets/Wallet.py
```diff
@@ -1205,6 +1213,18 @@ def GetSyncedBalances(self):
...
 
+    def GetSyncedState(self):
+        """
+        Get synchronisation status of the wallet.
+
+        Returns:
+            bool: True if wallet is synced.
+        """
+        if (int(100 * self._current_height / Blockchain.Default().Height)) < 100:
+            return None
```
metachris: this should return `False`.

neo/Wallets/Wallet.py
```diff
@@ -1205,6 +1213,18 @@ def GetSyncedBalances(self):
                 balances.append((asset.symbol, self.GetBalance(asset)))
         return balances
 
+    def GetSyncedState(self):
```
metachris: I would also call this method `IsSynced()`

neo/Wallets/Wallet.py
```diff
@@ -1054,6 +1054,14 @@ def MakeTransaction(self,
 
         self._vin_exclude = None
 
+        synced_state = self.GetSyncedState()
```
metachris: rename this to `is_synced` pls

neo/Wallets/Wallet.py
```diff
@@ -1054,6 +1054,14 @@ def MakeTransaction(self,
...
+
+        if not synced_state:
+            logger.warn("Wait for your wallet to be synced before doing \
```
metachris: Don't add line breaks, please use this syntax for long strings:
 
    s = ("this is a very"
         "long string too"
         "for sure ...")

JorritvandenBerg:
Ah those did not trigger on my machine. Will try with the is_synced block wrapped in the check for unspents.

Also added a check if the Blockchain.Default().Height in IsSynced is not zero (as that causes a zero division error).

________________________________________

metachris:
@JorritvandenBerg please.resolve the Changeling conflict

localhuman:
I guess I'd prefer `IsSynced` to be an `@property`

Also, should we use `"""` for longer strings?

metachris:
Yes, let's make it an `@property` 👍 

In general you like this?

Re strings, I personally think this is fine and pretty standard, probably fine as is

JorritvandenBerg:
Adding the property decorator and running a last check. I noticed that during an initial sync of the chain, percent_synced is at 100. So the insufficient funds error is still triggered in that case. Do we want to work on that?

metachris:
Hm, could add another check and a warning if sync is 100 but chain is unsynced. could also do this in nother PR though.

PR #260, opened 2018-02-26
Fix Bug With Edge Case Gas Calc

brianlenz:

Transactions consuming GAS > 9 and <=10 fail currently due to no network fee being attached to the transaction.

Fixed the GAS consumed check so that it is aligned with the NEO core reference implementation:

https://github.com/neo-project/neo-gui/blob/master/neo-gui/UI/InvokeContractDialog.cs#L37
https://github.com/neo-project/neo-gui/blob/master/neo-gui/UI/InvokeContractDialog.cs#L153

See changeset notes for more details.

Tested it heavily before and after.

Before:
```
Used 9.585 Gas 

---------------------------------
Test invoke successful
Total operations: 3226
Results ['Integer: 7 ']
Invoke TX GAS cost: 0.0
Invoke TX fee: 0.0
---------------------------------
```

After:
```
Used 9.585 Gas 

---------------------------------
Test invoke successful
Total operations: 3226
Results ['Integer: 7 ']
Invoke TX GAS cost: 0.0
Invoke TX fee: 0.001
---------------------------------
```

No.

- [X] Did you add any tests?


metachris:
Very nice!

brianlenz:
This change is worthy of a test, so I'm going to add one shortly.

brianlenz:
OK, test added!

localhuman:
This is great, thanks!

PR #263, opened 2018-02-26
Ubuntu 17.x + Anaconda (python 3.5) installing

dmunicio:
Add instructions to install it in Ubuntu 17.x where Python 3.5 is not available by using Anaconda.

- [ ] Did you add any tests?

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -68,11 +68,36 @@ OSX
 Ubuntu/Debian
 """""""""""""
 
+Ubuntu 16.x or earlier
 ::
 
     apt-get install libleveldb-dev python3.5-dev python3-pip libssl-dev
 
+Ubuntu 17.x or later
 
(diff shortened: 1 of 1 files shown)


metachris:
Hey, thanks for the contribution.

Damn, is it really that hard to install python3.5 in the new Ubuntu's? We're working on 3.6 support though.

In general I think anaconda is a bit big to suggest as default setup, but if there's nothing easier it might also be okay. Else I think it could make sense as a chapter in the documentation (see the `/docs` directory), and just linking there from the README.

dmunicio:
I couldn't find an easier way... 
Maybe when 3.6 support is completed, we can remove that Anaconda part.
Also installation could be much easier if it could be installed as a python library with pip.

localhuman:
Thanks for the PR!

metachris:
@dmunicio <3

if you don't mind, perhaps you could create another PR that adds a link to the README somewhere at the Ubuntu setup, which links to this new section of the documentation.

PR #266, opened 2018-02-27
Allow custom network tx fee

brianlenz:

Gives developers flexibility to specify a custom tx `net_fee` when automating bulk transactions to save on gas costs (the default of 0.001 is a bit higher than needed).

Added a new `net_fee_override` parameter to the `test_invoke` and `test_deploy_and_invoke` methods (along with calling methods, as appropriate) to allow developers to set a custom fee to use.

Tested it before and after and also updated an existing test case to test that it works properly.

No.

- [X] Did you add any tests?


neo/Prompt/Commands/Invoke.py
```diff
@@ -210,7 +210,7 @@ def TestInvokeContract(wallet, args, withdrawal_tx=None, parse_params=True, from
     return None, None, None, None
 
 
-def test_invoke(script, wallet, outputs, withdrawal_tx=None, from_addr=None):
+def test_invoke(script, wallet, outputs, withdrawal_tx=None, from_addr=None, net_fee_override=None):
```
localhuman: Would be easier to do this:
`def test_invoke(script, wallet, outputs, withdrawal_tx=None, from_addr=None, net_fee_override=Fixed8.FromDecimal(.001)):`

Then you can get rid of the conditional below.

localhuman: Also may make sense to just use the kwarg name `net_fee`

localhuman:
This is very useful, great works!

neo/Prompt/Commands/Invoke.py
brianlenz: So, I actually implemented that way initially, but with 4 different methods having this arg (and thus specifying the default value), I opted for the `net_fee_override` approach instead so that the default of `.001` is limited to just the two places it's used. I'm happy to change it if you guys think that's preferable.

I might suggest instead that the value is just the float value, though, so that callers don't have to deal with calling `Fixed8.FromDecimal()`; it'll be easier to just pass in the float value directly and let the method handle the type conversion.

Thoughts?

brianlenz: Actually, I'll just make it use a constant. I think that'll make the most sense. Update incoming shortly...

brianlenz: I decided to go with the name `min_fee` for the kwarg to disambiguate it from the existing `net_fee` variable used in `test_invoke`. I also opted for your suggestion of defining the kwarg as `Fixed8` for consistency with other `fee` args.

Let me know if you think anything else needs tweaking! Thanks for the feedback 👍

neo/Prompt/Commands/Invoke.py
```diff
@@ -35,7 +35,7 @@
 from neo.VM.OpCode import *
 import json
 
-DEFAULT_MIN_FEE = Fixed8.FromDecimal(.001)
+DEFAULT_MIN_FEE = Fixed8.FromDecimal(.0001)
```
brianlenz: @localhuman, what's your level of confidence that this is a sufficient fee? In my experience, I had a hard time getting every relayed transaction accepted when I used `.0001` or even `.0002`. When running a large number of automated transactions, I'd see some transactions not go through successfully. Only once I bumped it up to `.0004` was I able to consistently get all relayed transactions to get processed successfully.

PR #267, opened 2018-02-27
Add Full JSON-RPC Param Type Support

brianlenz:
* Updated the JSON-RPC implementation so that it supports setting the `ContractParameterType` consistently with the reference implementation. This includes not only the current hex representation, but also int value, string literal, and int string.

* Added appropriate tests to ensure each of these permutations is functional.

Not all of the usual JSON-RPC contract param types are supported currently. The current implementation only supports hex-encoded strings. This change adds support for the various types that are supported in the NEO reference implementation, including int (e.g. `7`), string int (e.g. `"7"`, and string literal (e.g. `"String"`).

Updated the parser to support auto-detecting the type.

Tested it and added a test for each new variation that is supported.

No.

- [X] Did you add any tests?


localhuman:
👍

metachris:
👍

PR #270, opened 2018-02-27
Move to Python 3.6+ and integration with new version of neo-boa

localhuman:

- move to only python 3.6+ support
- use new version of compiler ( neo-boa==0.3.3 ) based on python 3.6 wordcode
- full testing of VM and all compiled smart contracts
- adds new command `TestBuild` for running tests of compiled contracts

- wrote new version of compiler for python 3.6+
- added a lot of tests
- updated travis

- tests

- This is not compatible with Python 3.5.  You will need to delete your virtual environment and create a new one with Python 3.6 or greater, then re-install `requirements.txt`

- [x] Did you add any tests?


requirements_docs.txt
```diff
@@ -38,6 +38,7 @@ mock==2.0.0
 mpmath==1.0.0
 neo-boa==0.2.1
 neo-python-rpc==0.1.8
+neo-core==0.3.6
```
metachris: `neo-core` doesn't exist, the package is called `neocore`. Why do we need this in the docs requirements, actually?

requirements_docs.txt
```diff
@@ -9,7 +9,7 @@ base58==0.2.5
 bitcoin==1.1.42
 blessings==1.6.1
 bpython==0.17
-byteplay3==3.5.0
+bytecode==0.5
```
metachris: Why do we need this in the docs requirements?

metachris:
If we merge this into master, we should definitely do a minor version bump!

requirements_docs.txt
localhuman: I think at some point we would maybe be using the compiler when generating the docs?

metachris: Not sure about the use-case, but sure :D

metachris:
woo hoo

Issue #233, opened 2018-02-16
JsonRpcApi input & output differs from C#


ixje:
**issue 1**
C# accepts single quotes
```
{'params': ["ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9", "symbol"], 'id': '2', 'jsonrpc': '2.0', 'method': 'invokefunction'}
```
[JsonRpcApi](https://github.com/CityOfZion/neo-python/blob/development/examples/json-rpc-api-server.py) does not and will have an error saying:
```
{
    "jsonrpc": "2.0",
    "error": {
        "message": "Parse error",
        "code": -32700
    },
    "id": null
}
```
The actual error message is dropped and says
```
Expecting property name enclosed in double quotes: line 1 column 2 (char 1)
```

**issue 2**
C# returns full names for the VMstate and Stack types after execution e.g.
```
http://seed3.cityofzion.io:8080  (2.7.0 node)

{'params': ["ecc6b20d3ccac1ee9ef109af5a7cdb85706b1df9", "symbol"], 'id': '2', 'jsonrpc': '2.0', 'method': 'invokefunction'}

{
    "jsonrpc": "2.0",
    "id": "2",
    "result": {
        "script": "00c10673796d626f6c67f91d6b7085db7c5aaf09f19eeec1ca3c0db2c6ec",
        "state": "HALT, BREAK",    <-----
        "gas_consumed": "0.141",
        "stack": [
            {
                "type": "ByteArray", <------
                "value": "525058"
            }
        ]
    }
}
```

JsonRpcApi gives VM state and stack type as (string) numbers to indicate the same. Note: the only difference in the calls is the scripthash I'm talking to.
```
{"id": "2", "method": "invokefunction", "params": ["584660c663f114d803928b77cb9abdd585a0fc17", "symbol"], "jsonrpc": "2.0"}

{
    "jsonrpc": "2.0",
    "result": {
        "script": "303063313036373337393664363236663663363731376663613038356435626439616362373738623932303364383134663136336336363034363538",
        "gas_consumed": "0.372",
        "stack": [
            {
                "value": "50434731",
                "type": "05" <--------
            }
        ],
        "state": 5 <-------
    },
    "id": "2"
}
```

ixje:
I need to add one more difference in representation

C#
```
[...]

PR #272, opened 2018-02-28
JSON-RPC output fixes

brianlenz:

Fixes JSON-RPC output issues from #233.

Implemented appropriate fixes for each of the issues raised.

Tested JSON-RPC output and updated existing tests to account for new formats.

No.

- [ ] Did you add any tests?


localhuman:
👍

PR #279, opened 2018-03-01
Add num_peers to REST API

brianlenz:

Added a `num_peers` property to the `/status` endpoint of the REST API.

Fixed traceback error for script hash that doesn't exist for `/tokens` endpoint.

Implemented.

Tested.

No.

- [ ] Did you add any tests?


metachris:
👍

PR #280, opened 2018-03-02
Docs for Python 3.6 update

metachris:
* Updated README.md and docs/install.rst with infos about setting up the new Python 3.6 system, in particular relevant because Ubuntu <16.10 (eg. 16.04, the latest LTS) does not support python-3.6 out of the box.
* Also updated the Docker files


MediaServe:
Using the third party backport works. When using a third party repository we should add a disclaimer because we cannot verify the legitimacy of this source.

metachris:
Added a disclaimer, contributed by Thomas Lobker.

PR #281, opened 2018-03-03
API syslog / logger improvements

brianlenz:

Logging for neo-python API server was disjointed. Some logs would go to stdout and others to a log file and stderr. This change allows you to use syslog (and only syslog) for logging if you choose, which gives sys admins more control over how to handle log files, rotation, etc.

Upgraded to 1.4.0 of logzero and added support for 

Tested it and deployed it on our API nodes.

No.

- [ ] Did you add any tests?

### DIFF
--- a/api-server.py
+++ b/api-server.py
@@ -14,10 +14,12 @@ import argparse
 import threading
 from time import sleep
 
+import logzero
 from logzero import logger
 from twisted.internet import reactor, task, endpoints
 from twisted.web.server import Site
 from klein import Klein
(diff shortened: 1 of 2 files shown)


localhuman:
Would be interested in what @metachris thinks!

metachris:
I think it's good stuff. Would try to simplify things a little more, perhaps add some of it to neo.api utils. I also think we should rename the `-q` argument to `--disable-stderr` since it doesn't really make things quiet. I'll give it a more in-depth review a bit later today.

PR #283, opened 2018-03-04
Api server logging updates

metachris:
Updated api server with updates from https://github.com/CityOfZion/neo-python/pull/281 by @brianlenz and more customized and simplified logging setup.

### DIFF
--- a/api-server.py
+++ b/api-server.py
@@ -10,7 +10,6 @@ See also:
 * JSON-RPC api issues: https://github.com/CityOfZion/neo-python/issues/273
 """
 import os
-import sys
 import syslog
 import argparse
 import threading
@@ -146,6 +145,10 @@ def main():
     # Write a PID file to easily quit the service
     write_pid_file()
 
+    # Setup twisted / klein to reuse the logzero setup
+    observer = log.PythonLoggingObserver(loggerName=logzero.LOGZERO_DEFAULT_LOGGER)
+    observer.start()
+
     # Instantiate the blockchain and subscribe to notifications
     blockchain = LevelDBBlockchain(settings.LEVELDB_PATH)
     Blockchain.RegisterBlockchain(blockchain)
@@ -201,10 +204,6 @@ class ApiKlein(Klein):
         if syslog_facility is not None:
             facility = translate_syslog_facility(syslog_facility)
             startLogging(prefix="pyapi", facility=facility)
-
-        # Always reuse the api-server logzero setup, whether stdout or file
-        observer = log.PythonLoggingObserver(loggerName=logzero.LOGZERO_DEFAULT_LOGGER)
-        observer.start()
 
         if not endpoint_description:
             endpoint_description = "tcp:port={0}:interface={1}".format(port, host)
--- a/api-server.py
+++ b/api-server.py
@@ -3,27 +3,43 @@
 This api server runs one or both of the json-rpc and rest api. Uses
 neo.api.JSONRPC.JsonRpcApi and neo.api.REST.NotificationRestApi
 
-See also:
+See also
 
 * Tutorial on setting up an api server: https://gist.github.com/metachris/2be27cdff9503ebe7db1c27bfc60e435
 * Example systemd service config: https://gist.github.com/metachris/03d1cc47df7cddfbc4009d5249bdfc6c
 * JSON-RPC api issues: https://github.com/CityOfZion/neo-python/issues/273
+
+Logging
+-------
+
+This api-server can log to stdout/stderr, logfile and syslog.
+Check `api-server.py -h` for more details.
+
+Twisted uses a quite custom logging setup. Here we simply setup the Twisted logger
+to reuse our logzero logging setup. See also:
+
+* http://twisted.readthedocs.io/en/twisted-17.9.0/core/howto/logger.html
+* https://twistedmatrix.com/documents/17.9.0/api/twisted.logger.STDLibLogObserver.html
 """
 import os
 import syslog
 import argparse
 import threading
 from time import sleep
-from twisted.python import log
-from twisted.python.syslog import startLogging
+from logging.handlers import SysLogHandler
 
 import logzero
 from logzero import logger
+
+# Twisted logging
+from twisted.logger import STDLibLogObserver, globalLogPublisher
+from twisted.python.syslog import startLogging
+
+# Twisted and Klein methods and modules
 from twisted.internet import reactor, task, endpoints
 from twisted.web.server import Site
-from klein import Klein
-from logging.handlers import SysLogHandler
 
+# neo methods and modules
 from neo import __version__
 from neo.Core.Blockchain import Blockchain
 from neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain import LevelDBBlockchain
@@ -119,35 +135,45 @@ def main():
         settings.setup_coznet()
 
     syslog_facility = None
-    if args.syslog or args.syslog_local:
+    if args.syslog or args.syslog_local is not None:
         # Setup the syslog facility
         if args.syslog_local:
-            print("Logging to syslog local facility %s" % args.syslog_local)
+            print("Logging to syslog local%s facility and stdout" % args.syslog_local)
             syslog_facility = SysLogHandler.LOG_LOCAL0 + args.syslog_local
         else:
-            print("Logging to syslog user facility")
+            print("Logging to syslog user facility and stdout")
             syslog_facility = SysLogHandler.LOG_USER
 
         # Setup logzero to only use the syslog handler
         logzero.logfile(None, disableStderrLogger=args.disable_stderr)
         syslog_handler = SysLogHandler(facility=syslog_facility)
         logger.addHandler(syslog_handler)
+
+        # Setup the twisted syslog facility (TODO: refactor to have logzero use syslog?)
+        facility = translate_syslog_facility(syslog_facility)
+        startLogging(prefix="pyapi", facility=facility)
     else:
         # Setup file logging
         if args.logfile:
             logfile = os.path.abspath(args.logfile)
-            print("Logging to logfile and stdout: %s" % logfile)
+            if args.disable_stderr:
+                print("Logging to logfile: %s" % logfile)
+            else:
+                print("Logging to stderr and logfile: %s" % logfile)
             logzero.logfile(logfile, maxBytes=LOGFILE_MAX_BYTES, backupCount=LOGFILE_BACKUP_COUNT, disableStderrLogger=args.disable_stderr)
 
         else:
             print("Logging to stdout and stderr")
 
+    # Disable logging smart contract events
+    settings.set_log_smart_contract_events(False)
+
     # Write a PID file to easily quit the service
     write_pid_file()
 
-    # Setup twisted / klein to reuse the logzero setup
-    observer = log.PythonLoggingObserver(loggerName=logzero.LOGZERO_DEFAULT_LOGGER)
-    observer.start()
+    # Setup Twisted and Klein logging to use the logzero setup
+    observer = STDLibLogObserver(name=logzero.LOGZERO_DEFAULT_LOGGER)
+    globalLogPublisher.addObserver(observer)
 
     # Instantiate the blockchain and subscribe to notifications
     blockchain = LevelDBBlockchain(settings.LEVELDB_PATH)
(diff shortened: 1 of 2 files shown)


metachris:
Please don't merge before feedback from @brianlenz

api-server.py
```diff
@@ -140,8 +185,57 @@ def main():
...
+
+        if not endpoint_description:
+            endpoint_description = "tcp:port={0}:interface={1}".format(port, host)
+
+        endpoint = endpoints.serverFromString(reactor, endpoint_description)
+        endpoint.listen(Site(self.resource()))
+        reactor.run()
+
+
+def translate_syslog_facility(syslog_facility):
```
metachris: @brianlenz this looks like something that could be very useful in `logzero` itself. what do you think?

api-server.py
```diff
@@ -63,21 +66,30 @@ def custom_background_code():
...
+    group_modes = parser.add_argument_group(title="Mode(s)")
+    group_modes.add_argument("--port-rpc", type=int, help="port to use for the json-rpc api (eg. 10332)")
+    group_modes.add_argument("--port-rest", type=int, help="port to use for the rest api (eg. 80)")
+
+    # Advanced logging setup
+    group_logging = parser.add_argument_group(title="Logging options")
+    group_logging.add_argument("--logfile", action="store", type=str, help="Logfile")
+    group_logging.add_argument("--syslog", action="store_true", help="Log to syslog instead of to log file ('user' is the default facility)")
+    group_logging.add_argument("--syslog-local", action="store", type=int, choices=range(0, 7), metavar="[0-7]", help="Log to a local syslog facility instead of 'user'. Value must be between 0 and 7 (e.g. 0 for 'local0').")
+    group_logging.add_argument("--disable-stderr", action="store_true", help="Disable stderr logger")
```
brianlenz: The reason I had done `-q` is because that's how most *nix utilities work. It prevents output (stdout typically) by making the process run silently. But, I do see how this is different in that it's a daemon/server process, and those typically don't have any kind of quiet flag at all. What you've done here works :)

I do wonder, though: why do we log to stderr in the first place? It seems like it would make more sense to log to stdout by default. The only reason we default to stderr is because of logzero's default behavior, right?

api-server.py
```diff
@@ -63,21 +66,30 @@ def custom_background_code():
...
+    # Ports for RPC and REST api
+    group_modes = parser.add_argument_group(title="Mode(s)")
+    group_modes.add_argument("--port-rpc", type=int, help="port to use for the json-rpc api (eg. 10332)")
+    group_modes.add_argument("--port-rest", type=int, help="port to use for the rest api (eg. 80)")
+
+    # Advanced logging setup
+    group_logging = parser.add_argument_group(title="Logging options")
+    group_logging.add_argument("--logfile", action="store", type=str, help="Logfile")
+    group_logging.add_argument("--syslog", action="store_true", help="Log to syslog instead of to log file ('user' is the default facility)")
+    group_logging.add_argument("--syslog-local", action="store", type=int, choices=range(0, 7), metavar="[0-7]", help="Log to a local syslog facility instead of 'user'. Value must be between 0 and 7 (e.g. 0 for 'local0').")
```
brianlenz: 👍

I'm a relative newbie to python, so not as familiar with these standard utilities 😁

api-server.py
```diff
@@ -31,14 +36,12 @@
...
-LOGFILE_FN = os.path.join(DIR_PROJECT_ROOT, 'api-server.log')
+# Logfile default settings (only used if --logfile arg is used)
 LOGFILE_MAX_BYTES = 5e7  # 50 MB
 LOGFILE_BACKUP_COUNT = 3  # 3 logfiles history
-settings.set_logfile(LOGFILE_FN, LOGFILE_MAX_BYTES, LOGFILE_BACKUP_COUNT)
 
-# Set the PID file
-PID_FILE = "/tmp/neopython-api-server.pid"
+# Set the PID file, possible to override with env var PID_FILE
+PID_FILE = os.getenv("PID_FILE", "/tmp/neopython-api-server.pid")
```
brianlenz: 👍

api-server.py
```diff
@@ -101,9 +118,37 @@ def main():
     elif args.coznet:
         settings.setup_coznet()
 
+    syslog_facility = None
+    if args.syslog or args.syslog_local:
```
brianlenz: I had originally made `--syslog-local` only apply if you also applied `--syslog` intentionally, but this makes it a bit cleaner to use in that you can use one or the other (and it still won't hurt to use both) 👍

api-server.py
```diff
@@ -140,8 +185,57 @@ def main():
         endpoint_rest = "tcp:port={0}:interface={1}".format(args.port_rest, host)
         endpoints.serverFromString(reactor, endpoint_rest).listen(Site(api_server_rest.app.resource()))
 
-    app = Klein()
-    app.run(host, 9999)
+    app = ApiKlein()
+    app.run(host, 9999, syslog_facility=syslog_facility)
+
+    # After the reactor is stopped, gracefully shutdown the database.
+    logger.info("Shutting down - closing databases...")
```
brianlenz: 👍

api-server.py
```diff
@@ -140,8 +185,57 @@ def main():
...
+
+
+def translate_syslog_facility(syslog_facility):
+    """
+    SysLogHandler's facility is on a completely different scale than syslog, so
+    this method translates between the two
+    :param syslog_facility: the syslog facility value used by SysLogHandler
+    :return: the syslog facility value used by syslog (and thus Klein's logger)
+    """
+    mapping = {
```
brianlenz: So much nicer! 😂 👍

api-server.py
brianlenz: Yeah, it could be if we want to extend `logzero` to support logging to syslog instead. That makes a lot of sense, so it could be that we refactor this to be more embedded into `logzero` (where it's also reusable) as opposed to here in the API implementation (which I admittedly wasn't a huge fan of in the first place).

Is that where you were going?

api-server.py
```diff
@@ -140,8 +185,57 @@ def main():
...
+    app.run(host, 9999, syslog_facility=syslog_facility)
+
+    # After the reactor is stopped, gracefully shutdown the database.
+    logger.info("Shutting down - closing databases...")
+    NotificationDB.close()
+    Blockchain.Default().Dispose()
+    NodeLeader.Instance().Shutdown()
+
+
+class ApiKlein(Klein):
```
brianlenz: The reason I had to create `ApiKlein` is because the standard `Klein.run()` method gives you no way to customize the logging behavior. It forces you to use _either_ `stdout` or a log file. We needed more control over logging, which is why I extended the class. If there is an alternative approach you can think of, great! Let me know...

I don't love doing this kind of thing (overriding and copying 75% of the implementation), but we're beholden to how `Klein` is designed...

api-server.py
metachris: Actually I don't know why logging to stderr is enabled in first place. It must be because of some Python logging module internals that have this setup as default, I don't recall adding anything about this to logzero.

api-server.py
metachris: I think we take either this way, or show an error if only using `--syslog-local` and not `--syslog`

api-server.py
metachris: Understand! Perhaps we should think of taking that into the `neo.api` utilities somewhere

api-server.py
metachris: I think it would make sense, it's useful for many applications.

api-server.py
brianlenz: I think we can leave it and err on the side of flexibility.

api-server.py
brianlenz: This is where it comes from:

https://github.com/metachris/logzero/blob/master/logzero/__init__.py#L139

`logging.StreamHandler` defaults to `stderr` if no `stream` is provided. Don't ask me why they thought defaulting to `stderr` was [the right approach](https://docs.python.org/2/howto/logging.html#advanced-logging-tutorial)? 😬

api-server.py
```diff
@@ -140,8 +185,57 @@ def main():
...
+    """
+    ApiKlein extends Klein so that the logging behavior can be customized. Aside from logging,
+    the implementation is identical to Klein.run(): https://github.com/twisted/klein/blob/master/src/klein/_app.py#L376
+    """
+    def run(self, host=None, port=None, endpoint_description=None, syslog_facility=None):
+        if syslog_facility is not None:
+            facility = translate_syslog_facility(syslog_facility)
+            startLogging(prefix="pyapi", facility=facility)
```
brianlenz: With the approach you have here now, we will only call `startLogging` when logging to `syslog`. Shouldn't we be calling `startLoggingWithObserver` and pass in the `observer` created above? I'm actually wondering if we do that, do we even need all of this specialized `syslog_facility` and `translate_syslog_facility` at all? If the `PythonLoggingObserver` handles unifying the logging, it seems like that could unify the logs entirely and simplify this quite a bit?

api-server.py
brianlenz: I discovered a bug here in testing. With my original implementation, the type of `syslog_local` was a String, so this check would pass for the string `"0"`. Now that you've (properly) fixed it to parse the parameter as an `int`, this check no longer passes since the int `0` evaluates to false. Thus, this condition and the one just below need to be `args.syslog_local is not None`.

I don't think I have commit access to the neo-python repo, so I can't push the fix (and it seems like way overkill to create a branch on my repo to create yet-another PR from).

api-server.py
```diff
@@ -101,9 +118,37 @@ def main():
     elif args.coznet:
         settings.setup_coznet()
 
+    syslog_facility = None
+    if args.syslog or args.syslog_local:
+        # Setup the syslog facility
+        if args.syslog_local:
+            print("Logging to syslog local facility %s" % args.syslog_local)
```
brianlenz: I'd suggest changing this to:

```
print("Logging to syslog local%s facility" % args.syslog_local)
```
for clarity.

api-server.py
```diff
@@ -101,9 +118,37 @@ def main():
...
+
+        # Setup logzero to only use the syslog handler
+        logzero.logfile(None, disableStderrLogger=args.disable_stderr)
+        syslog_handler = SysLogHandler(facility=syslog_facility)
+        logger.addHandler(syslog_handler)
+    else:
+        # Setup file logging
+        if args.logfile:
+            logfile = os.path.abspath(args.logfile)
+            print("Logging to logfile and stdout: %s" % logfile)
```
brianlenz: This message isn't technically accurate for two reasons:

1) It's stderr, not stdout.

2) You can disable stderr, in which case only the logfile is used.

api-server.py
```diff
@@ -101,9 +118,37 @@ def main():
...
+
+        else:
+            print("Logging to stdout and stderr")
+
     # Write a PID file to easily quit the service
     write_pid_file()
 
+    # Setup twisted / klein to reuse the logzero setup
+    observer = log.PythonLoggingObserver(loggerName=logzero.LOGZERO_DEFAULT_LOGGER)
+    observer.start()
```
brianlenz: I'm not fully understanding how this is working (or supposed to work). In my testing, the Klein logging still isn't going to the logfile (when logfile is enabled). I see the normal blockchain logging in the logfile, but nothing from Klein.

api-server.py
brianlenz: I think the necessity of this refactor is contingent on the outcome of the other discussions above, so let's revisit this once we've got the rest of the behavior squared away?

metachris: 👍

api-server.py
metachris: No idea... guess we should keep that in logzero as it is though... to confirm with the Python logging module.

api-server.py
metachris: Oh right, got it. Updating right now for this. You could just make an external PR to the api-server branch, which would open a PR to this PR ;)

api-server.py
metachris: 👍

api-server.py
metachris: 👍

api-server.py
metachris: I think it would be a good approach to setup logzero to use the SyslogHandler, and just have the same setup for Twisted everywhere, to simply use the logzero logger setup.

api-server.py
metachris: Added a commit that should fix it. Would be great if you could test again.

api-server.py
brianlenz: Sounds good :)

PR #292, opened 2018-03-06
Update Dockerfile with command line parameters including bootstrapping option

JorritvandenBerg:
Users having trouble to bootstrap a test chain. This is an update to the Dockerfile, which enables the use of prompt.py and bootstrap.py command line parameters in Docker and introducing an additional --bootstrap parameter. This parameter allows users to bootstrap prior to starting the prompt.

Creating a bash script (entrypoint.sh) for this procedure.

Tested various combinations of parameters and created a fallback for combinations that don't work together as per the comments in prompt.py/bootstrap.py. But please experiment with to iron out any edge cases.

As an ENTRYPOINT is defined in the Docker file, one can not enter a bash shell anymore in the previous way. I have updated the example in the README and Dockerfile to explain how to open a bash shell with --entrypoint.

- [ ] Did you add any tests?
- [] Did you run `make test`? < still running, I'm on a very slow internet connection here. Will comment when done, but I don't expect any problems as the Python codebase hasn't changed.
- [x] Are you making a PR to a feature branch or development rather than master? < development


metachris:
I like the general idea. Will give it a review later today. :+1:

JorritvandenBerg:
Thanks! Tests failed due to a problem with an import from neo boa. @localhuman is working on a fix for that.

docker/Dockerfile
```diff
@@ -56,5 +60,15 @@ RUN pip3 install -e .
 # Download the privnet wallet, to have it handy for easy experimenting
 RUN wget https://s3.amazonaws.com/neo-experiments/neo-privnet.wallet
 
-# Example run command
-CMD /bin/bash
+# Remove /neo-python/Chains/SC234 directory at build time 
+RUN rm -rf /neo-python/Chains/SC234/
```
metachris: this is not needed, at the beginning the chain is not there

docker/Dockerfile
```diff
@@ -56,5 +60,15 @@ RUN pip3 install -e .
...
+
+# Copy entrypoint.sh script to image and make it executable
+COPY ./entrypoint.sh /entrypoint.sh
+RUN chmod +x /entrypoint.sh
+
+# Define the entrypoint
+ENTRYPOINT ["/entrypoint.sh"]
+
+# Empty command (defaults to running prompt.py without parameters)
+CMD []
```
metachris: this is not needed when ENTRYPOINT exists

docker/README.md
```diff
@@ -29,11 +29,47 @@ See also https://hub.docker.com/r/cityofzion/neo-privatenet
...
+
+    $ docker volume create my-chains-vol
+    $ docker run -it -v my-chains-vol:/neo-python/Chains neopython -m --bootstrap
+
+To list and remove volumes
+    $ docker volume ls
+    $ docker volume rm my-vol
+
+The Docker image supports the following command line parameters:
+-m / --mainnet
```
metachris: this is not well formatted

docker/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+PROMPT_OPTS=()
+BOOTSTRAP_OPTS=()
+BOOTSTRAP=""
+MAIN_NET=""
```
metachris: I feel this file is overly complex. 

 Also some of the vars are not yet used, eg `MAIN_NET`.

docker/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+echo "$1"
+
+while [ -n "$1" ]; do
+    case "$1" in
+        -m|--mainnet)
+            PROMPT_OPTS+="-m"
+            BOOTSTRAP_OPTS+="-m"
+            MAIN_NET="-m"
+        ;;
+        -p|--privnet)
```
metachris: this should throw an error in the bootstrap step, else it downloads the testnet chain

docker/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+            PROMPT_OPTS+="-p"
+            PRIV_NET="-p"
+        ;;
+        --coznet)
+            PROMPT_OPTS+="--coznet"
```
metachris: this should throw an error in the bootstrap step, else it downloads the testnet chain

docker/entrypoint.sh
```diff
@@ -0,0 +1,68 @@
...
+            COZ_NET="--coznet"
+        ;;
+        -c|--config)
+            if [[ "$MAIN_NET" == "-m" ]]; then
+                echo "Can not use config file on main net, proceeding without config file.."
```
metachris: uargh, same also goes for other combinations

metachris:
First of all, thanks for the effort, it's something I generally like -- making it easier for people to bootstrap the chain.

What I don't like is how complex the `entrypoint.sh` script is, and that it doesn't yet work with a variety of edge-cases and that if has some weird behaviour (eg. downloading the testnet chain when running with `-p`).

I feel there should be a simpler way to accomplish what you want to achieve.

Can you define the target users of this, and the specific use-cases? 

Also, why not have people log in with bash and run bootstrap manually?

If you want to keep working on this, I'm happy to provide more feedback and guidance. We can start merging things into a docker-bootstrap branch and later on merge things into development once it is ready.

docker/Dockerfile
```diff
@@ -20,7 +20,11 @@
 #
 # Start a container interactively, opening a bash in `/neo-python`, and mounting the current directory as `/neo-python/sc`:
 #
-#    $ docker run --rm -it -v $(pwd):/neo-python/sc --net=host -h neopython --name neopython neopython /bin/bash
+#    $ docker run --rm -it --entrypoint "/bin/bash" -v $(pwd):/neo-python/sc --net=host -h neopython --name neopython neopython
```
metachris: I'm not sure about this -- making the default usage more complex. Perhaps it would be better to leave the default entrypoint as is, and use the `--entrypoint` arg when running the bootstrap version,

JorritvandenBerg:
Hi Chris,
Thanks for reviewing. I have updated the bash script to fix the edge cases, added error handling and removed the docker ENTRYPOINT. Leaving the choice on wether to run /bin/bash or /entrypoint.sh to the user and not changing the default behaviour. 

The usecase is the ease of bootstrapping and starting prompt.py right away from the Docker CLI / command in a docker-compose.yml. Personally I hardly ever use docker run, let alone running a shell process on a container. I have created a docker-bootstrap branch on my fork, but can not merge this as the branch does not exist on the upstream CoZ repo. Could you create the branch?

localhuman:
I'll defer to @metachris on this one

metachris:
Just a note -- this PR is not forgotten, I just need a little bit to find some time to play around with a better Docker setup myself. I'll get to it a bit later this week :)

JorritvandenBerg:
Thanks, no worries. Let me know if you need any help.

metachris:
Just for reference: the Docker compose tooling and setup seems to be focusing here in the future: https://github.com/CityOfZion/neo-local

metachris:
Closing this because Docker setup moved to https://github.com/CityOfZion/neo-local. Please consider starting this feature/discussion there 👍

PR #293, opened 2018-03-06
Rename NotificationRestApi to RestAPi

metachris:

RestApi is a more fitting name, because this thing is already evolving to be the standard REST API for notifications and much more ^^

Rename

Test

No

- [ ] Did you add any tests?


metachris:
@localhuman - good to go or other thoughts?

PR #294, opened 2018-03-07
Sc debug improvements

ixje:
The following was discussed on discord, this implements the first 2 items, and I can push the 3rd item to this PR once we agree on the implementation of 1+2.

> During the dApp competition i made some local changes to make debugging smart contracts easier and I'm thinking about what can/should flow back. Would love some input on the following ideas;
> 
> 1) add instruction counter to the error message/stacktrace. 
> Rationale: it allows for setting a conditional breakpoint exactly on the instruction that's going to fail next. The counter also gives some indication whether it failed early during execution or later on (giving a feel how deep the error lays). 
> 
> 2) add a debug toggle to emit Notify events even if smart contracts fail. 
> Rationale: right now Notify events are only emitted when the SC execution is successful, which makes sense from an execution and operational standpoint, but not from developing. It drops a lot of potential useful information emitted in your Notify events that could help solve the problem.  
> 
> 3) add instruction logging
> rationale; having a historical log of all calls executed gave me nearly all the information needed to exactly pin point which code line of my SC the VM was executing (by looking at function calls, DROP/RET statements etc).
> 
> 4) add additional error messages to ExecutionEngine. 
> rationale; there's a lot of error feedback we can simply give that we're not doing leading to guess work or a lot of stepping through the VM internals.

> localhuman - Yesterday at 2:55 PM
> I like all of those things :smiley:

implement 1 + 2

manual testing

no

- [ ] Did you add any tests?

Will update `changelog.rst` after it's clear which changes will all be approved. If I check it now, I/we might forget to add all changes.

### DIFF
--- a/neo/SmartContract/StateReader.py
+++ b/neo/SmartContract/StateReader.py
@@ -21,10 +21,10 @@ from neocore.Cryptography.ECCurve import ECDSA
 from neo.SmartContract.TriggerType import Application, Verification
 
 from neo.VM.InteropService import StackItem, stack_item_to_py
+from neo.Settings import settings
 
 
 class StateReader(InteropService):
-
     notifications = None
 
     events_to_dispatch = []
@@ -285,6 +285,15 @@ class StateReader(InteropService):
         )
 
         self.notifications.append(args)
+
+        if settings.emit_notify_events_on_sc_execution_error:
+            # emit Notify events even if the SC execution might fail.
+            tx_hash = engine.ScriptContainer.Hash
+            height = Blockchain.Default().Height
+            success = None
+            self.events_to_dispatch.append(NotifyEvent(SmartContractEvent.RUNTIME_NOTIFY, args.State,
+                                                       args.ScriptHash, height, tx_hash,
+                                                       success, engine.testMode))
 
         return True
 
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -67,6 +67,9 @@ class SettingsHolder:
     # Logging settings
     log_smart_contract_events = False
 
+    # Emit Notify events when smart contract execution failed. Use for debugging purposes only.
+    emit_notify_events_on_sc_execution_error = False
+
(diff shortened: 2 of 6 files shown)


neo/SmartContract/StateReader.py
```diff
@@ -286,6 +286,15 @@ def Runtime_Notify(self, engine):
...
 
+        if settings.emit_notify_events_on_sc_execution_error:
+            # emit Notify events even if the SC execution might fail.
+            tx_hash = engine.ScriptContainer.Hash
+            height = Blockchain.Default().Height
+            success = None
+            self.events_to_dispatch.append(NotifyEvent(SmartContractEvent.RUNTIME_NOTIFY, args.State,
+                                                       args.ScriptHash, height, tx_hash,
+                                                       success, engine.testMode))
+
```
ixje: I've chosen to add event dispatching here over the [alternative location](https://github.com/CityOfZion/neo-python/blob/master/neo/SmartContract/StateReader.py#L220) because this solution makes it so that `Notify` and `Log` events are emitted in order, as opposed to out of order when using the alternative solution. I believe we have to assume people mix `Notify`, `Log` and `Print` statements in their SC code for whatever good reason they might have.

To elaborate take the following SC:
```
def Main(args):
    ctr = 0
    limit = 3
    Notify("Start main")

    while ctr < limit:
        Log(ctr)
        ctr += 1

    Notify("End main")
```

What you want as output is (in order):

> Notify - Start main
> Log - 0
> Log - 1
> Log - 2
> Notify - End Main

What you get in the alternative solution is:
> Log - 0
> Log - 1
> Log - 2
> Notify - Start main
> Notify - End Main

localhuman:
You can use the new `TestBuild` command to add some tests for new notification functionality and error messaging :) 
This test build catches the notifications https://github.com/CityOfZion/neo-boa/blob/master/boa_test/tests/test_event.py

PR #296, opened 2018-03-07
prompt.py -v/--verbose argument: show sc-events

metachris:

Added `-v/--verbose` argument to `prompt.py`: show sc-events

Add flag in prompt.py

Manual tests

No

- [ ] Did you add any tests?


localhuman:
I'd add that you can make it always start with verbose events by changing the setting in your `protocol.x.json` as well

metachris:
@localhuman how would you do that?

localhuman:
Actually now that I look, I was wrong about that.  Though adding it to the *.json and `settings.py` would not be hard

PR #298, opened 2018-03-07
Logzero syslog support

brianlenz:

Changes custom syslog implementation to use logzero 1.5.0's native syslog support.

Updated to logzero 1.5.0 and updated the integration accordingly.

Tested it.

No.

- [ ] Did you add any tests?


metachris:
nice 👍

PR #299, opened 2018-03-08
Update makefile to phony 'test'

ixje:
if a user has a file called `test` in the root directory, then `make test` will always claim its up to date and not run. (it happened to me because I had a wallet called `test`)

add 'test to the phony list

`make test` now works :)

no

- [ ] Did you add any tests?
Do we add such a minor change to the changelog?


metachris:
Can you also add the other makefile commands? I think `lint` is missing. anything else?

metachris:
👍

PR #300, opened 2018-03-08
SS debug improvements (changelog update)

ixje:
this was supposed to be part of #294 

add description to changelog

run `make docs`

no

- [ ] Did you add any tests?


metachris:
👍

PR #308, opened 2018-03-10
Privnet chain check

metachris:

In prompt.py, added a check if the privnet database is of the same privnet container. An old database is a common error case, without an error message and the only indicator a not increasing block count.

Code

Test

No

- [ ] Did you add any tests?

### DIFF
--- a/prompt.py
+++ b/prompt.py
@@ -960,7 +960,7 @@ def check_privatenet():
         if nonce_chain != nonce_container:
             raise PrivnetWrongChainDatabaseError(
                 "Chain database in Chains/privnet is for a different private network than the current container. "
-                "Consider deleting the Chain directory with 'rm -rf Chains/privnet."
+                "Consider deleting the Chain directory with 'rm -rf Chains/privnet*'."
             )
     else:
         with open(neopy_chain_meta_filename, "w") as f:
--- a/prompt.py
+++ b/prompt.py
@@ -8,6 +8,7 @@ import psutil
 import traceback
 import logging
 
+from logzero import logger
 from prompt_toolkit import prompt
 from prompt_toolkit.contrib.completers import WordCompleter
 from prompt_toolkit.history import FileHistory
@@ -44,6 +45,9 @@ from neo.UserPreferences import preferences
 from neocore.KeyPair import KeyPair
 from neocore.UInt256 import UInt256
 
+from neorpc.Client import RPCClient
+from neorpc.Settings import settings as rpc_settings
+
 # Logfile settings & setup
 LOGFILE_FN = os.path.join(DIR_PROJECT_ROOT, 'prompt.log')
 LOGFILE_MAX_BYTES = 5e7  # 50 MB
@@ -52,6 +56,10 @@ settings.set_logfile(LOGFILE_FN, LOGFILE_MAX_BYTES, LOGFILE_BACKUP_COUNT)
 
 # Prompt history filename
 FILENAME_PROMPT_HISTORY = os.path.join(DIR_PROJECT_ROOT, '.prompt.py.history')
+
+
+class PrivnetWrongChainDatabaseError(Exception):
+    pass
 
 
 class PromptInterface(object):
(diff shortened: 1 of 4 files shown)


protocol.privnet.json
```diff
@@ -27,7 +27,7 @@
 
   "ApplicationConfiguration": {
     "DataDirectoryPath": "./Chains/privnet",
-    "NotificationDataPath": "Chains/Priv_Notif",
+    "NotificationDataPath": "Chains/privnet_notif",
```
metachris: I changed this so we can use `rm -rf Chains/privnet*`

prompt.py
```diff
@@ -933,6 +941,32 @@ def run(self):
                 traceback.print_exc()
 
 
+def check_privatenet():
```
metachris: Not sure we want this in prompt.py, but added it here for a first proof of concept.

metachris:
Just a proof of concept. Thoughts?

prompt.py
```diff
@@ -933,6 +941,32 @@ def run(self):
...
+    """ Check if privatenet is running, and if container is same as the chain file """
+    host = "http://127.0.0.1:30333"
```
ixje: Can we read this from the settings/protocol.privnet.json (or whatever option is used to indicate which network to use) instead? I don't think we should assume people only run a privnet on their local machine.

prompt.py
```diff
@@ -933,6 +941,32 @@ def run(self):
...
+    rpc_settings.setup([host])
+    client = RPCClient()
+    version = client.get_version()
+    # print("version", version)
+    if not version:
+        logger.error("Error: private network doesn't seem to be running")
```
ixje: "or RPC server is not enabled."

In continuation of the previous comment. I don't think we should assume everybody to use the CoZ or your docker container. Some might have followed the docs.neo.org steps and don't have the RPC server enabled.

prompt.py
brianlenz: Yeah, I think this all belongs in `SettingsHolder.setup_privnet()`. That way, any utility using the privnet (API servers and other custom utilities) get the same benefit. That should also give you access to the config data you'll need for RPC config, as recommended below.

protocol.privnet.json
brianlenz: Perhaps this should be mentioned explicitly in the CHANGELOG since existing `Priv_Notif` chains will no longer be used?

prompt.py
metachris: done

prompt.py
metachris: done

prompt.py
metachris: Good idea

prompt.py
```diff
@@ -959,7 +958,7 @@ def check_privatenet():
         nonce_chain = open(neopy_chain_meta_filename, "r").read()
         if nonce_chain != nonce_container:
             raise PrivnetWrongChainDatabaseError(
-                "Chain database in Chains/privnet is for a different private network than the current container. "
+                "Chain database in Chains/privnet is for a different private network than the current container (or RPC is not enabled). "
```
brianlenz: If RPC isn't running, the code should short out above, right? So, I don't think this exception will be raised when RPC isn't running...

protocol.privnet.json
metachris: absolutely, added right now

prompt.py
```diff
@@ -933,6 +941,31 @@ def run(self):
...
+        return
+
+    # Now check if nonce is the same as in the chain path
+    nonce_container = str(version["nonce"])
+    neopy_chain_meta_filename = os.path.join(settings.LEVELDB_PATH, ".privnet-nonce")
+    if os.path.isfile(neopy_chain_meta_filename):
+        nonce_chain = open(neopy_chain_meta_filename, "r").read()
+        if nonce_chain != nonce_container:
+            raise PrivnetWrongChainDatabaseError(
+                "Chain database in Chains/privnet is for a different private network than the current container (or RPC is not enabled). "
```
brianlenz: If RPC isn't running, the code should short out above, right? So, I don't think this exception will be raised when RPC isn't running...

ixje: I think so too. My "RPC is not enabled" comment was also at Line 951, so I think it's just a matter of moving it up there :)

ixje: @metachris did you get a chance to try out the concern @brianlenz raised? It would be good to make sure the "RPC not enabled" error message gets printed if RPC isn't running.

metachris: Yeah good points, was a bit tired when adding it here! Fixed with https://github.com/CityOfZion/neo-python/commit/791bd6c017d321e8119c76efe469bfaaf0576038

prompt.py
metachris: Will do this soon. Was kind of hesitant at the beginning to have actual network code executed when setting up the settings, but this would really be the best place for this.

prompt.py
ixje: 🥇

PR #309, opened 2018-03-10
Fix multi-signature contract usage and add NEO TestNet fund retrieval instructions

ixje:
It was unclear how to retrieve NEO TestNet funds using NEO Python as that required a multi-signature contract for which no visible support was present on the prompt. I exposed the `import multisig_addr` command and fixed the `Contract` implementation such that its functional.

write docs & fix code

retrieved my testnet funds and send them to my own wallet address. NEOSCAN confirms it's successful

nah

- [ ] Did you add any tests?

### DIFF
--- a/neo/SmartContract/Contract.py
+++ b/neo/SmartContract/Contract.py
@@ -91,8 +91,8 @@ class Contract(SerializableMixin, VerificationCode):
     @staticmethod
     def CreateMultiSigRedeemScript(m, publicKeys):
 
-        if m < 2 or m > len(publicKeys) or len(publicKeys) > 1024:
-            raise Exception('Invalid keys')
+        if m < 1 or m > len(publicKeys) or len(publicKeys) > 1024:
+            raise Exception('Argument Exception')
 
         sb = ScriptBuilder()
         sb.push(m)
@@ -114,7 +114,7 @@ class Contract(SerializableMixin, VerificationCode):
 
         pk = [ECDSA.decode_secp256r1(p).G for p in publicKeys]
         return Contract(Contract.CreateMultiSigRedeemScript(m, pk),
-                        bytearray(b'\x00\x00\x00'),
+                        bytearray(m),
                         publicKeyHash)
 
     @staticmethod
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,9 @@ All notable changes to this project are documented in this file.
 - Added VM instruction counter to ``ExecutionEngine.py`` error messages to indicate the final instruction that failed. Allows for setting conditional breakpoints to support SC debugging.
 - Renamed ``neo.api.REST.NotificationRestApi`` to ``neo.api.REST.RestApi``
 - Added ``-v/--verbose`` argument to prompt.py, which makes prompt.py show smart contract events by default
+- Fix multi-signature contract import to allow using a single signature
+- Fix fund sending from multi-signature contract
+- Added instructions on retrieving NEO TestNet funds
 
 
 [0.5.3] 2018-03-04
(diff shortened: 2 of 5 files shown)


neo/SmartContract/Contract.py
```diff
@@ -91,8 +91,8 @@ def Create(redeemScript, parameterList, publicKeyHash):
     @staticmethod
     def CreateMultiSigRedeemScript(m, publicKeys):
 
-        if m < 2 or m > len(publicKeys) or len(publicKeys) > 1024:
-            raise Exception('Invalid keys')
+        if m < 1 or m > len(publicKeys) or len(publicKeys) > 1024:
```
ixje: The [C# code](https://github.com/neo-project/neo/blob/879952a8259347db3f04211502efebe1dec202cd/neo/SmartContract/Contract.cs#L75) also allows 1 signature,  now this does so as well.

neo/SmartContract/Contract.py
```diff
@@ -114,7 +114,7 @@ def CreateMultiSigContract(publicKeyHash, m, publicKeys):
 
         pk = [ECDSA.decode_secp256r1(p).G for p in publicKeys]
         return Contract(Contract.CreateMultiSigRedeemScript(m, pk),
-                        bytearray(b'\x00\x00\x00'),
+                        bytearray(m),
```
ixje: The `ParameterList` argument was fixed to 3 signatures, while it should be dynamic based on how many signatures the contract requires ([C# source](https://github.com/neo-project/neo/blob/879952a8259347db3f04211502efebe1dec202cd/neo/SmartContract/Contract.cs#L69)). A ContractParameterType of type `signature` is `0x00`, so we can simply re-use the `bytearray` constructor.

neo/SmartContract/Contract.py
```diff
@@ -91,8 +91,8 @@ def Create(redeemScript, parameterList, publicKeyHash):
     @staticmethod
     def CreateMultiSigRedeemScript(m, publicKeys):
 
-        if m < 2 or m > len(publicKeys) or len(publicKeys) > 1024:
-            raise Exception('Invalid keys')
+        if m < 1 or m > len(publicKeys) or len(publicKeys) > 1024:
+            raise Exception('Argument Exception')
```
metachris: "Argument exception" is perhaps too generic? Would be great to have a more user-friendly message.

ixje: Yes correct, it now is generic. Reason for that is because previously it would give a misleading message saying "invalid keys" while tit could be he minimum amount of signatures required was too low (e.g. if m < 2). I guess I would have to split it up then.

How about?
```python
if m < 1:
    raise Exception("Minimum required signature count is 1, specified {}.".format(m))

if m > len(publicKeys):
    raise Exception("Invalid public key count. Minimum required signatures is bigger than supplied public keys count.")

if len(publicKeys) > 1024:
   raise Exception("Supplied public key count ({}) exceeds maximum of 1024.".format(len(publicKeys))

```

localhuman: That makes sense

neo/SmartContract/Contract.py
```diff
@@ -91,8 +91,14 @@ def Create(redeemScript, parameterList, publicKeyHash):
...
-        if m < 1 or m > len(publicKeys) or len(publicKeys) > 1024:
-            raise Exception('Argument Exception')
+        if m < 1:
+            raise Exception("Minimum required signature count is 1, specified {}.".format(m))
+
+        if m > len(publicKeys):
+            raise Exception("Invalid public key count. Minimum required signatures is bigger than supplied public keys count.")
+
+        if len(publicKeys) > 1024:
+            raise Exception("Supplied public key count ({}) exceeds maximum of 1024.".format(len(publicKeys)))
```
ixje: Note the feedback fix is inhere as well

metachris: 👍

PR #310, opened 2018-03-11
Add VM instruction logging option

ixje:
A continuation of #294 , specifically point 3

> add instruction logging
rationale; having a historical log of all calls executed gave me nearly all the information needed to exactly pin point which code line of my SC the VM was executing (by looking at function calls, DROP/RET statements etc).

Add some file writing statements to ApplicationEngine/ExecutionEngine and StateReader.

write tests + `make test`

no

- [X] Did you add any tests?


ixje:
Note: I specifically choose to implement the file writing in the `Execute()` function over the `StepInto()` function because the latter would mean you'd get hundreds of file open/closes per script execution (literally for every instruction).

neo/SmartContract/StateReader.py
```diff
@@ -284,6 +284,9 @@ def Runtime_Notify(self, engine):
             payload
         )
 
+        message = payload[0] if len(payload) > 0 else payload
+        engine.write_log(str(message))
+
```
ixje: FYI; This and the one below is added to make sure that we don't just see a 
```
14 SYSCALL
Neo.Runtime.Notify
```
but also see the string/argument it's printing in the log
```
14 SYSCALL
Neo.Runtime.Notify
b'Start main'
```

ixje:
The concern raised was speed, which is indeed important to keep in mind. Of course there's some impact to code additions, therefore I did a basic optimisation on the calls that get called _every_ instruction. This yielded a 9% improvement to offset the <2% loss by the new functionality.

localhuman:
Looks good! Tests seem to be wanting `timeit` from the `neo.Utils` module

localhuman:
This will be very helpful for many people until we get some actual debugging in place!

PR #311, opened 2018-03-12
Settings.check_dependencies to verify dependencies

metachris:

When people do a `git pull`, they often get cryptic error messages. This PR adds a dependency check to verify that all required dependencies are those actually installed.

Test

No

- [ ] Did you add any tests?

### DIFF
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -77,10 +77,10 @@ def check_depdendencies():
 
     # Now check if each package specified in requirements.txt is actually installed
     deps_filename = os.path.join(DIR_PROJECT_ROOT, "requirements.txt")
-    deps_list = open(deps_filename, "r").read().split()
-    for dep in deps_list:
-        if not dep.lower() in installed_packages_list:
-            raise DependencyError("Required dependency %s is not installed. Please run 'pip install -e .'." % dep)
+    with open(deps_filename, "r") as f:
+        for dep in f.read().split():
+            if not dep.lower() in installed_packages_list:
+                raise DependencyError("Required dependency %s is not installed. Please run 'pip install -e .'." % dep)
 
 
 class SettingsHolder:
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -71,10 +71,11 @@ def check_privatenet():
 
 
 def check_depdendencies():
+    # Get installed packages
     installed_packages = pip.get_installed_distributions()
     installed_packages_list = sorted(["%s==%s" % (i.key, i.version) for i in installed_packages])
 
-    # Now check if each required package is actually available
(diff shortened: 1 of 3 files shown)


neo/Settings.py
```diff
@@ -32,6 +37,52 @@
...
 
+class PrivnetConnectionError(Exception):
+    pass
+
+
+class DependencyError(Exception):
+    pass
+
+
+def check_privatenet():
```
metachris: also, i moved check_privatenet from prompt.py here to settings.

neo/Settings.py
```diff
@@ -191,3 +243,8 @@ def set_loglevel(self, level):
 
 # By default, set loglevel to INFO. DEBUG just print a lot of internal debug statements
 settings.set_loglevel(logging.INFO)
+
+# Check if currently installed dependencies match the requirements
+# Can be bypassed with `SKIP_DEPS_CHECK=1 python prompt.py`
+if not os.getenv("SKIP_DEPS_CHECK"):
```
ixje: good thinking!

neo/Settings.py
```diff
@@ -32,6 +37,52 @@
...
+    # Get installed packages
+    installed_packages = pip.get_installed_distributions()
+    installed_packages_list = sorted(["%s==%s" % (i.key, i.version) for i in installed_packages])
+
+    # Now check if each package specified in requirements.txt is actually installed
+    deps_filename = os.path.join(DIR_PROJECT_ROOT, "requirements.txt")
+    deps_list = open(deps_filename, "r").read().split()
+    for dep in deps_list:
+        if not dep.lower() in installed_packages_list:
+            raise DependencyError("Required dependency %s is not installed. Please run 'pip install -e .'." % dep)
```
ixje: Don't forget to close the file to free up the system resources (even if it's just a small file). Ideally use `with open` because that will make sure to close the file even if for some reason an exception is raised.

metachris: thanks! downside is it adds 2 lines :P

metachris: updated

neo/Settings.py
brianlenz: This is nitpicky, but does this more appropriately belong in the `SettingsHolder` class? It seems like that may be a more appropriate OO-design, and then you don't need to reference `settings` and instead can just get those directly from `self`. Without this, you could technically have multiple `SettingsHolder` instances with different `RPC_LIST` and `LEVELDB_PATH` values and this would not work as expected.

metachris:
Thanks, great suggestion! Updated

neo/Settings.py
metachris: done

localhuman:
nice implementation!

PR #312, opened 2018-03-12
Optionally specifiy custom host with `-p`

metachris:

This PR allows `prompt.py` users to specify custom hosts with the `-p` privatenet argument. Useful for instance for events such as hackathons, where people want to connect to the privnet running on a shared server somewhere.

Help output:

    $ python prompt.py -h
    usage: prompt.py [-h] [-m | -p [host] | --coznet | -c CONFIG]
                    [-t {dark,light}] [-v] [--version]

    optional arguments:
    -h, --help            show this help message and exit
    -m, --mainnet         Use MainNet instead of the default TestNet
    -p [host], --privnet [host]
                            Use a private net instead of the default TestNet,
                            optionally using a custom host (default: 127.0.0.1)
    --coznet              Use the CoZ network instead of the default TestNet
    -c CONFIG, --config CONFIG
                            Use a specific config file
    -t {dark,light}, --set-default-theme {dark,light}
                            Set the default theme to be loaded from the config
                            file. Default: 'dark'
    -v, --verbose         Show smart-contract events by default
    --version             show program's version number and exit

Normal privatenet behavior (as before):

    $ python prompt.py -p

With a custom host:

    $ python prompt.py -p example.com
    Using custom privatenet host: example.com
    - Seeds: example.com:20333, example.com:20334, example.com:20335, example.com:20336
    - RPC: http://example.com:30333

With an invalid host:

    $ python prompt.py -p http://example.com
    File "prompt.py", line 1048, in <module>
        main()
    File "prompt.py", line 1004, in main
        settings.setup_privnet(args.privnet)
    File "/Users/chris/Projects/chris/neo/neo-python/neo/Settings.py", line 161, in setup_privnet
        raise Exception("No protocol prefix or port allowed in host, use just the IP or domain.")
    Exception: No protocol prefix or port allowed in host, use just the IP or domain.
    
Added nargs to argparse line in prompt, and added a `host` argument in `Settings.setup_privnet`

Manual tests

No

- [ ] Did you add any tests?

### DIFF
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -160,9 +160,9 @@ class SettingsHolder:
             if ":" in host:
                 raise Exception("No protocol prefix or port allowed in host, use just the IP or domain.")
             print("Using custom privatenet host:", host)
-            self.SEED_LIST = [entry.replace("127.0.0.1", host) for entry in self.SEED_LIST]
-            self.RPC_LIST = [entry.replace("127.0.0.1", host) for entry in self.RPC_LIST]
-            print("- Seeds:", ", ".join(self.SEED_LIST))
+            self.SEED_LIST = ["%s:20333" % host]
+            self.RPC_LIST = ["http://%s:30333" % host]
+            print("- P2P:", ", ".join(self.SEED_LIST))
             print("- RPC:", ", ".join(self.RPC_LIST))
 
     def setup_coznet(self):
--- a/neo/Settings.py
+++ b/neo/Settings.py
@@ -147,9 +147,23 @@ class SettingsHolder:
         """ Load settings from the testnet JSON config file """
         self.setup(FILENAME_SETTINGS_TESTNET)
 
-    def setup_privnet(self):
-        """ Load settings from the privnet JSON config file """
+    def setup_privnet(self, host=None):
+        """
+        Load settings from the privnet JSON config file
+
+        Args:
+            host (string): can be either an IP which will be used for seeds with standard ports and RPC
+                           or in the format of `IP:port`, which will be used as the only seed
+        """
         self.setup(FILENAME_SETTINGS_PRIVNET)
+        if isinstance(host, str):
+            if ":" in host:
+                raise Exception("No protocol prefix or port allowed in host, use just the IP or domain.")
+            print("Using custom privatenet host:", host)
+            self.SEED_LIST = [entry.replace("127.0.0.1", host) for entry in self.SEED_LIST]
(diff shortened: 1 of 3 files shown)


neo/Settings.py
```diff
@@ -147,9 +147,23 @@ def setup_testnet(self):
...
 
-    def setup_privnet(self):
-        """ Load settings from the privnet JSON config file """
+    def setup_privnet(self, host=None):
+        """
+        Load settings from the privnet JSON config file
+
+        Args:
+            host (string): can be either an IP which will be used for seeds with standard ports and RPC
+                           or in the format of `IP:port`, which will be used as the only seed
```
brianlenz: How would you use this `IP:port` format? It looks like colons aren't supported below, but I could be missing something...

metachris: Oh i left that comment in. The problem I faced with port was that you'd need 2 ports: at least a p2p and a rpc server port. Didn't have an idea about how to do this properly, so i thought let's start simple and improve later. If you have any ideas i'd be happy to hear them! ;)

ixje: While I don't think you'll need more than 1 host on a privatenet (and thus this solution would suffice), one thing I'd like to mention just for consideration is the option to specify a `list` of 'host:port' pairs and replace the whole SEED_LIST or RPC_LIST with that. 

I am struggling a bit with the thought that this PR sort of solves the same problems as the `-c` option does, no?

brianlenz: That is a good point. Perhaps for the hackathon "shared" private net use case, they should just be distributing a json file for everyone to use with `-c`?

metachris: It just makes it easier. For instance a hackathon, you spin up a VM somewhere with privnet + neoscan + faucet. Then you just tell people to run `./prompt.py -p hackathon.cityofzion.io` and that's it. No need to copy around protocol-config json files beforehand.

True, a single host should be enough.

The question is how we'd define the RPC port, but at the same time keep things as simple as possible.

metachris: At hackathons, every thing you don't need to do for the system setup is worth it's weight in gold. Just make it as easy as possible for participants, and I think this is it.

brianlenz: Cool, you definitely know better than me on this one 😉 I'm good with keeping it simple, but I think we probably only need to support the single host scenario. For anything more complex, I would think you should just be using a config file.

metachris: This is a great middle ground.

metachris:
Thanks so much for the brainstorming guys!

localhuman:
Looks great!

PR #315, opened 2018-03-14
add ability to specify --datadir for where to store leveldb directories

localhuman:
- add ability to specify an absolute path/ alternative path for data storage to `prompt` and `api-server` by passing an argument with `--datadir=/path/to/wherever`
 
- added more args to parse, updated `neo/Settings.py`

- tests

- should be backwards compatible 

- [x] Did you add any tests?


localhuman:
Going to push some changes for the bootstrap file too, just testing it out now

prompt.py
```diff
@@ -970,6 +970,9 @@ def main():
     parser.add_argument("-v", "--verbose", action="store_true", default=False,
                         help="Show smart-contract events by default")
 
+    # Where to store stuff
+    parser.add_argument("--datadir", action="store", help="Absolute path to prepend to paths for database directories")
```
metachris: should we call it `--db-path` to make it more explicit? `--datadir` works too though.

Would change help to "Absolute path to use for database directories"

metachris:
Like it a lot 👍

Issue #316, opened 2018-03-14
Change README.md to README.rst


metachris:
We should convert README.md to .rst format, because it's reused here and it's fucked up :P https://pypi.python.org/pypi/neo-python

Since the docs are in .rst format, we are all familiar with rst syntax anyway.

localhuman:
I feel like I tried this before for something else.  Perhaps have it autoconverted? https://stackoverflow.com/questions/10718767/have-the-same-readme-both-in-markdown-and-restructuredtext

localhuman:
fixed in https://github.com/CityOfZion/neo-python/pull/317

PR #317, opened 2018-03-14
Converted README.md to README.rst

metachris:
see #316


localhuman:
removing the .md messes up the github page 🤕 
https://github.com/CityOfZion/neo-python/tree/readme.rst

metachris:
It's not because of removing the .md, it's because the .rst has a syntax error and that's why it's not rendering properly. Sorry, taking a look atm.

metachris:
Works now!

PR #320, opened 2018-03-14
Add ExecutionEngine error messages to improve SC debugging feedback

ixje:
A continuation of #294 , specifically point 4 (the final one)
>add additional error messages to ExecutionEngine.
rationale; there's a lot of error feedback we can simply give that we're not doing leading to guess work or a lot of stepping through the VM internals.

As discussed here
> How about the following for 4;
> all interesting spots for adding additional error information seem to end with:
>                     
> self._VMState |= VMState.FAULT
> return
> 
> what if we replace that with an error reporting function e.g.
> def VM_FAULT_and_report(id, *args):
>     self._VMState |= VMState.FAULT
> 
>     # additional error handling/message providing based on id
>     ...
> 
> That way we can keep further handling all in 1 spot and keep ExecutionEngine as bare as possible

> March 8, 2018
> localhuman - Last Thursday at 1:55 PM
> I think its a good approach

PoC, no tests yet.

no

- [X] Did you add any tests?


localhuman:
I think it looks good but I worry about maintaining the integers for the error reports and keeping them matched up between the execution and the reporting. 

 So instead of doing `self.VM_FAULT_and_report(2)` it would be `self.VM_FAULT_and_report( VMFAULT.INVALID_JUMP)`

Does that make sense?

metachris:
Agree, makes a lot of sense to use constants for this!

ixje:
@metachris Note that I didn't use the testing approach you gave me, because I couldn't get `capsys` from the `pytest` package to work that you used for capturing the `stdoutput/err`. I'm guessing it requires a different test runner than what we use in `neo-python`. Either way, my current approach doesn't require a new package so I guess it should suffice.

neo/VM/ExecutionEngine.py
```diff
@@ -669,8 +650,7 @@ def ExecuteOp(self, opcode, context):
                 size = estack.Pop().GetBigInteger()
 
                 if size < 0 or size > estack.Count:
-                    self._VMState |= VMState.FAULT
-                    return
+                    return self.VM_FAULT_and_report(18)
```
localhuman: should `18` be a `VMFault`?

neo/VM/ExecutionEngine.py
```diff
@@ -246,8 +238,7 @@ def ExecuteOp(self, opcode, context):
                 n = estack.Pop().GetBigInteger()
 
                 if n < 0:
-                    self._VMState |= VMState.FAULT
-                    return
+                    return self.VM_FAULT_and_report(VMFault.UNKNOWN3)
```
localhuman: Will `UNKNOWN3-6` be replaced with something?

neo/VM/ExecutionEngine.py
ixje: oops missed

neo/VM/ExecutionEngine.py
ixje: For now not, unless you have a clear idea what a good fault statement would be for those. I decided to keep them `generic unknown` until with some more experience I figure out a good name for it.

localhuman:
Nice work!

PR #314, opened 2018-03-12
For development and workshop convenience, added neo-privnet.wallet

metachris:
This PR adds the standard `neo-privnet.wallet` to the project root.

I know it's perhaps a controversial PR, to add a semi-necessary file to the project root, but I think the developer simplicity, especially for newcomers and hackathons, trumps it. And it doesn't really hurt.

**How did you solve this problem?**

* Added `neo-privnet.wallet`

**How did you make sure your solution works?**

Test

**Are there any special changes in the code that we should be aware of?**

No

**Please check the following, if applicable:**

- [ ] Did you add any tests?
- [X] Did you run `make lint`?
- [X] Did you run `make test`?
- [X] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

PR #321, opened 2018-03-15
Major changeset for pypi compatibility.  

localhuman:
- move data files like `protocol.testnet.json` to `neo/data`, 
- move executables to `neo/bin` 
- add console_scripts for executables with a prefix of `np-`
- adds following in settings:
  - `DIR_PROJECT_ROOT` is `os.getcwd()`
  - `DIR_PROJECT_INSTALL` is the same, unless it is installed via pypi, then it is the `venv/lib/site-packages`
  - `ROOT_INSTALL_PATH` is one of those
  - `IS_PACKAGE_INSTALL` is whether it is installed locally/editably  or via pip

scripts can now be run normally, but `python prompt.py` would now need to be `python neo/bin/prompt.py`.  But prompt can also be run with the command `np-prompt` whether installed from pip or if you run `pip install -e .`  same goes for things like `python bootstrap.py` becoming simply `np-bootstrap` and `python api_server.py` becoming `np-api-server`

- adds ability to easily use when installed from pip

- so much fun

- ran tests

- see above

- [ ] Did you add any tests?

I suspect / hope everyone has some feedback on this, as it changes the structure of the project quite a bit!


neo/Settings.py
```diff
@@ -19,22 +19,35 @@
...
 from neorpc.Client import RPCClient
 from neorpc.Settings import settings as rpc_settings
-
+import sys
 
 # Create am absolute references to the project root folder. Used for
 # specifying the various filenames.
 dir_current = os.path.dirname(os.path.abspath(__file__))
-DIR_PROJECT_ROOT = os.path.abspath(os.path.join(dir_current, ".."))
+DIR_PROJECT_ROOT = os.getcwd()
```
metachris: I think this solution is not so great. Imagine I run `np_prompt` from my home directory, and then from any other directory. It will create 2x the Chain directory and start syncing from scratch. I think we should use something like `~/.neopython/Chains` as default directory. [XDG Base Directory](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support) might be helpful for this too.

metachris:
First of all, I really like where this is going! 👍 

My major concern: Using `os.getcwd` to set the path for the chain database. If a user runs `np_prompt` from different directories, it will create different chain databases and start syncing from scratch. I think the proper solution here would be to store the chain by default in `~/.neopython/chains`.

A smaller thing: I'd much rather have the commands be `np-api-server` or `np-prompt` than with underscores. Dash is pretty much the standard for separated program names for the cli, and I think it's much more beautiful too.

Minor thing: please don't move the privnet wallet to `neo-privnet.sample.wallet`. It is supposed to be `neo-privnet.wallet` because that is compatible with all the guides and instructions that are out there.

metachris:
Re @ixje: you can use a custom `--datadir` argument if you want to run multiple neo-python's in parallel, which normal users typically don't want to do. The `~/.neopython` solution would be much better imo, because users would run the system-wide `np-prompt` command from anywhere, and it would still use the same database, as it should imo.

ixje:
didn't know about the `--data-dir` option 👍

localhuman:
is `~/.neopython/chains` windows compatible?

metachris:
No it is not, I think. This should work as base directory for all users data: `os.environ['ALLUSERSPROFILE']`

brianlenz:
@metachris, I think it's preferable to keep neo-privnet.wallet out of version control. I explained in my comment in your PR: https://github.com/CityOfZion/neo-python/pull/314#commitcomment-28093360

I _much_ prefer the `.sample` filename approach here to avoid confusion and issues with having a binary data file that is regularly used and modified under version control. If someone wants to use that wallet file, they should make a copy of it to the standard `neo-privnet.wallet`, which will be ignored in `.gitignore`. This is a very common practice for version control for "default" files.

I agree that it might be better to put the `bin` and `data` dirs in the project root as opposed to in the `neo` dir.

localhuman:
I'd prefer to have `bin` and `data` outside the root `neo` directory, but it doesn't seem very possible or well structured to do so when packaging.  From what I've run into, it seems any packaged data needs to be within a package.  I'll give it another shot though.

localhuman:
So in regards to /bin and /data being in the neo package, we can move them outside to the root of the directory, but then when packaging/installed via pip/setup.py they get installed at `/bin` and `/data` in the `venv/lib/site-packages/python3.6` directory

I feel like this is bad practice, and could lead to these conflicting with other packages.
The other option is to move `bin` and `data` to something like `/neo_bin` and `/neo_data` so when they get put into site packages there's no chance of conflict

But then i'm not sure if that is as clean as just keeping them in the `neo` package. 

I think I'll keep it this way for now, until a better solution can be determined

metachris:
Sounds good. Only addition I'd love to see before merging is using `~/.neo-python/chains` as default db path instead of the current working directory. I can take a quick look tomorrow morning.

neo/Settings.py
metachris: Since `DIR_PROJECT_ROOT` changes it's meaning from project root to current working directory, perhaps we should remove it completely? What's the use case for leaving it in?

neo/bin/prompt.py
```diff
@@ -40,19 +44,19 @@
...
 # Logfile settings & setup
-LOGFILE_FN = os.path.join(DIR_PROJECT_ROOT, 'prompt.log')
+LOGFILE_FN = os.path.join(os.getcwd(), 'prompt.log')
 LOGFILE_MAX_BYTES = 5e7  # 50 MB
 LOGFILE_BACKUP_COUNT = 3  # 3 logfiles history
 settings.set_logfile(LOGFILE_FN, LOGFILE_MAX_BYTES, LOGFILE_BACKUP_COUNT)
 
 # Prompt history filename
-FILENAME_PROMPT_HISTORY = os.path.join(DIR_PROJECT_ROOT, '.prompt.py.history')
+FILENAME_PROMPT_HISTORY = os.path.join(os.getcwd(), '.prompt.py.history')
```
metachris: I think the history should also be in `~/.neopython/.prompt.py.history`, else it would create a new history file wherever it is executed

neo/bin/prompt.py
```diff
@@ -40,19 +44,19 @@
...
 from neo.Prompt.InputParser import InputParser
-from neo.Settings import settings, DIR_PROJECT_ROOT, PrivnetConnectionError
+from neo.Settings import settings, PrivnetConnectionError
 from neo.UserPreferences import preferences
 from neocore.KeyPair import KeyPair
 from neocore.UInt256 import UInt256
 
 # Logfile settings & setup
-LOGFILE_FN = os.path.join(DIR_PROJECT_ROOT, 'prompt.log')
+LOGFILE_FN = os.path.join(os.getcwd(), 'prompt.log')
```
metachris: Should this perhaps too go by default into `~/.neopython/prompt.log`?

neo/Settings.py
```diff
@@ -19,22 +19,32 @@
...
+ROOT_INSTALL_PATH = None
+
+# This detects if we are running from
+# an 'editable' version ( like ``python neo/bin/prompt.py`` )
+# or from a packaged install version from pip
+if 'site-packages/neo' in dir_current:
+    ROOT_INSTALL_PATH = os.path.abspath(os.path.join(dir_current, ".."))
+    IS_PACKAGE_INSTALL = True
+else:
+    ROOT_INSTALL_PATH = os.getcwd()
```
metachris: i don't think that works in all circumstances. Let's say I'm in a subdirectory and execute `python ../prompt.py` then the references based on ROOT_INSTALL_PATH will fail. In this case here I think we should use `ROOT_INSTALL_PATH = os.path.abspath(os.path.join(dir_current, ".."))`

metachris:
Looks good, great to get rid of DIR_PROJECT_ROOT. I left a few more comments inline. 

Honestly, I would try to get rid of `os.getcwd()` because with the new packaging and system-wide cli commands, this will lead to weird behaviour. I think we should replace this everywhere with either `~/.neopython/` or the actual code root, from where we can reference the other things. I think this would be an important improvement.

examples/rest-api-server.py
```diff
@@ -18,11 +18,11 @@
 from neo.Implementations.Notifications.LevelDB.NotificationDB import NotificationDB
```
metachris: Should we perhaps remove this example completely in favor of `api-server.py`, or does this still provide a benefit?

neo/Settings.py
metachris: Thinking about it, I don't think we need this at all? We could just set `ROOT_INSTALL_PATH = os.path.abspath(os.path.join(dir_current, ".."))` no matter if installed as package or not?

neo/Settings.py
```diff
@@ -309,5 +320,5 @@ def check_privatenet(self):
 
 # Check if currently installed dependencies match the requirements
 # Can be bypassed with `SKIP_DEPS_CHECK=1 python prompt.py`
-if not os.getenv("SKIP_DEPS_CHECK"):
+if not os.getenv("SKIP_DEPS_CHECK") and not IS_PACKAGE_INSTALL:
```
metachris: Why would you skip the dependency check when installed as package?

neo/bin/prompt.py
```diff
@@ -17,6 +17,10 @@
 from prompt_toolkit.token import Token
 from twisted.internet import reactor, task
 
+
+if os.path.exists(os.path.join(os.getcwd(), 'neo')):
```
metachris: why do we need this -- neo should be already importable, either way? I remember this being an issue way back, before using `pip install -e .`

neo/Settings.py
localhuman: Requirements.txt isn't included in the package, so we can't read it.  Also, If we're running from a package wouldn't it be safe to assume that the bundled code would have the correct dependencies?

localhuman:
Thanks for the great suggestions!

neo/Settings.py
metachris: yeah good point, that actually should work just fine. but then again, wouldn't the check work anyway? ^^ but yeah, this is mostly intended for people pulling from the dev branch, so disabling it for using as package is fine for me. there might even be weird conflicts where the user updates a specific package and neo-python would scream because it's not the exact version it expects, so perhaps even better like this. 👍

localhuman:
Warning for moving chain directory:
```
[I 180316 10:46:02 Settings:282] [MIGRATION Warning] You are now using the blockchain data at /Users/thomassaunders/.neopython/Chains, but it appears you have existing data at /Users/thomassaunders/Workshop/neo-python/Chains
[I 180316 10:46:02 Settings:283] [MIGRATION Warning] If you would like to use your existing data, please move any data at /Users/thomassaunders/Workshop/neo-python/Chains to /Users/thomassaunders/.neopython/Chains 
[I 180316 10:46:02 Settings:284] [Migration Warning] Or you can continue using your existing data by starting your script with the `--datadir=.` flag
```

brianlenz:
Great work, guys!

PR #326, opened 2018-03-16
pypi PR minor fixes in Settings and prompt: use proper paths rather than `os.getcwd()`

metachris:
Replaced `os.getcwd()` with proper paths, because using the current working directory can lead to various unwanted edge-cases (like creating a prompt-history file in whatever directory the user currently is in).

Also cleaned up settings paths a little bit.

This is a PR to the other open PR #321.

### DIFF
--- a/neo/bin/prompt.py
+++ b/neo/bin/prompt.py
@@ -17,10 +17,6 @@ from prompt_toolkit.styles import style_from_dict
 from prompt_toolkit.token import Token
 from twisted.internet import reactor, task
 
-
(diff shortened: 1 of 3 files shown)


neo/bin/prompt.py
```diff
@@ -17,10 +17,6 @@
 from prompt_toolkit.token import Token
 from twisted.internet import reactor, task
 
-
-if os.path.exists(os.path.join(os.getcwd(), 'neo')):
```
metachris: this should be solved by the change in travis.yml to `pip install -e .`

PR #327, opened 2018-03-16
fixed docker deployment on new tag (hopefully)

metachris:
Just removed the global encrypted DOCKER vars from `.travis.yml` and added them in the Travis CI settings. Docker deployments should now work... hopefully :P


metachris:
doesn't work yet :(

Issue #322, opened 2018-03-15
Implement a way to choose from which address to invoke a smart contract if there are multiple addresses in a wallet.


metachris:
Implement a way to choose from which address to invoke a smart contract if there are multiple addresses in a wallet.

brianlenz:
I've got a POC in place for this. I just need to put some finishing touches and create the PR.

localhuman:
That's great!

brianlenz:
This can be closed now...

localhuman:
💯

PR #329, opened 2018-03-16
Invoke from address selection

brianlenz:

#322

Added support for `--from-addr=` parameters for the various invoke methods in both `prompt.py` and other relevant commands.

Tested it. Updated a test to use it.

No.

- [X] Did you add any tests?

### DIFF
--- a/neo/bin/prompt.py
+++ b/neo/bin/prompt.py
@@ -104,7 +104,7 @@ class PromptInterface(object):
                 'wallet tkn_approve {token symbol} {address_from} {address to} {amount}',
                 'wallet tkn_allowance {token symbol} {address_from} {address to}',
                 'wallet tkn_mint {token symbol} {mint_to_addr} (--attach-neo={amount}, --attach-gas={amount})',
-                'wallet tkn_register {addr} ({addr}...)',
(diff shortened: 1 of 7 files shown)


neo/bin/prompt.py
```diff
@@ -104,7 +104,7 @@ class PromptInterface(object):
                 'wallet tkn_approve {token symbol} {address_from} {address to} {amount}',
                 'wallet tkn_allowance {token symbol} {address_from} {address to}',
                 'wallet tkn_mint {token symbol} {mint_to_addr} (--attach-neo={amount}, --attach-gas={amount})',
-                'wallet tkn_register {addr} ({addr}...)',
+                'wallet tkn_register {addr} ({addr}...) (--from-addr={addr})',
```
brianlenz: I do wonder if we even want to support `tkn_register` as a native command at all. It's not part of NEP-5, and there are conversations around changing the de facto standard (e.g. camel case instead of underscores). It's easy enough to do with `testinvoke`, so we may just want to get rid of support for it entirely?

localhuman: I think its good to keep around.  We're working on an NEP for sale related ICO methods, so once that is complete we can change the syntax to camelCase or whatever it ends up being.

PR #331, opened 2018-03-17
Fix to redact WIF keys, nep2 keys and contract metadata from the command history file (fixes #301)

### LINKED ISSUE #301: Redact certain commands in prompt history
We should redact certain commands or parts in the prompt history

* `import wif <wif>`
* The meta info when importing a smart contract (name, description)

Others?

BSathvik:
Fixes #301 

By changing the content of ``.prompt.py.history`` file after a particular command among `import wif {wif}`, `export wif {wif}`, `import nep2 {nep2}`, ``export nep2 {nep2}`` are executed. The WIF keys and the nep2 keys are replaced with `<wif>` and `<nep2>` respectively. Metadata (name, description) generated by the command `import contract {path/to/file.avm} {params} {returntype} {needs_storage} {needs_dynamic_invoke}` is also purged from the `.prompt.py.history` file.
 
I tried running the above specified commands and they are successfully redacted.

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.6.2-dev] in progress
 -----------------------
-- ...
+- Fixed ``prompt.py`` to redact WIF keys, nep2 keys and contract metadata from ``.prompt.py.history``
 
 
 [0.6.1] 2018-03-16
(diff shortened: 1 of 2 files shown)


metachris:
Thanks for the contribution 👍 

I'm not sure that opening the history file, searching for a command, and writing back all lines except the match is the best approach to this. I think better would be to not write it in the history file in first place.

This could be achieved with a custom [History](http://python-prompt-toolkit.readthedocs.io/en/master/pages/reference.html#prompt_toolkit.history.History) object that doesn't store specific things. What do you think? See also http://python-prompt-toolkit.readthedocs.io/en/master/pages/building_prompts.html#history

PR #332, opened 2018-03-17
Finally, fixed the Docker auto deploys on new releases/tags

metachris:
That was painful :P

I think the main reason for the problems were special characters in the docker password.

Edit: Now I've added the DOCKER_USER and DOCKER_PASS in the Travis-CI settings, with quotes ^^


localhuman:
awesome!

PR #334, opened 2018-03-17
[Hackathon]: Banlist to protect against DDOS and sybil attacks

nettijoe96:
Issue 325

Theory: potential bad peers are:
1. those relaying false information
2. unstable connection or high latency connection
3. peers that only request but don't send out blocks (leechers; usually because of misconfigurations of port forwarding)

Solution: I added a ban list in the codebase. This solves a HUGE vulnerability because DDOS and sybil attacks are currently not protected against because any false blocks and headers are just discarded without any repercussion for the sender. I created a few functions that would create ratings for peers.

The commit needs some tweaking and testing (I couldn't compile neo correctly yet because make install kept erroring). Specifically, the FALSE_HEADER and FALSE_BLOCK errors need to be curtailed before merging because not all block errors come from malicious nodes.

Ideas for the future:

1. Peers have a second rating called "network trust." Network trust is how much you trust a peers peers. This is based on testing by connecting to a peers peers and seeing the result. If they misbehave, they get banned and the peer that shared those peers network trust goes down. This is a good way to mitigate bad peer of peers and works great alongside banlists.
2. When a node is testing new peer A that they got from seeding or from one of their peers, that node can get an indicator of trust by asking their other peers if they have seen A misbehaving. This is only one indicator of trust.

### DIFF
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -37,7 +37,7 @@ class NeoNode(Protocol):
         self.myblockrequests = set()
         self.bytes_in = 0
         self.bytes_out = 0
-
+        self.rating = None
         self.host = None
         self.port = None
 
@@ -86,6 +86,28 @@ class NeoNode(Protocol):
         self.host = self.endpoint.host
         self.port = int(self.endpoint.port)
         self.leader.AddConnectedPeer(self)
+        # write new entry into Banlist (and make sure that node was not already banned)
+        filename = "misbehaving.txt"
+        f = open(filename, 'r+')
+        ip = None
+        for line in f:
+            attributes = line.split(",")
+            if(attributes[0] == "\n") continue
+            if(attributes[0] == self.host):
+                self.rating = int(attributes[1])
+                if self.rating > 90:
+                    Disconnect()
+                f.close()
+                break
+
+        f.close()
+
+        f = open(filename, 'a')
+        f.write(str(self.host) + "," + "0" + "\n")
+
+        f.close()
+
+
         self.Log("Connection from %s" % self.endpoint)
 
     def connectionLost(self, reason=None):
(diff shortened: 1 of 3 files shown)


ixje:
Building seems to fail due to a syntax error (missing colon):
```
  File "/home/travis/build/CityOfZion/neo-python/neo/Network/NodeLeader.py", line 168
    elif error == "FALSE_BLOCK"
```
 Also please run the `make` commands suggested in the PR template, because for example now there are code style errors. cheers

nettijoe96:
On my side I have unresolved reference to address payload but that was originally in source code (just commented out)

neo/Network/NeoNode.py
```diff
@@ -86,6 +86,28 @@ def connectionMade(self):
         self.host = self.endpoint.host
         self.port = int(self.endpoint.port)
         self.leader.AddConnectedPeer(self)
+        # write new entry into Banlist (and make sure that node was not already banned)
+        filename = "misbehaving.txt"
+        f = open(filename, 'r+')
+        ip = None
+        for line in f:
+            attributes = line.split(",")
```
metachris: Does this work without calling f.read()?

neo/Network/NeoNode.py
```diff
@@ -114,6 +136,50 @@ def dataReceived(self, data):
         self.buffer_in = self.buffer_in + data
         self.CheckDataReceived()
 
+    def changeRating(self, newRating):
+        """changes misbehaving.txt and object rating feild and disconnectes if too high.
+        Misbehaving file format:     IP,rating 0-100              (from least infractions to most and/or worst infractions)
+        """
+
+        filename = "misbehaving.txt"
```
metachris: Filenames should be constants

neo/Network/NeoNode.py
```diff
@@ -337,7 +402,17 @@ def HandleBlockHeadersReceived(self, inventory):
         inventory = IOHelper.AsSerializableWithType(inventory, 'neo.Network.Payloads.HeadersPayload.HeadersPayload')
 
         if inventory is not None:
-            BC.Default().AddHeaders(inventory.Headers)
+            error = BC.Default().AddHeaders(inventory.Headers)
+
+            if error == "FALSE_HEADER":
```
metachris: Use constants instead of inline strings

neo/Network/NeoNode.py
nettijoe96: I never use f.read unless it is implicit in the for line in f loop. Should I avoid it?

localhuman:
Thanks for the PR! If you're having problems getting it to install, reach out on discord or create an issue on the repo, and we'll try our best to help you get it working :)

nettijoe96:
Hi localhuman! Thanks, I'll check out the discord!

localhuman:
Going to close this one for now. We can re-open if further work occurs!

PR #336, opened 2018-03-17
Fix to redact WIF keys, nep2 keys and contract metadata from the command history file (fixes #301)

### LINKED ISSUE #301: Redact certain commands in prompt history

BSathvik:
Fixes #301 

By extending the `FileHistory` class to make a custom `History` object that redacts commands before appending them to ``.prompt.py.history`` file when one of the following commands `import wif {wif}`, `export wif {wif}`, `import nep2 {nep2}`, ``export nep2 {nep2}`` are executed. The WIF keys and the nep2 keys are replaced with `<wif>` and `<nep2>` respectively. Metadata (name, description) generated by the command `import contract {path/to/file.avm} {params} {returntype} {needs_storage} {needs_dynamic_invoke}` is also not added to the `.prompt.py.history` file.

I tried running the above specified commands and they are successfully redacted.

- [ ] Did you add any tests?

### DIFF
--- a/neo/bin/prompt.py
+++ b/neo/bin/prompt.py
@@ -55,6 +55,47 @@ settings.set_logfile(LOGFILE_FN, LOGFILE_MAX_BYTES, LOGFILE_BACKUP_COUNT)
 FILENAME_PROMPT_HISTORY = os.path.join(PATH_USER_DATA, '.prompt.py.history')
 
 
+class NeoHistory(FileHistory):
+    contract_meta_count = 0
+
+    def append(self, string):
+        string = self.redact_command(string)
+        if len(string) == 0:
+            return
+        self.strings.append(string)
+
+        # Save to file.
+        with open(self.filename, 'ab') as f:
+            def write(t):
+                f.write(t.encode('utf-8'))
+
+            write('\n# %s\n' % datetime.datetime.now())
+            for line in string.split('\n'):
+                write('+%s\n' % line)
+
+    def redact_command(self, string):
+        if len(string) == 0:
+            return string
+        command = [comm for comm in ['import wif', 'export wif', 'import nep2', 'export nep2'] if comm in string]
+        if len(command) > 0:
+            command = command[0]
+            if command in string and len(command + " ") < len(string):
+                return command + " <" + command.split(" ")[1] + ">"
+            else:
+                return string
+        elif "import contract" in string:
+            self.contract_meta_count += 1
+            return string
+        elif self.contract_meta_count != 0 and self.contract_meta_count != 5:
+            self.contract_meta_count += 1
+            return ""
+        elif self.contract_meta_count != 0 and self.contract_meta_count == 5:
+            self.contract_meta_count = 0
+            return ""
+
+        return string
+
+
 class PromptInterface(object):
     go_on = True
 
@@ -120,7 +161,7 @@ class PromptInterface(object):
                 'debugstorage {on/off/reset}'
                 ]
 
-    history = FileHistory(FILENAME_PROMPT_HISTORY)
+    history = NeoHistory(FILENAME_PROMPT_HISTORY)
 
     token_style = None
     start_height = None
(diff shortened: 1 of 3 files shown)


neo/bin/prompt.py
```diff
@@ -55,6 +55,47 @@
...
+        if len(string) == 0:
+            return string
+        command = [comm for comm in ['import wif', 'export wif', 'import nep2', 'export nep2'] if comm in string]
+        if len(command) > 0:
+            command = command[0]
+            if command in string and len(command + " ") < len(string):
+                return command + " <" + command.split(" ")[1] + ">"
+            else:
+                return string
+        elif "import contract" in string:
```
metachris: The meta-count stuff needs documentation / comments, perhaps a function docstring. What are you trying achieve here?

metachris:
Very nice! Thanks. :+1: 

Just a small question / suggestion about this meta count thing.

neo/bin/prompt.py
BSathvik: The `import contract` command has Metadata asking for name, version author, email and description that has to be removed. Since there are 5 metadata entries, the `contract_meta_count` variable keeps a count of how many metadata entries have taken place after the command `import contract`. Since we already know that there are going to be five entries after the `import contract` command, we just keep a count and not add them until all five are entered.

I will add documentation ASAP.

neo/bin/prompt.py
```diff
@@ -55,6 +55,47 @@
 FILENAME_PROMPT_HISTORY = os.path.join(PATH_USER_DATA, '.prompt.py.history')
 
 
+class NeoHistory(FileHistory):
+    contract_meta_count = 0
+
+    def append(self, string):
```
metachris: this is good! (see also the original FileHistory here https://github.com/jonathanslenders/python-prompt-toolkit/blob/313dd70224f508c43b0165fc0bfa0a80c54e6294/prompt_toolkit/history.py)

neo/bin/prompt.py
```diff
@@ -55,6 +55,47 @@
 FILENAME_PROMPT_HISTORY = os.path.join(PATH_USER_DATA, '.prompt.py.history')
 
 
+class NeoHistory(FileHistory):
```
metachris: Could you please rename it to `PromptHistory`?

neo/bin/prompt.py
metachris: Oh yeah, I hate if these names are part of the history, very good catch!

The problem here is that it wouldn't work if the users presses ctrl+c in one of these steps. I feel the proper solution there would be to not use prompt-toolkit to enter these 5 parameters, but a simple raw_input. Then we could get rid of all these steps here.

What do you think?

metachris:
Very nice, I like this a lot! 👍 Just a few more minor comments inline.

neo/bin/prompt.py
BSathvik: Yeah, I think its definitely better idea. I wasn't too comfortable with not using prompt-toolkit, since its used everywhere. I'll do that.

neo/bin/prompt.py
BSathvik: Sure

neo/bin/prompt.py
BSathvik: Thank! that's where I got most of the `append` function from.

BSathvik:
Ok, Instead of using `raw_input` or `input` i just removed the `history` object argument in the `prompt`s of the `import contract` Metadata. Which makes it look more consistent and was an easy fix.

localhuman:
That's great! I think you've arrived at an ideal implementation 😉

metachris:
Great! Thanks a lot :+1:

BSathvik:
Thanks @metachris and @localhuman! This is one of my first open source contributions. I like it, I think I'm going to keep doing more.

PR #339, opened 2018-03-18
Add initial Storage Iterator implementation

### LINKED ISSUE #211: [Feature] Add new `StorageIterator`
https://github.com/neo-project/neo/commit/a234978f2fd7cb622513969c582c6d3b054070cb

@localhuman !

ixje:
Fixes #211 

This adds the new `Storage Iterator` API for smart contracts. Note that we should wait with merging until the C# side is fixed (see the issue linked in #211). I wanted to get the feedback loop started while the C# side sorts out their implementation. 

Port code

Compiled a contract in C#, invoked using `neo-python`. Default tests currently fail because it requires updating the fixtures (see reason below) which I'm waiting with until the C# side is finalised + feedback here is finalised. 

Yes, the leveldb keys for the Storage Put/Get/Delete calls have been altered to make prefix searching possible (discussed with @localhuman) which requires rebuilding the chain + recreating the fixtures.

- [ ] Did you add any tests?


localhuman:
This is a great start! Hopefully things get fixed on the C# side ...

ixje:
So it seems like the C# side is fixed with this commit; https://github.com/neo-project/neo/commit/6d5cf6311d8e132f5edca7c24ab3df38b3224412

Most of the changes are just specific to their database wrapper so to speak. One thing I was 'worried' about for a moment are the changes to the `serialize` and `deserialize` methods of `StorageKey`. A quick reference search doesn't really show them to be used. I also did a caveman approach by putting a breakpoint on both routines and then sync up my testnet from 930K to ~980. No breakpoint was hit so I don't think it's used anywhere, but it might be good if the reader can also give it some thought if it could be used somewhere now (or possibly in the future).

If after this all you still think that the serialize/deserialize changes don't affect us, please let me know so I can go and update the fixtures and write tests

localhuman:
I don't think we'll need or use `Serialize` or `Deserialize` for storage keys.  We could make the `StorageKey` not inherit `SerializableMixin` and remove those methods so that nobody is tempted to use them in the future.

ixje:
@localhuman has to upload the new fixtures. If the build fails because of `fixtures_v6.tar.gz` not found, please manually see if there file is already there and restart the build if you notice it before me :)

ixje:
Hooray build passed! :)

So there's 3 things left
1) somehow inform users that they'll have to resync their chain once they're on a version that includes these changes. How did we do that previous times?
2) add a bootstrap file. Do you already hate me @localhuman 😆 
3) add `neo-boa` support so automated tests can be added. I currently have manually tested it by compiling and invoking the C# contract mentioned here: https://github.com/neo-project/neo-devpack-dotnet/issues/19#issuecomment-373934451

Now, for the boa support I can add a `Find(context, needle)` call to the [Storage](https://github.com/CityOfZion/neo-boa/blob/master/boa/interop/Neo/Storage.py) class, but that's not enough.

Take something like this:
```
from boa.interop.Neo.Runtime import Notify
from boa.interop.Neo.Storage import GetContext, Put, Find

def Main(args):
    ctx = GetContext()
    Put(ctx, 'my_prefixA', "1")
    Put(ctx, 'my_prefixB', "2")
    Put(ctx, 'something', "3")

    result_iterator = Find(ctx, 'my_prefix')
    while result_iterator.Next():
        Notify(result_iterator.Key()) 
```
The compiler will complain it can't find `Next()` and `Key()`. I gave it a quick try but it just kept getting new errors so I think it's best if someone that understands the neo-boa architecture well would do it. *cough* that would be you @localhuman 😆 *cough*

localhuman:
Sounds like a plan!

localhuman:
Note, i've created a new branch here https://github.com/CityOfZion/neo-python/tree/feature-storage-iterator which have these changes.  I needed to change a few things to get it working.

Also, do you have the compiled avm from C# which I can test with 😀

localhuman:
Also, for reference see https://github.com/CityOfZion/neo-boa/pull/70

localhuman:
closing in favor of #406

Issue #338, opened 2018-03-17
Fix bootstrap.py to use new user data for / chains folder


metachris:
Currently breaks / doesn't download to the correct target dir.

See also #333

metachris:
Protip: search & replace all `LEVELDB_PATH` occurrences in the project outside of `Settings.py`

PR #343, opened 2018-03-19
Update bootstrap script with new paths

dethos:
https://github.com/CityOfZion/neo-python/issues/338 - With the new user data paths, the download of the bootstrap files are failing.

Removed the hard-coded relative paths in the bootstrap script and added dynamic user paths based on the new `DATA_DIR_PATH`.

Checked it manually. Downloaded the chains on my personal computer using these changes.

Not really, I just added a call to `os.remove()` on the downloaded file since it isn't removed and stays there wasting space.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Bootstrap.py
+++ b/neo/Prompt/Commands/Bootstrap.py
@@ -7,9 +7,6 @@ import tarfile
 import shutil
 import os
 
-DEFAULT_TMP_BOOTSTRAP_FILE = os.path.join(settings.DATA_DIR_PATH,
-                                          'bootstrap.tar.gz')
-
 
 def BootstrapBlockchainFile(target_dir, download_file, require_confirm=True):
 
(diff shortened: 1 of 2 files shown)


neo/Prompt/Commands/Bootstrap.py
```diff
@@ -7,6 +7,9 @@
 import shutil
 import os
 
+DEFAULT_TMP_BOOTSTRAP_FILE = os.path.join(settings.DATA_DIR_PATH,
```
metachris: Move this const into the function, because the settings data for might be changed using the --datadir argument, which would not be used like this

neo/Prompt/Commands/Bootstrap.py
```diff
@@ -20,13 +23,17 @@ def BootstrapBlockchainFile(target_dir, download_file, require_confirm=True):
...
+                            target_dir,
+                            tmp_file_name=os.path.join(settings.DATA_DIR_PATH, 'btest.tar.gz'),
+                            tmp_chain_name='btestchain')
 
     print("bootstrap cancelled")
     sys.exit(0)
 
 
-def do_bootstrap(bootstrap_file, destination_dir, tmp_file_name='./Chains/bootstrap.tar.gz', tmp_chain_name='tmpchain'):
+def do_bootstrap(bootstrap_file, destination_dir, tmp_file_name=DEFAULT_TMP_BOOTSTRAP_FILE, tmp_chain_name='tmpchain'):
```
metachris: Could use `None` instead and then use the data dir

neo/Prompt/Commands/Bootstrap.py
```diff
@@ -92,6 +99,7 @@ def do_bootstrap(bootstrap_file, destination_dir, tmp_file_name='./Chains/bootst
         print("cleaning up %s " % tmp_chain_name)
         if os.path.exists(tmp_chain_name):
             shutil.rmtree(tmp_chain_name)
+            os.remove(tmp_file_name)
```
metachris: We might want to keep the downloaded file, for test bootstrap perhaps? Cc @localhuman

metachris:
Thanks for contributing :+2:

Issue #344, opened 2018-03-20
Current block height is always one block behind


MediaServe:
It seems that block height is always one block behind compared to the `neo-cli` implementation. This is not just slow syncing, it is actually not requesting a new block until the chain has two more blocks.

MediaServe:
I justed the changes and it's working as expected. So I'm closing this issue.

Thank you all for the quick fixes!

PR #345, opened 2018-03-20
Fixes CityOfZion/neo-python#344 block synchronization being behind with one block

MediaServe:
CityOfZion/neo-python#344
Block height is always 1 block behind in synchronization

Found a `- 1` in the HeaderHeight property that I believe should not be there. 

Tested synchornization and block retrieval, compared to `neo-cli` implementation

I don't think so


metachris:
Thanks.

A small change, but might have bigger impact. Great that you checked conformity with neo-cli! @localhuman what's your thoughts on this?

MediaServe:
It probably needs more extensive testing before merging. I'm going to do more testing and let you know my results.

localhuman:
I would not make this change this way, as this matches the current C# implementation of `HeaderHeight`.  
https://github.com/neo-project/neo/blob/master/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.cs#L34

There is probably a different solution to this issue 😉

MediaServe:
Thanks! I agree that the proposed change is more a workaround than a fix. I couldn't find any other flaws in the synchronization logic. For now the change seems to work well.  I'll keep looking for a better way to fix the issue. - Thomas Lobker

PR #346, opened 2018-03-20
Block Height fixes

localhuman:
fix for current block height lag behind other RPC implementations by 1-3 blocks

- addresses #344 

- always ask for more headers
- if you only get one header, ask for the corresponding block right away

- ran tests

- no

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,7 +11,7 @@ All notable changes to this project are documented in this file.
 - Fixed missing uri locations in ``neo/api/REST/RestApi.py`` (`PR #342 <https://github.com/CityOfZion/neo-python/pull/342>`_)
 - Fixed privatenet check by fixing the chain path for checks in Settings (`PR #341 <https://github.com/CityOfZion/neo-python/pull/341>`_)
 - Fixed ``neo-privnet.sample.wallet``
-
+- Fix for current block height lag behind other RPC implementations by 1-3 blocks
 
 [0.6.1] 2018-03-16
 ----------------------------
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -264,7 +264,7 @@ class NeoNode(Protocol):
             reactor.callLater(20, self.DoAskForMoreBlocks)
 
     def DoAskForSingleBlock(self, block_hash):
-        if not block_hash in self.myblockrequests:
+        if block_hash not in self.myblockrequests:
             message = Message("getdata", InvPayload(InventoryType.Block, [block_hash]))
             self.myblockrequests.add(block_hash)
             self.SendSerializedMessage(message)
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -225,7 +225,7 @@ class NeoNode(Protocol):
 
         hashes = []
         hashstart = BC.Default().Height + 1
-        current_header_height = BC.Default().HeaderHeight
+        current_header_height = BC.Default().HeaderHeight + 1
 
         do_go_ahead = False
         if BC.Default().BlockSearchTries > 400 and len(BC.Default().BlockRequests) > 0:
@@ -262,6 +262,11 @@ class NeoNode(Protocol):
             # self.Log("all caught up!!!!!! hashes is zero")
             self.AskForMoreHeaders()
             reactor.callLater(20, self.DoAskForMoreBlocks)
+
+    def DoAskForSingleBlock(self, block_hash):
+        #        print("DO ASK FOR SINGLE BLOCK! %s " % block_hash)
+        message = Message("getdata", InvPayload(InventoryType.Block, [block_hash]))
+        self.SendSerializedMessage(message)
 
     def RequestPeerInfo(self):
         """Request the peer address information from the remote client."""
@@ -339,8 +344,11 @@ class NeoNode(Protocol):
         if inventory is not None:
             BC.Default().AddHeaders(inventory.Headers)
 
-        if BC.Default().HeaderHeight < self.Version.StartHeight:
-            self.AskForMoreHeaders()
+        if len(inventory.Headers) == 1:
+            self.DoAskForSingleBlock(inventory.Headers[0].Hash.ToBytes())
+
+#        elif BC.Default().HeaderHeight < self.Version.StartHeight:
+        self.AskForMoreHeaders()
 
     def HandleBlockReceived(self, inventory):
         """
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -264,9 +264,10 @@ class NeoNode(Protocol):
             reactor.callLater(20, self.DoAskForMoreBlocks)
 
     def DoAskForSingleBlock(self, block_hash):
-        #        print("DO ASK FOR SINGLE BLOCK! %s " % block_hash)
-        message = Message("getdata", InvPayload(InventoryType.Block, [block_hash]))
-        self.SendSerializedMessage(message)
+        if not block_hash in self.myblockrequests:
+            message = Message("getdata", InvPayload(InventoryType.Block, [block_hash]))
+            self.myblockrequests.add(block_hash)
+            self.SendSerializedMessage(message)
 
     def RequestPeerInfo(self):
         """Request the peer address information from the remote client."""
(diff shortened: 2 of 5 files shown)


localhuman:
Have tested this out on testnet/mainnet, and seems to keep up better with other RPC nodes.  Would be interested in any thoughts :)

metachris:
Looks good. Guess we should add `str(Blockchain.Default().HeaderHeight + 1))` to the RPC calls too?

localhuman:
I don't think there's a method to get the height of the headers in JSONRPC.  The `getblockcount` method should be fine as it is https://github.com/CityOfZion/neo-python/blob/master/neo/api/JSONRPC/JsonRpcApi.py#L151

CHANGELOG.rst
```diff
@@ -12,9 +12,11 @@ All notable changes to this project are documented in this file.
 - Fixed missing uri locations in ``neo/api/REST/RestApi.py`` (`PR #342 <https://github.com/CityOfZion/neo-python/pull/342>`_)
 - Fixed privatenet check by fixing the chain path for checks in Settings (`PR #341 <https://github.com/CityOfZion/neo-python/pull/341>`_)
 - Fixed ``neo-privnet.sample.wallet``
+- Fix for current block height lag behind other RPC implementations by 1-3 blocks
 - Fixed ``bootstrap.py`` to use the specified data directory, instead of hard-coded relative paths.
 
 
+>>>>>>> 6f4b5d4feb45aa1d11ebde380e761e213b17765f
```
brianlenz: @localhuman you probably didn't mean to include this merge conflict line 😀

localhuman: Good catch :)

neo/Network/NeoNode.py
```diff
@@ -339,8 +345,11 @@ def HandleBlockHeadersReceived(self, inventory):
         if inventory is not None:
             BC.Default().AddHeaders(inventory.Headers)
 
-        if BC.Default().HeaderHeight < self.Version.StartHeight:
-            self.AskForMoreHeaders()
+        if len(inventory.Headers) == 1 and BC.Default().HeaderHeight - BC.Default().Height < 5:
+            self.DoAskForSingleBlock(inventory.Headers[0].Hash.ToBytes())
+
+#        elif BC.Default().HeaderHeight < self.Version.StartHeight:
+        self.AskForMoreHeaders()
```
brianlenz: @localhuman: I've updated a couple of our pyrest and pyrpc nodes to the latest `development` branch, and they are now pegging the CPU. It looks like this change is causing it. Previously, the `python` process would use <10% CPU on average, but now, it's averaging 100% CPU usage. I added this `elif` back into place on those nodes, and CPU usage is back to normal.

Should we just add the `elif` back here? Any other ideas on best way to handle this?

localhuman: You're supposed to review before it gets merged 😉 

That sounds good.  I'll try that out and see how it works on my end

localhuman: 1f9aa17

brianlenz: Found the commit: https://github.com/CityOfZion/neo-python/commit/4a9822e007d41709ff5dd99dafa1cff51445c6b6

That definitely helps, but I'm still seeing pretty regular CPU spikes even with this change in place. I wonder if it's just load? When the CPU spikes, our RPC and REST endpoints become unresponsive until python lets up on the CPU. The spikes are lasting for quite a while (45-60 seconds). Perhaps the issue isn't related to this changeset directly, though.

Is it because there's just a single thread processing requests, so all other requests back up and that's what leads to the unresponsiveness?

You can see it happen by monitoring: http://monitor.cityofzion.io/

You'll see the pyrpc[1-4].narrative.network/pyrest[1-4].narrative.network nodes regularly go unreachable for periods of time.

I'm using `pyrasite` to inspect the `python` process. I'm not seeing anything particularly out of the ordinary in the tracebacks. Seeing some RPC method invokes and some `PersistBlocks` calls.

This issue didn't seem to happen when we were on 0.5.9-dev, at least not nearly to this extend. You can compare to the redpulse endpoints.

@metachris, you might try upgrading to `0.6.3-dev`, and I suspect you'll see the same issue.

Any other ideas on what I might do to try to diagnose the issue?

Could it just be that with the block heights being more up-to-date now, we've exposed a performance issue with neo-python not being able to handle load?

localhuman: What are the specs of the server you're running on?

brianlenz: Yeah, I was wondering if it could be that they were actually processing RPC invokes now that the block heights are caught up with the neo-cli nodes. Previously, we were still getting almost the same volume of traffic, but I can't tell from the logs what types of RPC calls they are since that information isn't included in the logs.

Three of the nodes are 2 vCPU @ 2.2GHz with 4GB RAM and SSD.
One of the nodes has 16 HT cores @ 2.4GHz with 2GB RAM and SSD.
Memory and I/O aren't an issue; it's completely CPU bound.

Each node is averaging between 5-20 req/s, which I would think it should be able to handle without issue. There's clearly a bottleneck somewhere, though, and the fact that it's only internally using a single thread is limiting the ability to take advantage of additional CPUs. It looks like the `reactor` thread is doing all of the processing: handling all incoming requests as well as all block processing. Perhaps we could at least split out the block processing into its own thread separate from the web requests? Or will that introduce concurrency issues?

Any ideas of other profiling or diagnostics I could run to try to help pinpoint the bottleneck?

localhuman: are you running REST and JSONRPC at the same time?

brianlenz: Yeah we are. And I think @metachris is for the redpulse neo-python nodes as well.

MediaServe: I can second this experience. Running the development branch from a couple of days ago (HEAD at f96810ae) with REST and RPC enabled. It seems like in the beginning it's running smooth, but after a while it's going to 100% CPU usage, blocks are lagging behind and the REST interface becomes unresponsive. It's recovering too after a while, but only shortly.

It would be better if the web-interface would run in a seperate thread, so it won't become unresponsive when the node is busy. Not sure if this is possible at all.

localhuman: We can look into thread separation issues.  My first recommendation would be to not run both services at once.

MediaServe: I did two tests:
* Running `api_server.py`with 40 peers: After a short while the node gets slow, starts lagging on new blocks and using 100% CPU
* Running `api_server.py` with the same config, but only connect to 1 peer: After about two hours the node is still fast and CPU usage remains low.

I will let it run for a couple more hours to see if the performance stays okay.

localhuman: Its a good test, but the standard setup is 5 peers without p2p joining.

FWIW http://notifications.neeeo.org/ runs only the notification server and is using between 2 and 10 % CPU, with the standard 5 peers.

MediaServe: Initially with one peer it went a lot better. Unfortunately after about five hours the process went suddenly to 100% CPU again. Probably interesting that right when it happened I got this message (many times):

```[E 180322 18:15:31 Helper:75] COULD NOT WRITE INVENTORY HASHES argument should be bytes, buffer or ASCII string, not 'NoneType'```

After that it just stopped syncing all together (with 1 connected seed)

```[E 180322 18:15:31 Helper:75] COULD NOT WRITE INVENTORY HASHES argument should be bytes, buffer or ASCII string, not 'NoneType'
[I 180322 18:15:32 threading:864] [MainNet] Block 2053527 / 2053534
[I 180322 18:15:47 threading:864] [MainNet] Block 2053533 / 2053534
[I 180322 18:16:02 threading:864] [MainNet] Block 2053533 / 2053534
[I 180322 18:16:17 threading:864] [MainNet] Block 2053533 / 2053534```
And now about 30 minutes later it's still stuck at 2053533

brianlenz: I disabled the REST endpoint on one of our nodes. It's having the same CPU performance issue with just the RPC endpoint, so I don't think the issue results directly from running RPC and REST together. I think there is some other performance bottleneck in the neo-python node implementation. While I think that splitting out the block processing into a separate thread may help, I think there is something else fundamental going on that is causing this new CPU spike that will need to be addressed.

At this point, I'm thinking we should split this into its own issue since it's no longer directly related to this PR. I created #352 for this.

PR #348, opened 2018-03-21
Make home directory optional

brianlenz:

Not all users will have a home directory, so only attempt to make the ~/.neopython directory if the user's home directory actually exists. Without this check, startup will fail.

Implemented it.

Tested it on one of our pyapi nodes.

No.

- [ ] Did you add any tests?


localhuman:
Looks good

PR #350, opened 2018-03-22
Add gzip compression support to RPC calls

ixje:
Adds optional gzip compression if the caller supports it. This feature was recently added here: https://github.com/neo-project/neo/pull/201/commits/6e1921346326e6f89bbca4c39e6123b3eae1c060

update the json decorator used in the RPC server

added a test

nope

- [X] Did you add any tests?

### DIFF
--- a/neo/api/utils.py
+++ b/neo/api/utils.py
@@ -15,15 +15,15 @@ def json_response(func):
     @wraps(func)
     def wrapper(self, request, *args, **kwargs):
         res = func(self, request, *args, **kwargs)
-        response_data = json.dumps(res) if isinstance(res, dict) else res
+        response_data = json.dumps(res) if isinstance(res, (dict, list)) else res
         request.setHeader('Content-Type', 'application/json')
 
         if len(response_data) > COMPRESS_THRESHOLD:
-            accept_encoding = request.getHeader('Accept-Encoding')
-            if accept_encoding:
-                encodings = accept_encoding.split(',')
+            accepted_encodings = request.requestHeaders.getRawHeaders('Accept-Encoding')
+            if accepted_encodings:
+                use_gzip = any("gzip" in encoding for encoding in accepted_encodings)
 
-                if 'gzip' in encodings:
+                if use_gzip:
                     response_data = gzip.compress(bytes(response_data, 'utf-8'), compresslevel=COMPRESS_FASTEST)
                     request.setHeader('Content-Encoding', 'gzip')
                     request.setHeader('Content-Length', len(response_data))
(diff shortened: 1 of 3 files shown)


neo/api/utils.py
```diff
@@ -1,24 +1,46 @@
...
+
 
 # @json_response decorator for class methods
 def json_response(func):
     """ @json_response decorator adds header and dumps response object """
+
     @wraps(func)
     def wrapper(self, request, *args, **kwargs):
         res = func(self, request, *args, **kwargs)
+        response_data = json.dumps(res) if isinstance(res, dict) else res
```
metachris: Seeing this, we might update it to `if isinstance(res, [dict, list])`

neo/api/utils.py
```diff
@@ -1,24 +1,46 @@
...
         request.setHeader('Content-Type', 'application/json')
-        return json.dumps(res) if isinstance(res, dict) else res
+
+        if len(response_data) > COMPRESS_THRESHOLD:
+            accept_encoding = request.getHeader('Accept-Encoding')
```
ixje: The idea here is to only apply compression when the data is actually larger than a single network packet, otherwise we're just wasting CPU cycles compressing data while it still goes into the same packet.
An empty string is 49 bytes in size, every additional character is 1 byte. So we do a rough estimate of the threshold to use.

neo/api/utils.py
```diff
@@ -1,24 +1,46 @@
...
+            if accept_encoding:
+                encodings = accept_encoding.split(',')
```
metachris: i think this won't be enought, because the header might have spaces too: `Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5`

Perhaps update to `encodings = [encoding.strip() for encoding in accept_encoding.split(',')`]

metachris:
Very nice! 👍

neo/api/utils.py
metachris: Oh, shit this doesn't even fully work, because the gzip part may be ` gzip;q=1.0, *;q=0.5` (so a space and then more options)

metachris: Not sure if we want to support the gzip options. If not, we might use this:

    use_gzip = any("gzip" in encoding for for encoding in accept_encoding.split(',')

metachris:
Looks great! Wondering if we should create a special @gzip decorator for this?

ixje:
@metachris I thought about moving it into a separate decorator but then at least the order of the decorators would become important, and I'd probably have to do more type checking on the data I'm trying to compress for the decorator to be generic. I also didn't see another use-case at this point that would require the same routines used now. This led me to keep it straight forward thinking we can always extract it when needed.

Issue #202, opened 2018-01-29
Remove eval


rickyhan:
https://github.com/CityOfZion/neo-python/blob/79770902337ced902513d59fa7a3453b81f987ca/neo/Prompt/Utils.py#L118

Tommah:
I think that call to eval can be replaced with a call to ast.literal_eval.

This would almost work for the calls to eval in parse_param, but literal_eval cannot handle strings that contain a call to bytearray().

dethos:
@localhuman is there a any spec of the content (and types) that should be allowed in the transaction attributes?

I ask this because for what I can see in the code examples (PR, tests and change log), this could be solved by providing a json `array` or `object` for this argument instead of the python `list` or `dict`. This way it would be saver to parse instead of using `eval`.

localhuman:
That's a good idea.  I personally like the free form way, but we could do something like how the JSONRPC Api accepts contract parameters here:
https://github.com/CityOfZion/neo-python/blob/master/neo/api/JSONRPC/JsonRpcApi.py#L223

ixje:
solved in above pr

PR #351, opened 2018-03-22
Limit the usage of "eval()"

dethos:
This PR fixes https://github.com/CityOfZion/neo-python/issues/202 and also addresses some concerns on other 2 `eval()` calls.

Given `--tx-attr` expects data that can be handled by JSON format, the parsing of the attributes was changed and is now done by the `json` module instead of `eval()`.
For the other 2 calls to `eval()` the available functionality was reduced by limiting the available `__builtins__` ([more information can be found here about the security concerns](https://vipulchaskar.blogspot.pt/2012/10/exploiting-eval-function-in-python.html))

Tested manually (trying to exploit the `eval()` calls) and wasn't able to import and execute other functions from the `prompt`. Tests still pass using `make test`.

No, by I would recommend to re-factor the `parse_param` function, in order to completely remove the `eval` calls and parse the parameters in way that is secure.

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.6.4-dev] in progress
 -----------------------
-
+- Changed the ``eval()`` call when parsing the `--tx-attr` param to parse only json. Reduced the surface and options available on the other 2 eval calls to improve security.
 
 [0.6.3] 2018-03-21
 -----------------------


metachris:
👍 

@localhuman what do you think about those changes?

localhuman:
Perhaps we should refactor `parse_param` here too?

dethos:
@localhuman :+1: it depends if you prefer to add these "protections" now and work on the refactor later, or put it on hold until the refactor is done.

localhuman:
Think we can merge this in now, and add in the other improvements later.  👍

Issue #352, opened 2018-03-22
RPC node CPU bottleneck


brianlenz:
### Current behavior

- What is the problem?

CPU is pegged at 100% about 75% of the time on neo-python RPC nodes. Background discussion can be found here:

https://github.com/CityOfZion/neo-python/pull/346#discussion_r176271059

### Expected behavior

- What should be happening?

Normal CPU usage and stable request processing throughput.

### How to reproduce

- Please explain the steps to reproduce the problem. If you are having an issue with your machine or build tools, the issue belongs on another repository as that is outside of the scope of this project.

Start up a neo-python RPC node with a default configuration. Once the node is fully sync'd and normal traffic levels are attained, CPU spikes will happen for 45-60 seconds at a time, which makes the node unresponsive.

### Your environment

Let us know in what environment you're running into the issue:

Debian Linux, neo-python 0.6.3-dev

brianlenz:
I think I've tracked one major performance issue: https://github.com/CityOfZion/neo-python/commit/ffef8ba4cc9c5fe0eb62c2b30dc670a53e44c4c3#diff-3648ab6503ed4bb603c9432e9ebc9c2eR852.

I've definitely seen the `ToName` method in many of the hot tracebacks I've been generating in assessing performance.

I then ran a full `cProfile` of the process and let it run for a few minutes. In those 3 minutes, about 1 minute (33%) of time was spent evaluating the `ToName` function on the above line (129934 total invocations in 3 minutes of runtime). I created PR #354 to address this issue.

The overall CPU issue remains even after this improvement, so I'm still looking into what else might be done to help..

ixje:
I'm going to make an educated guess that we should also print/log this [section](https://github.com/CityOfZion/neo-python/blob/master/neo/VM/ExecutionEngine.py#L951-L1023) only when [debug logging](https://github.com/CityOfZion/neo-python/blob/master/neo/bin/prompt.py#L836) is enabled instead of by default. Python doesn't have a `switch` keyword so this will not compile to a jump table like x86/arm/mips etc would, it's going to slow down with all the `if` checks. I guess there's just more faulty smart contracts on the MainNet than I expected.

brianlenz:
Performance seems much more acceptable on the latest `development` branch now! I think we can consider this resolved until/unless we see further performance issues. I've not seen the same pattern of unresponsive nodes that we were seeing before.

Great team effort, everyone!

brianlenz:
[...]

PR #354, opened 2018-03-22
VM Engine Performance Improvements

brianlenz:
cc: @ixje 

Excessive CPU usage on RPC nodes: #352 

Note that this only helps avoid some unnecessary calls, but it doesn't solve the CPU load issue entirely.

Removed unnecessary calls to `write_log` and `ToName`.

Profiled performance before and after over a similar time period. The number of invocations reduced from 130k to 9.

No.

- [ ] Did you add any tests?

### DIFF
--- a/neo/VM/ExecutionEngine.py
+++ b/neo/VM/ExecutionEngine.py
@@ -36,9 +36,7 @@ class ExecutionEngine():
     log_file_name = 'vm_instructions.log'
     # file descriptor
     log_file = None
-
-    def is_write_log(self):
-        return settings.log_vm_instructions and self.log_file and not self.log_file.closed
+    _is_write_log = False
 
     def write_log(self, message):
         """
@@ -47,7 +45,7 @@ class ExecutionEngine():
         Args:
             message (str): string message to write to file.
         """
-        if self.is_write_log():
+        if self._is_write_log and self.log_file and not self.log_file.closed:
             self.log_file.write(message + '\n')
 
     @property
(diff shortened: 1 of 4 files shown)


localhuman:
I cant remember if debug vm is on by default.  If so it should probably be off.

brianlenz:
Pretty sure it is off by default which is why this change is needed and had an impact.

brianlenz:
Confirmed it's disabled by default, which is why this fix is needed to remove the excessive and unnecessary `ToName` invocations. You have to run `config vm-log on` to enable it.

localhuman:
Another simple change is to remove this line from `api_server.py` https://github.com/CityOfZion/neo-python/blob/master/neo/bin/api_server.py#L196

brianlenz:
Why would we want to get rid of the background logging thread?

metachris:
Yeah I wouldn't remove that, it also has neglectible impact on performance, but is very useful to see block counts in the logs.

neo/VM/ExecutionEngine.py
```diff
@@ -37,14 +37,17 @@ class ExecutionEngine():
     # file descriptor
     log_file = None
 
+    def is_write_log(self):
```
localhuman: I think one way to further speed things up determine on `__init__` a property called `_is_write_log` which is just the value of `settings.log_vm_instructions`.  

Then in line 912 you can just do `if self._is_write_log: ...` and in the `write_log` method you can check if the log file exists and if its not closed before actually writing.

brianlenz: Sure thing; done!

PR #355, opened 2018-03-23
Change VMFault reporting to only output when debug logging is enabled

ixje:
This changes VM fault reporting to only occur when `config debug` (debug logging) is enabled. There's 2 reasons for this;
1) it could help with performance as described in my comment here: https://github.com/CityOfZion/neo-python/issues/352#issuecomment-375582004
2) after the VMFault reporting was added, I saw some users on Discord who thought their neo-python install was broken because they were now notified of the errors that were actually always happening anyway. This should also give them some peace of mind.  

make VM Fault reporting dependent on `setting.log_level`

added a new test case

no

- [X] Did you add any tests?

### DIFF
--- a/neo/VM/ExecutionEngine.py
+++ b/neo/VM/ExecutionEngine.py
@@ -11,6 +11,7 @@ from neo.VM.InteropService import Array, Struct, StackItem, CollectionMixin, Map
 from neocore.UInt160 import UInt160
 from neo.Settings import settings
 from neo.VM.VMFault import VMFault
+from logging import DEBUG as LOGGING_LEVEL_DEBUG
 
 
 class ExecutionEngine():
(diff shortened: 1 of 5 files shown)


neo/VM/ExecutionEngine.py
```diff
@@ -948,6 +944,9 @@ def StepOver(self):
     def VM_FAULT_and_report(self, id, *args):
         self._VMState |= VMState.FAULT
 
+        if settings.log_level != LOGGING_LEVEL_DEBUG:
```
brianlenz: There is also a `logger.error()` output below. Are we sure we only want to output that error when debug logging is enabled? Or should this also be checking for `VMFault.THROW` and `VMFault.THROWIFNOT`? I think what you have is probably intended and proper (both for performance and for reducing log output), but wanted to raise this just in case.

PR #356, opened 2018-03-23
engine should fail when it runs out of gas

localhuman:
- @MediaServe brought to our attention [this TX](https://neoscan.io/transaction/c20359f4694dd7078f1f3d246430a92ca53d7c1acfc86b602ad9f2eec3fd1b51) which contains 6 transfers of tokens.  After this TX, there is a discrepancy between `neo-python` balance and `neo-cli` balance.  This string of invokes should run out of gas and not persist any changes, but it processes 3 of them and returns as if a successful invoke.

- currently if a smart contract execution encounters errors or runs out of gas, the engine doesn't fail. 

- update `VMState` to `Fault` upon running out of gas or encountering other issues.

- added a test

- [x] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.6.4-dev] in progress
 -----------------------
-
+- fix engine error states
 
 [0.6.3] 2018-03-21
 -----------------------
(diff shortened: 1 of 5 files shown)


localhuman:
Its also possible that this was contributing to performance issues.

PR #361, opened 2018-03-28
Fix document to indicate the new way to start prompt

melvin0008:
Documentation still points to the old way of starting prompt

Changed prompt.py to np-prompt

It is documentation changes

None

- [ ] Did you add any tests? No
- [ ] Did you run `make lint`? No 
- [ ] Did you run `make test`? No
- [ ] Are you making a PR to a feature branch or development rather than master? No 

### DIFF
--- a/docs/source/basicusage.rst
+++ b/docs/source/basicusage.rst
@@ -1,13 +1,13 @@
 Basic Usage
 -----------
 
-There are two main ways to use neo-python: ``prompt.py`` and running just the node with custom
+There are two main ways to use neo-python: ``np-prompt`` and running just the node with custom
 code.
 
-prompt.py
+np-prompt
 """""""""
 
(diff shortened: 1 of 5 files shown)


docs/source/basicusage.rst
```diff
@@ -17,7 +17,7 @@ Show help with all available arguments:
 
 ::
 
-    $ python prompt.py -h
+    $ python np-prompt -h
```
brianlenz: I don't think we want `python` here, do we?

localhuman: 0380ad6

brianlenz: https://github.com/CityOfZion/neo-python/commit/bf8f00c0d9f559b9459b76422d0b2f9ce6969f2d 👍 😁

PR #366, opened 2018-03-29
"getpeers" JSON RPC method

dethos:
According to issue #189, `getpeers` RPC method was not yet implemented. I added the possible functionality for it.
I added the functionality based on the methods used by the equivalent commands on the `prompt`.
Manually, plus automated tests 
No

- [x] Did you add any tests?


localhuman:
Looks great!

metachris:
Awesome

jcrubino:
I am trying to map Neo nodes to countries by their IP address.
Is there a way to query peers of peers using the json_api in any of the clients?

PR #368, opened 2018-03-30
IsWalletTransaction fix

hal0x2328:

In IsWalletTransaction in neo/Wallets/Wallet.py, as part of evaluating whether a transaction belongs to a wallet or not, neo-python is currently comparing the VerificationScript to all the contract scripthashes in the wallet. However this will never return true, because a scripthash is by definition a hash of a script so it has already gone through the hashing process before the comparison, even if they were the same scripts to start with. As a result, ```wallet verbose``` was not showing the complete set of transactions for added contract addresses and watch-only addresses (outbound transactions were missing)

Changed the comparisons to compare script to script for contracts, and scripthash to scripthash for watch_only addresses where we don't know what the original VerificationScript looks like.

Tested it manually on a local wallet

No

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.6.5-dev] in progress
 -----------------------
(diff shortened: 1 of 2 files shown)


localhuman:
Looks good, thanks!

PR #373, opened 2018-04-03
Check Python version in Settings.py

metachris:

Make sure that at run with Python 3.6+

Adding a System check in Settings.py

Test

Won't run with Python <3.6

- [ ] Did you add any tests?


metachris:
Just keep seeing this in the channels as question, wonder why we didnt have it already :P

Issue #375, opened 2018-04-04
Contract Migrate fails


skkiran:
### Current behavior

My c# migrate method.
```
    private static string migrate(byte[] newScript, string version)
    {
        if (isOwner())
        {
            Contract.Migrate(newScript, new byte[] { 0x07, 0x10 }, 0x07, true, "Name", version, "Company", "email", "description");
            return "true";
        }
        return "false";
    }
```
calling it from neon-js using:
```
const sb = Neon.create.scriptBuilder();
sb.emitAppCall(oldContractHash, 'migrate', [account.publicKey, script, neon_js.u.str2hexstring(nextVersion)]);

const tx = Neon.create.invocationTx(balance, [], sb.str, 500);
const signedTx = Neon.sign.transaction(tx, account.privateKey);
rpc.Query.sendRawTransaction(signedTx.serialize(true)).execute(url).then((res) => {
  console.log("success", res);
}).catch((e) => {
  console.log("error sending raw transaction", e);
});
```

I see following error in neopy (with sc-events on)
```
> [E 180404 14:32:07 ExecutionEngine:835] COULD NOT EXECUTE OP: object of type 'method' has no len() b'h' SYSCALL
     [E 180404 14:32:07 ExecutionEngine:836] object of type 'method' has no len()
         Traceback (most recent call last):
           File "/neo-python/neo/VM/ExecutionEngine.py", line 829, in StepInto
             self.ExecuteOp(op, self.CurrentContext)
           File "/neo-python/neo/VM/ExecutionEngine.py", line 201, in ExecuteOp
             if not self._Service.Invoke(call, self):
           File "/neo-python/neo/VM/InteropService.py", line 308, in Invoke
             return func(engine)
           File "/neo-python/neo/SmartContract/StateMachine.py", line 399, in Contract_Migrate
             self._contracts_created[hash.ToBytes()] = UInt160(data=engine.CurrentContext.ScriptHash)
           File "/usr/local/lib/python3.6/dist-packages/neocore/UInt160.py", line 6, in __init__
             super(UInt160, self).__init__(num_bytes=20, data=data)
           File "/usr/local/lib/python3.6/dist-packages/neocore/UIntBase.py", line 17, in __init__
             if len(data) != num_bytes:
         TypeError: object of type 'method' has no len()
     [E 180404 14:32:07 ExecutionEngine:835] COULD NOT EXECUTE OP: Invalid list operation b'u' DROP
[...]

PR #377, opened 2018-04-04
#375 Exception during Contract.Migrate

skkiran:
Tested it with my testnet and it works fine after the fix. i could see all the data that was in old-contract in the new-contract and the old-contract was also dropped.


localhuman:
Thanks for the contribution!

PR #381, opened 2018-04-07
Development

hal0x2328:
The notification DB were showing the previous block height instead of the actual block height the notification tx was part of. Also, refund() notifications were not being persisted in the notification DB, which would be useful information for anyone running a token crowdsale.

I added a check for NotifyType.REFUND to the notification DB persist code, and incremented the block height argument value by one when creating SmartContractEvents in the StateReader/StateMachine code.

Ran np-prompt for a while on mainnet and manually compared reported event block heights with the height of the reported transaction, also cross-checking against stored notifications for that block to make sure height was properly persisted matching the transaction height as well.

No, but the *_Notif bootstrap files should probably be rebuilt after merging so that event block heights will be correctly stored for historical events

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,8 @@ All notable changes to this project are documented in this file.
 
 [0.6.8-dev] in progress
 -----------------------
+- fixed sc-events and notification DB showing previous block height instead of final block height of event
+- persist refund() notify events in notification DB
 
 
 [0.6.7] 2018-04-06
--- a/neo/Implementations/Notifications/LevelDB/NotificationDB.py
+++ b/neo/Implementations/Notifications/LevelDB/NotificationDB.py
@@ -117,8 +117,9 @@ class NotificationDB():
         if not isinstance(sc_event, NotifyEvent):
             logger.info("Not Notify Event instance")
             return
-        if sc_event.ShouldPersist and sc_event.notify_type == NotifyType.TRANSFER:
-            self._events_to_write.append(sc_event)
+        if sc_event.ShouldPersist:
+            if sc_event.notify_type == NotifyType.TRANSFER or sc_event.notify_type == NotifyType.REFUND:
+                self._events_to_write.append(sc_event)
 
     def on_persist_completed(self, block):
         """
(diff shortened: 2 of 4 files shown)


brianlenz:
cc: @hal0x2328 @localhuman

I tried upgrading our nodes to 0.6.9 today, and our smart contract event handler stopped working, which I believe is caused by the changes here. Specifically, I believe this change is the problem:

https://github.com/CityOfZion/neo-python/commit/d246c879256b916822223537b4de6b929ede3c22#diff-ea9a12ba295d6241603556797d390522R183

Our event handler listens for `Runtime.Notify` events and then does the following:

```
blockchain.GetHeaderByHeight(event.block_number).Timestamp
```

where `blockchain` is the `LevelDBBlockchain`. This used to work previously without issue, but now it's resulting in an error:

```
  File "handler.py", line 155, in do_sc_notify
    timestamp = self.blockchain.GetHeaderByHeight(block_number).Timestamp
AttributeError: 'NoneType' object has no attribute 'Timestamp'
```

> This might be a breaking change for anyone who is subscribing to events based on block height, so we will need to communicate that.

We don't subscribe to events based on block height, but we do look up the current block by height when events are processed. I wouldn't think this should be an issue, should it? Am I missing something or is this just a bug? It seems to me that the current block an event is being emitted for should always be found...

hal0x2328:
Not sure why that would happen - in `Persist` dispatching events is the very last thing that happens:

```            wb.put(DBPrefix.SYS_CurrentBlock, block.Hash.ToBytes() + block.IndexBytes())
            self._current_block_height = block.Index
            self._persisting_block = None

            self.TXProcessed += len(block.Transactions)

            for event in to_dispatch:
                events.emit(event.event_type, event)
```
Can you add some debugging there to see what self._current_block_height is set to in relation to the emitted events right after?

brianlenz:
Sure, I tried this out. Here's one example:

```
[I 180519 09:49:10 LevelDBBlockchain:811] Persisted _current_block_height 2287579
```

Then, in the event handler:

```
[I 180519 09:49:10 handler:157] Loading block 2287579
[I 180519 09:49:10 handler:159] Block 2287579 loaded: None
```

So, the `LevelDBBlockchain` doesn't have the block that was just persisted...? I'm still looking to see if I can identify how that might be. Any ideas?

brianlenz:
I think figured it out. Give me a bit to track it down. Ultimately, I don't think it's related to the change here. Hoping to have the change that caused the issue pinpointed shortly...

brianlenz:
I traced the root of the issue, but I don't have a fix yet. It's not related to the changes here, so apologies for confusing this. Issue being tracked here now: #424.

Thanks!

PR #382, opened 2018-04-07
Merge recent updates  into consenus branch

localhuman:
- merges changes since this branch was last updated into this branch
- lots of changing `neo.IO` to `neocore.IO`
- merges in networking updates for relaying blocks

- fun

- ran tsets

- no

- [ ] Did you add any tests?


ixje:
just got back from holidays. Should I still review?

PR #384, opened 2018-04-09
Fix for deleted NEP-5 tokens persisting in wallet file after re-open

hal0x2328:
When using the `wallet delete_token` command, tokens were removed from the in-memory wallet object but remained in the wallet file, appearing again in memory after the wallet was closed and re-opened. This is because the DeleteNEP5Token function is passed a scripthash instead of a token object, and the wallet db query fails silently while trying to dereference the scripthash from the non-existent object.

I changed the code to pass the NEP5Token object to the db Get function, as a copy was conveniently available as the return from the `super` call right above it.

Tested it manually and also modified the existing import_token unit test to verify both importing and removal of a token from a wallet file

No

- [x] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.6.8-dev] in progress
 -----------------------
+- fix for token_delete command not removing tokens from wallet file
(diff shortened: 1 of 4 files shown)


localhuman:
👍 looks like its need a `make lint`fix

hal0x2328:
Fixed.

PR #397, opened 2018-04-25
Fixing Contact.Migrate by storing state prefixed with contract script…

imcu:
Currently Contract.Migrate does not work on neo-python. My change fixes it.

I added the contract script hash as a prefix to storage keys so they can be migrated to the newly created contract.

Tested locally by implementing a Migrate feature to a smart contract

The new type of keys for storage might impact performance.

- [ ] Did you add any tests?


localhuman:
closed in favor of #408

PR #398, opened 2018-04-26
Blockchain Persist performance tweak

ixje:
The `Persist()` call in `LevelDBBlockchain.py` loops over the inputs and unnecessary recalculates the `systemshare` every time

https://github.com/CityOfZion/neo-python/blob/ea794abbf941638fbe957d511614f3d66a7c2ef1/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py#L677

add a small cache
**Before**
> (venv) Eriks-Air:neo-python erik$ python -m timeit -s 'from neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain import LevelDBBlockchain' 'LevelDBBlockchain.SystemShare().Hash.ToBytes()'
10000 loops, best of 3: 130 usec per loop

**After**
> (venv) Eriks-Air:neo-python erik$ python -m timeit -s 'from neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain import LevelDBBlockchain' 'LevelDBBlockchain.SystemShare().Hash.ToBytes()'
100000 loops, best of 3: 3.01 usec per loop

`timeit` test

no

- [ ] Did you add any tests?

### DIFF
--- a/neo/Core/Blockchain.py
+++ b/neo/Core/Blockchain.py
@@ -17,6 +17,7 @@ from collections import Counter
 from neocore.Fixed8 import Fixed8
 from neocore.Cryptography.ECCurve import ECDSA
 from neocore.UInt256 import UInt256
+from functools import lru_cache
 
 
 class Blockchain(object):
@@ -58,6 +59,7 @@ class Blockchain(object):
         return Blockchain.__validators
 
     @staticmethod
+    @lru_cache(maxsize=2)
     def SystemShare():
         """
         Register AntShare.
@@ -73,6 +75,7 @@ class Blockchain(object):
                                    amount, 0, owner, admin)
 
     @staticmethod
+    @lru_cache(maxsize=2)
     def SystemCoin():
         """
         Register AntCoin
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,6 +13,7 @@ All notable changes to this project are documented in this file.
 - add Runtime.Serialize/Deserialize support for MAP
 - fix for debug breakpoints not being cleared.
 - add VERIFY op to ExecutionEngine
+- add caching to systemshare and systemcoin creation to help in block persistence.
 
 [0.6.7] 2018-04-06
 -----------------------


neo/Core/Blockchain.py
```diff
@@ -58,6 +59,7 @@ def StandbyValidators():
         return Blockchain.__validators
 
     @staticmethod
+    @lru_cache(maxsize=2)
```
ixje: fyi; `maxsize=2` is the smallest cache size configurable.

localhuman:
That's great!

Perhaps we could make it even faster by storing a reference of `Blockchain.SystemShare().Hash.ToBytes()` ( same with SystemCoin) in LevelDBBlockchain?

ixje:
@localhuman it turns out that's actually slower 🤔 I ran the test a couple of times and it averages around 3.45 (with the lowest time at 3.18 usec). It's just a small difference between the two solutions. 

> (venv) Eriks-Air:neo-python erik$ python -m timeit -s 'from neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain import LevelDBBlockchain' 'LevelDBBlockchain.SystemShare().Hash.ToBytes()'
100000 loops, best of 3: 3.45 usec per loop

for
```
    @staticmethod
    def SystemShare():
        """
        Register AntShare.

        Returns:
            RegisterTransaction:
        """
        if not Blockchain._SystemShare:
            amount = Fixed8.FromDecimal(sum(Blockchain.GENERATION_AMOUNT) * Blockchain.DECREMENT_INTERVAL)
            owner = ECDSA.secp256r1().Curve.Infinity
            admin = Crypto.ToScriptHash(PUSHT)
            Blockchain._SystemShare = RegisterTransaction([], [], AssetType.GoverningToken,
                                                          "[{\"lang\":\"zh-CN\",\"name\":\"小蚁股\"},{\"lang\":\"en\",\"name\":\"AntShare\"}]",
                                                          amount, 0, owner, admin)
        return Blockchain._SystemShare
```

localhuman:
I was thinking of caching/storing the actual `hash.ToBytes` value

localhuman:
We can circle back on further performance changes, this is good for now!

PR #403, opened 2018-04-28
Remove redundant inheritance from object and redundant empty inheritance parentheses 

iddoberger:
Since we are in Python 3.x there is no need for inheriting explicitly from object (old-style classes of Python 2.x are gone)
In addition, there is no need for empty parentheses after a class declaration


localhuman:
💯

PR #404, opened 2018-04-28
Replace import star with explicit imports

iddoberger:
- in some places the star importing brought to the namespace a different package 
e.g, in neo.Core.BlockBase:
from neocore.Cryptography.Helper import *
->
import binascii
from neocore.Cryptography.Helper import bin_dbl_sha256

- in some places the star importing was redundant 
e.g, in neo.Core.Blockchain:
from neocore.Cryptography.Crypto import * 
-> 
∅

- in one place I think the star import is better than explicitly importing all the used objects:
neo.VM.ExecutionEngine

### DIFF
--- a/neo/Core/BlockBase.py
+++ b/neo/Core/BlockBase.py
@@ -1,6 +1,7 @@
 import ctypes
 from .Mixins import VerifiableMixin
-from neocore.Cryptography.Helper import *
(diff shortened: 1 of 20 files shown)


metachris:
Very nice, thanks! 

@localhuman any thoughts?

Issue #412, opened 2018-05-10
string arguments in load_run are sometimes parsed incorrectly


fyrchik:
### Current behavior

I have experienced a very STRANGE bug. I want to provide a string argument to smart-contract, However sometimes I get an error (even BEFORE execution starts, so the problem is in neo-python shell):
`Could not load script argument should be bytes, buffer or ASCII string, not 'float'`

Strings, that caused this error: "ny", "anf", "puhnaty"
Strings, that didnt cause error: "kn", "antz" ...

See the pattern?
Exactly! If we cast strings to byte array and then to hex, strings in first line will contain only numbers and 'e', so they can be interpreted as float (!)
Seems, like it is happening in `neo/Prompt/Utils.py:parse_param`, where `eval` takes as argument hex representation of string, however I had not found a quick way to fix it.

### Expected behavior

All strings are processed correctly.

### How to reproduce

Probably any load_run *.avm where one of arguments is string.

### Your environment

Let us know in what environment you're running into the issue:

Linux dzeta-pc 4.13.0-41-generic #46-Ubuntu SMP Wed May 2 13:38:30 UTC 2018 x86_64 GNU/Linux
neo-python v0.7.0-dev
Python 3.6.3

localhuman:
Thanks for submitting the issue!

For me, this only happens when using these strings in a list in the arguments.  So the following works: `testinvoke  {hash} ny []`  whereas this does not: `testinvoke {hash} mystr ['ny']`

localhuman:
I've been able to avoid this issue by commenting out the following line:
https://github.com/CityOfZion/neo-python/blob/master/neo/Prompt/Utils.py#L185

If you're interested in making a PR for the fix that would be great!  Otherwise we can get it done.  Will need to add some more tests to catch cases like these 😆

fyrchik:
@localhuman yes, removing return indeed fixes all of my problems :grinning:. I had created a PR.

dethos:
If this issue is solved, perhaps would be better to close it.

PR #413, opened 2018-05-11
fix error with string parameter parsing

fyrchik:
Some string parameters, which can be parsed as float in hex, e.g. `binascii.hexlify(b'anf') == b'616e66'`, are parsed incorrectly.
This PR fixes https://github.com/CityOfZion/neo-python/issues/412

removed unwanted return
I have also added proper (?) formatting to `nodes` command.

I have tried to invoke my contract with 'problem' strings (as well as with normal).

No.

- [-] Did you add any tests?
- [+] Did you run `make lint`?
- [+] Did you run `make test`? (executing right now)
- [+] Are you making a PR to a feature branch or development rather than master?
- [+] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

### DIFF
--- a/neo/Prompt/Utils.py
+++ b/neo/Prompt/Utils.py
@@ -174,9 +174,7 @@ def parse_param(p, wallet=None, ignore_int=False, prefer_hex=True):
 
     try:
         val = eval(p, {"__builtins__": {'bytearray': bytearray, 'bytes': bytes}}, {})
-        if type(val) is bytearray:
-            return val
-        elif type(val) is bytes:
+        if type(val) is bytes:
             # try to unhex
             try:
                 val = binascii.unhexlify(val)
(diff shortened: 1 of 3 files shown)


neo/Prompt/Utils.py
```diff
@@ -185,7 +185,6 @@ def parse_param(p, wallet=None, ignore_int=False, prefer_hex=True):
             # now it should be unhexxed no matter what, and we can hex it
             return val.hex().encode('utf-8')
 
-        return val
```
metachris: If we don't remove the value here, should we perhaps also raise an Exception? But then we have an exception handler right afterwards that swallows it, so not sure if necessary.

metachris:
Thanks for contributing. Guess it works as expected. @localhuman any thoughts?

neo/Prompt/Utils.py
localhuman: I actually found in further testing that if we don't check for boolean types then some things can get messed up ( `False` parsed as `True`).  So I think we'll need to add a check for `if isInstance(val, bool)`

fyrchik: I have updated PR and changed order of conditions and use only `type(val) is not float` to minimize risks.

localhuman:
Great work! Thanks 💯

PR #416, opened 2018-05-15
Update install instructions on the docs

dethos:
The instructions to install `neo-python` present on the documentation (`docs` folder) were not up to date. 
I just synced with the `README.rst` so the content is presented in the same (and correct) order, then added the missing `Install from PyPi` section. 
Compiled the sphinx docs and checked
No

**Note**: To make it easier to keep the documentation up to date and avoid duplication, I would suggest that in the future there should be a single source of truth for the documentation. For example, the `README.rst` should point to the docs instead of having install instructions there.

- [ ] Did you add any tests?


localhuman:
Thanks!

PR #418, opened 2018-05-17
RPC JSON Property Tweaks

brianlenz:

As reported by @lucianoengel in Discord #neoscan-api, updated `next_consensus` JSON property to `nextconsensus`. Also completed an audit of the block and transaction JSON to make sure it complies with the C# implementation. This included adding the missing `nonce` property. The primary outstanding issue (which I've documented with todo comments) is to address incorrect size calculations.

Implemented the changes.

Tested it. Added tests to confirm `nonce` and `nextconsensus` are being generated properly.

I need some direction/confirmation on how to address the block and transaction size issue to address the outstanding todo comments. Curious to get your thoughts: @localhuman @metachris 

- [X] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,15 +4,16 @@ Changelog
 All notable changes to this project are documented in this file.
 
 [0.7.0] in progress
------------------------
+-------------------
 - fix a bug with smart-contract parameter string parsing `#412 <https://github.com/CityOfZion/neo-python/issues/412>`_
 - fix ``StateMachine.Contract_Migrate`` and add tests
 - add ability to attach tx attrs to build command and testinvoke.  altered tx attr parsing
 - updated the install instructions present on ``docs``
+- fix issues with some JSON-RPC properties `#418 <https://github.com/CityOfZion/neo-python/issues/418>`_
 
 
 [0.6.9] 2018-04-30
------------------------
+------------------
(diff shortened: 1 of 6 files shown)


localhuman:
Could we remove the `Size` from the json for now, since as @ixje points out it is incorrectly/incompletely implemented at the moment?

brianlenz:
Yeah, for sure. Sounds like that may be the best plan, and then I'll create a separate issue to redesign the `Size` calculation design per @ixje's great recommendations above.

ixje:
perhaps for readability we might want to have some utility class along the lines of
```
uint8 = 1
uint16 = 2
uint32 = 4
uint160 = 20
uint256 = 32
```

then reading `Size()` becomes a little bit more intuitive without `magic values`. from
```
def Size(self):
    return 4 + 32
```
to 
```
def Size(self):
    return uint32 + uint256
```
Or maybe we should just have a very clear docstring for  `Size()`? Something like 
> Returns the amount of bytes the serializable data of `self` consists off 

something like that?🤔

brianlenz:
We aren't currently including `size` in the JSON, so nothing to do there.

I created #427 to address the `size` issue, so I think this particular PR can be considered done?

localhuman:
Looks good, and thanks for creating the issue for size calculation

PR #419, opened 2018-05-17
High Volume GAS Claim Improvement

brianlenz:

Addresses with thousands of UTXOs may have issues with transaction size preventing GAS claims from working in neo-python.

@localhuman: will be curious to hear your thoughts on this. Does the approach seem reasonable? Necessary? Anything that could be better about it?

Claim GAS for at most 200 UTXOs at a time.

Tested and used the process.

No.

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,7 @@ All notable changes to this project are documented in this file.
 - fix ``StateMachine.Contract_Migrate`` and add tests
 - add ability to attach tx attrs to build command and testinvoke.  altered tx attr parsing
 - updated the install instructions present on ``docs``
+- added support for chunking through GAS claims 200 UTXOs at a time `#419 <https://github.com/CityOfZion/neo-python/issues/419>`_
 
 
 [0.6.9] 2018-04-30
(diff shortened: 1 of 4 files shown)


localhuman:
Is there an example / test /use case where addresses with lots of claimable are currently failingl?

brianlenz:
I've had this happen a couple of times, and the only fix I found was to chunk through the claims incrementally (which has worked without issue). The main use case for us was with our token sale SC address, which had 2k-3k claims available. At two different points in time (mid-sale and post-sale), I had tried using the standard claim process, and the tx would never go through. It would break the state of the wallet, and I'd have to rebuild the wallet. I suspected size was the issue, so I started playing around with it. I tried chunks of 500 claims at a time, and that didn't work. I found the only way around it was to chunk through the transactions in smaller increments, and 200 seemed to work in our case. It's definitely not scientific in any way, which is why I was curious to get your thoughts. If this is something that shouldn't technically be necessary, then perhaps there is some separate issue that needs to be addressed (and this approach scrapped altogether).

localhuman:
Thanks for the explanation.  I'd be interested in the root cause of the issue, but would have to be able to reproduce to determine that.  I wonder if maybe we're running into the max size of a transaction?  

Perhaps for now it would be best to allow users to pass in the size of the chunk, otherwise it defaults to current behavior?

brianlenz:
I like that idea! I'll take a crack at it :)

brianlenz:
Should be good to go now, I believe.

localhuman:
Thanks for the changes!

PR #420, opened 2018-05-17
Support RPC and REST in parallel

brianlenz:

RPC and REST endpoints weren't supported in parallel.

Changed it back to the previous implementation so they can both run in parallel.

Tested it.

No.

- [ ] Did you add any tests?


ixje:
do you happen to remember why we moved away from that approach in the first place?

metachris:
Iirc Tom thought we wouldn't run them in parallel anyway, and we've recently revisited this in #neopython

Issue #217, opened 2018-02-11
importcontract fails with returntype >= 16


igormcoelho:
Perhaps I misunderstood something, but I cannot import a contract with return type >= 16 (example, Void = ff = 255).
What happens is that "def GatherContractDetails(function_code, prompter)" will call "return generate_deploy_script(function_code.Script, name, version, author, email, description,                                   function_code.ContractProperties, ord(function_code.ReturnType), function_code.ParameterList)" on line 199 of LoadSmartContract.py, and since returntype is passed in ord() function, it will be set as an int < 256.
Function "def generate_deploy_script" on line 204 of LoadSmartContract.py will receive return_type and push it "sb.push(return_type)" at line 220. Remember that return_type is always "int" in this case (because of ord() function called previously), but push function in ScriptBuilder.py will enter "if type(int) is int or ..." and will fail "else" command: return self.push(binascii.hexlify(data.ToByteArray())) because "int" cannot have ToByteArray.
I believe this code only works for return types < 16, such as ByteArray 0x05, Boolean 0x01, etc (because it wont  enter "else" condition), and I think an easy solution is to force return value to be BigInteger (in line 220 of LoadSmartContract.py). It will become: "sb.push(BigInteger(return_type))". 
I can make a pull request, but since it seems a rather complicated issue, I prefer to ask for opinions first :)

localhuman:
Yes, this has been fixed.

vncoelho:
It is not in the master yet.

The error is still happening here.

igormcoelho:
The problem was solved indeed, I can confirm it. The solution was the adoption of a bytearray( ff ), which is quite similar to my older proposition using BigInteger (ff)... that's what we have used successfully in neocompiler.io for the past few months.
Vitor noticed that another issue indeed was happening, related to #422, but not in deploy time... it seemed databased was being blocked when a wrong readbyte was performed in a further invocation.. but that seems to be solved in 'development' branch now (not in current master https://github.com/CityOfZion/neo-python/commit/42e0bb41866d0b2c7aa3ee460a8f8ce3ae2f1f22).

As a solution, we are using development branch now, even for production :)

igormcoelho:
In fact, I think a better solution would be this: https://github.com/neo-project/neo/pull/253

PR #422, opened 2018-05-18
Fix error with invalid return-types 

dethos:
When trying to reproduce issue #217, I've imported a smart contract that contained values that caused the generated block to produce this error on my `neo-python` node:
```
neo> [I 180517 15:05:58 LevelDBBlockchain:842] Could not persist block bytes must be in range(0, 256) 
Unhandled error in Deferred:
```

and more detailed:
```
  File "/home/user/neo-python/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py", line 838, in PersistBlocks
    self.Persist(block)
  File "/home/user/neo-python/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py", line 805, in Persist
    contracts.Commit(wb)
  File "/home/user/neo-python/neo/Implementations/Blockchains/LevelDB/DBCollection.py", line 65, in Commit                                                         
    self.DB.put(self.Prefix + keyval, self.Collection[keyval].ToByteArray())
  File "/home/user/neo-python/neo/Core/State/StateBase.py", line 61, in ToByteArray
    self.Serialize(writer)
  File "/home/user/neo-python/neo/Core/State/ContractState.py", line 138, in Serialize
    self.Code.Serialize(writer)
  File "/home/user/neo-python/neo/Core/FunctionCode.py", line 84, in Serialize
    writer.WriteByte(self.ReturnType)
  File "/home/user/.virtualenvs/neopython/lib/python3.6/site-packages/neocore/IO/BinaryWriter.py", line 86, in WriteByte
    self.stream.write(bytes([value]))
ValueError: bytes must be in range(0, 256)
```
This problem prevented the node of continuing with the chain synchronization.
Inspecting it further with help from @ixje and @localhuman the source of the problem was traced back to an invalid return_type on `FunctionCode`:

```
Return type: -1 type: <class 'int'>
```
Made sure the `return_type` is within the valid range, otherwise make it `void` (`ff` or `255`)
Manually on my private network
No

- [ ] Did you add any tests?

### DIFF
--- a/neo/Core/FunctionCode.py
+++ b/neo/Core/FunctionCode.py
@@ -44,6 +44,8 @@ class FunctionCode(SerializableMixin):
 
         if return_type and not isinstance(return_type, int) and len(return_type) > 1:
             return_type = return_type[0:1]
+        elif isinstance(return_type, int) and not 0 < return_type <= 255:
+            return_type = 255
 
         self.ReturnType = return_type
 


localhuman:
👍

igormcoelho:
@localhuman and @dethos  I don't exactly what patch was done in Python, but the same problem is still happening in C#. I provided a PR with a permanent solution to the problem: https://github.com/neo-project/neo/pull/253
Please, take a look when you can. Vitor and I was struggling with this issue, but now we found a solution.

PR #423, opened 2018-05-19
add VarSize helper for Brian

ixje:
Add a quick helper method from old consensus branch to avoid @brianlenz having to recreate or search for it when dealing with the `Size()` issue in #418

Note to @brianlenz , you'll likely need to inline it as shown below to avoid circular imports
https://github.com/CityOfZion/neo-python/blob/b33cf1488e8e33aa5618dc412732c2e4e717b065/neo/Core/Witness.py#L42-L50

add helper method from old consensus tree

not

no

- [ ] Did you add any tests?

### DIFF
--- a/neo/IO/Helper.py
+++ b/neo/IO/Helper.py
@@ -53,3 +53,23 @@ class Helper:
         tx = Transaction.DeserializeFrom(reader)
 
         return tx
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


ixje:
I will be handling this part in another PR where I will address #427

PR #433, opened 2018-05-24
Added mint event type to NotificationDB

metachris:

Added mint events to the NotificationDB.

Updated `NotificationDB.py` and `SmartContractEvent.py`

Added a test, but not live tested.

Couldn't get any current mint events on my bootstrapped chains. Is there an easy way to test it, besides on a private net and deploying a NEP-5 SC? (Reminds me, we should a NEP-5 SC to the privatenet default)

No

- [X] Did you add any tests?

### DIFF
--- a/neo/SmartContract/SmartContractEvent.py
+++ b/neo/SmartContract/SmartContractEvent.py
@@ -162,12 +162,10 @@ class SmartContractEvent(SerializableMixin):
 
 
 class NotifyType:
-
     TRANSFER = b'transfer'  # OnTransfer = RegisterAction('transfer', 'to', 'from', 'amount')
-
     APPROVE = b'approve'  # OnApprove = RegisterAction('approve', 'addr_from', 'addr_to', 'amount')
-
     REFUND = b'refund'  # OnRefund = RegisterAction('refund', 'to', 'amount')
+    MINT = b'mint'  # OnMint = RegisterAction('mint', 'addr_to', 'amount')
 
 
 class NotifyEvent(SmartContractEvent):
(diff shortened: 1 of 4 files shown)


neo/SmartContract/SmartContractEvent.py
```diff
@@ -231,11 +229,18 @@ def __init__(self, event_type, event_payload, contract_hash, block_number, tx_ha
                     self.amount = int(BigInteger.FromBytes(event_payload[3])) if isinstance(event_payload[3], bytes) else int(event_payload[3])
                     self.is_standard_notify = True
 
-                elif plen == 3 and self.notify_type == NotifyType.REFUND:
+                elif self.notify_type == NotifyType.REFUND and plen >= 3:  # Might have more arguments
```
metachris: Saw some contracts that have 4 arguments, using this signature:

    OnRefund = RegisterAction('refund', 'addr_to', 'amount', 'asset')

localhuman: Interesting, I wonder what the 4th argument is?

metachris: whether it's gas or neo?

neo/SmartContract/SmartContractEvent.py
```diff
@@ -231,11 +229,18 @@ def __init__(self, event_type, event_payload, contract_hash, block_number, tx_ha
...
                     self.addr_to = UInt160(data=self.event_payload[1]) if len(self.event_payload[1]) == 20 else empty
                     self.amount = int(BigInteger.FromBytes(event_payload[2])) if isinstance(event_payload[2], bytes) else int(event_payload[2])
                     self.addr_from = self.contract_hash
                     self.is_standard_notify = True
+
+                elif self.notify_type == NotifyType.MINT and plen == 3:
+                    self.addr_to = UInt160(data=self.event_payload[1]) if len(self.event_payload[1]) == 20 else empty
+                    self.amount = int(BigInteger.FromBytes(event_payload[2])) if isinstance(event_payload[2], bytes) else int(event_payload[2])
+                    self.addr_from = self.contract_hash
```
brianlenz: Is this the correct approach? I know that there isn't a defined standard around minting yet, but many smart contracts use a `null` `from` address for the minting operations. I realize this is just for the mint SC event notification, but I do wonder if this would be expected for someone writing an event handler. Probably not a big deal, but I thought I'd raise it as a consideration anyway...

ThomasLobker:
In my database I'm considering the all zeroes public key as the "minting source", since most contracts seem to have used that approach. I would encourage this approach to be the standard, to keep behavior uniform.

PR #442, opened 2018-05-30
Allow SmartContract to use a hash with or without 0x prefix

metachris:

Allow SmartContract to use a hash with or without 0x prefix as output by neo-python contract operations.

Check for `0x` prefix and remove it

Manual test

No

- [ ] Did you add any tests?


localhuman:
looks like a `pycodestyle` issue :)

PR #443, opened 2018-05-31
CheckWitness must FAULT if input is bad

igormcoelho:

When a wrong input is passed to CheckWitness (empty byte array, for example) it must FAULT (not return false).

I detected two different executions from python and C#, then I located the source of the problem, a single line.

I tested on Eco platform (neocompiler.io), for both scenarios.

Now it is consistent with the original C# design.

- [ ] Did you add any tests?
This question is very good! I always forget to put to development branch... :)

### DIFF
--- a/neo/SmartContract/StateReader.py
+++ b/neo/SmartContract/StateReader.py
@@ -288,7 +288,7 @@ class StateReader(InteropService):
             point = ECDSA.decode_secp256r1(hashOrPubkey, unhex=False).G
             result = self.CheckWitnessPubkey(engine, point)
         else:
-            result = False
+            return False
 
         engine.EvaluationStack.PushT(result)
 


igormcoelho:
Please @localhuman  @metachris @canesin accept this in master as soon as possible... we depend on it to continue our tests efficiently on Eco platform neocompiler.io.

igormcoelho:
please @localhuman  @metachris @canesin accept this in master as soon as possible... we depend on it to continue our tests efficiently on Eco platform neocompiler.io.

localhuman:
Looks good, though there is a `neo-boa` test that is failing because of this change.  We'll need to update that before we can merge to master.

igormcoelho:
Ok, I'm trying to fix this error, but I don't find boa tests anymore. They were usually in neo/ folder, but are they only in boa project now?
If that's the situation, it would be nice to have all tests in the same project, because I'll probably need to fix the test, because the test is probably wrong.

igormcoelho:
I'm taking a look now at the tests, it's very strange that this is failing now, because it seems the input on the test has 20 bytes, so that wouldn't matter regarding the change I made. False should continue False and True should continue True. Maybe it's something else, I don't know.

localhuman:
I updated the test in neo-boa and published a new release.  The test server runs the neo-boa tests from neo-python when there is a push, that's why they run but you wouldn't see them in this repository.  Anyways this should be good now!

igormcoelho:
Thanks a lot for the attention, @localhuman , @metachris and @canesin . Congratulations for the good work!

PR #444, opened 2018-05-31
return type must be bigint

igormcoelho:

According to a fruitful discussion in neo project ( https://github.com/neo-project/neo/pull/253 ), we concluded that the return type must be coded as a BigInteger, not a simple hex... For most types it doesn't matter (eg. boolean is 01, and BigInt(1).toHex is also 01), but for Void, the neo machine expects to receive \xff00 which is the real BigInt(255), not \xff, which is BigInt(-1)... This will also cause problems with InteropInterface type, that is passing now as \xf0, but it's supposed to be the BigInt representation of it.

Struggling for many months, and this last discussion in NEO: https://github.com/neo-project/neo/pull/253

This works already on Eco platform (neocompiler.io), when building transactions correctly.
On neo-python, I haven't tested well yet... this may have some impacts in FunctionCode, and I still haven't time to view all of it yet. But still, this needs to be fixed as soon as possible, in order to avoid other strange errors (such as https://github.com/CityOfZion/neo-python/pull/422 and https://github.com/CityOfZion/neo-python/issues/217)

Not yet, we have to see the impacts.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/LoadSmartContract.py
+++ b/neo/Prompt/Commands/LoadSmartContract.py
@@ -119,7 +119,7 @@ def GatherLoadedContractParams(args, script):
     if type(params) is str:
         params = params.encode('utf-8')
 
-    return_type = bytearray(binascii.unhexlify(str(args[1]).encode('utf-8')))
+    return_type = BigInteger(int(str(args[1]).encode('utf-8')),16)
 
     needs_storage = bool(parse_param(args[2]))
     needs_dynamic_invoke = bool(parse_param(args[3]))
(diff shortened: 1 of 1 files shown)


igormcoelho:
please dont take a look at this yet.i will improve and try to put some tests, when I have time

canesin:
@igormcoelho I am reopening this, you can commit any test fix directly to your branch and it will appear in the PR automatically. Thanks for the catch.

igormcoelho:
Guys, `make test` is still failing (on 6 cases) and I don't have any idea on the outcome (they are complaining on some `TypeError: 'module' object is not callable`, but that should be some random message, I don't know the exact failures). Sorry, but I don't know if I will manage to finish all these tests, my time is quite constrained and I was just trying to help. If someone else manages to finish this, I believe it's a good direction (better a breaking code than a code that compiles but has misleading behavior).

localhuman:
I can take a look

localhuman:
Closing this in favor of #448

PR #447, opened 2018-05-31
Feature Invoke Attrs and multi-sign

localhuman:
- Adds ability to attach tx attributes to SC invokes
- Adds the abililty to attach an arbitrary number of signing 'owners' to a transaction
- Other various fixes

- Add optional `invoke_attrs` and `owners` parameters to the invoke routines

- tests

- no

- [x] Did you add any tests?


neo/Core/Block.py
```diff
@@ -194,7 +194,7 @@ def Equals(self, other):
         return self.Hash == other.Hash
 
     @staticmethod
-    def FromTrimmedData(byts, index=None, transaction_method=None):
+    def FromTrimmedData(byts):
```
ixje: please also remove the parameters from the docsstrings

neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -401,6 +398,7 @@ def OnSaveTransaction(self, tx, added, changed, deleted):
 
     def OnCoinsChanged(self, added, changed, deleted):
 
+        some = False
```
ixje: delete left over/unused var

neo/Core/FunctionCode.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, script=None, param_list=None, return_type=None, contract_prop
 
         if return_type and not isinstance(return_type, int) and len(return_type) > 1:
             return_type = return_type[0:1]
-        elif isinstance(return_type, int) and not 0 < return_type <= 255:
+        elif isinstance(return_type, int) and return_type > 255:
```
ixje: just something I observed; `FunctionCode` is often initialized as `FunctionCode()` throughout the code base. This creates a return_type of `None`. There is no check that makes sure it becomes a valid return type. Perhaps we should make the default value to be `0xff` (void)?

neo/Wallets/NEP5Token.py
```diff
@@ -244,7 +244,7 @@ def Approve(self, wallet, owner_addr, requestor_addr, amount):
 
         return tx, fee, results
 
-    def Mint(self, wallet, mint_to_addr, attachment_args):
+    def Mint(self, wallet, mint_to_addr, attachment_args, invoke_attrs=None):
```
ixje: please update the docstring for the new argument

localhuman:
Thanks for the feedback!  Will do that and also add some tests

PR #448, opened 2018-06-01
ReturnType is now written by ScriptBuilder as a BigInteger

localhuman:
- using this in favor of https://github.com/CityOfZion/neo-python/pull/444

- `generate_deploy_script` now expects `BigInteger` for the `ReturnType` paramater

- tests

- no

- [ ] Did you add any tests?


igormcoelho:
Congratulations @localhuman , this bug was much harder for me to fix correctly, this way was much better well done. I'm very happy to see it fixed, this has bothered me for "centuries" (it felt like :D), now the dragon has been killed ;)

PR #449, opened 2018-06-02
Bugfix neopython prompt return value

igormcoelho:

This commit fixes a small issue, regading invocation errors using prompt.py (parse error on return value type), and should contribute to PR https://github.com/CityOfZion/neo-python/pull/448. 

`import contract test.avm "" 02 True False`

Conversion error at: `File "/neo-python/neo/Core/FunctionCode.py", line 19, in ReturnTypeBigInteger    return BigInteger(self.ReturnType)`

I tested locally and realized prompt.py was failing (with conversion errors b'\xff' to BigInteger), so I converted data to string b'ff', then to int, then to BigInteger (if convert b'ff' to BigInteger it will become -1, not 255 as intended). The same applies to other types (int 0x02 was also failing with conversion error).

I tested locally and now all invocations work (for Void ff, and other types, boolean, int...). This version is running online at Eco platform neocompiler.io.

It contributes to the PR https://github.com/CityOfZion/neo-python/pull/448

- [ ] Did you add any tests?

Thanks for the attention!

### DIFF
--- a/neo/Core/FunctionCode.py
+++ b/neo/Core/FunctionCode.py
@@ -1,6 +1,7 @@
 from neocore.IO.Mixins import SerializableMixin
 from neocore.Cryptography.Crypto import Crypto
 from neocore.BigInteger import BigInteger
+from binascii import hexlify
 
 
 class FunctionCode(SerializableMixin):
@@ -16,7 +17,7 @@ class FunctionCode(SerializableMixin):
 
     @property
     def ReturnTypeBigInteger(self):
-        return BigInteger(self.ReturnType)
+        return BigInteger(int(hexlify(self.ReturnType), 16))
 
     @property
     def HasStorage(self):


metachris:
@localhuman not sure about the implications of this change, please can you take a quick look?

There's some issue with the versions, why dies the version number change in this PR?  Seems there is one commit too many, can you please make sure you have the development branch merged in again @igormcoelho

igormcoelho:
In fact I just want to send commit https://github.com/CityOfZion/neo-python/pull/449/commits/432e460d54bdd4c6f4e9c200dc361615d1d22f84
Somehow, my branch is also pushing this commit https://github.com/CityOfZion/neo-python/pull/449/commits/47d23c3357462bf505b3b55f1fbdc635cf6205b3 which is not in my branch: https://github.com/igormcoelho/neo-python/tree/bugfix_neopython_prompt_return_value

Can you remove that @metachris ?

igormcoelho:
Ok, I managed to destroy that commit from branch :)

igormcoelho:
Ok, I'll test.

igormcoelho:
Oh, that seems nice @localhuman , congratulations!
However.... (there's always some hidden stuff hahaha) do you know what is this `Results ['2222', '3']` in python output?
My contract was just a simple `return 3`, but I noticed that on Csharp and Python it also returns this strange value. I'm just saying, because even when contract is `void`, it is doing `Results ['2222']`.

It came to a point that I want to understand it 100%, to avoid future problems hahah better to beat them now :)

Issue #441, opened 2018-05-30
Feature request: Save raw blocks to allow offline resync


ThomasLobker:
### Current behavior

- Whenever a bug is fixed or something that would require a resync to fix the current data in the backend, `neo-python` needs to redownload all blocks.

### Expected behavior

- Because the already fetched raw blocks should never change, they can be safely stored on disk and used again. Whenever a users needs to or wants to resync, `neo-python` could use the cached blocks instead of downloading them from other nodes.

`neo-cli` has something like this where you can safe the blocks as an `acc` file.

ixje:
Just to make sure you're aware; neo-python has a feature to [bootstrap](https://github.com/CityOfZion/neo-python#bootstrapping-the-blockchain). This is basically a zip'ed copy of the blockchain + notification db so you don't even have to process the blocks anymore. It's also updated every 1-2 weeks.

ThomasLobker:
> This is basically a zip'ed copy of the blockchain + notification db so you don't even have to process the blocks anymore.

Yes I'm aware and this is actually exactly my point. I've had multiple occasions where I found that the processed data was faulty, so processing the raw blocks again with an updated version of `neo-python` would be necessary. I personally think it's best practice to let node operators maintain their own backups and snapshots and not rely on centralized pre-processed bootstrap files. No offence, but `neo-python` has currently no way to verify the integrity of a bootstrap file.

I would be okay too with a feature where `neo-python` can verfiy the integrity of the bootstrap file in the background. This would probably be a much bigger change though.

ixje:
Good point on the integrity (and obviously no offence taken) This issue of course also exists with the `.acc` file provided by NEO. It's all trust based. 

I've thought a bit about other ways of bootstrapping in the past where consensus nodes sign off on certain data to build trust without having to process everything yourself. That idea originated when thinking about pruning large chains (something we'll likely want to support in a distant future) but the solution can potentially overlap with your request. If you'd like we can brainstorm a bit on Discord on a bigger scale than just for `neo-python`?

dethos:
This issue seem to have been solved with the new `export`/`import` features. Perhaps it is time to close it.

PR #451, opened 2018-06-03
Feature import export

localhuman:
- features for #441 

- added `np-import` and `np-export` commands 

- manual tests

usage: 
  - `np-export -m --output chain.pcc`
  - `np-import -m --input chain.pcc`

 - also use flag `-l` for import to display smart contract events

- [ ] Did you add any tests?

### DIFF
--- a/neo/bin/export.py
+++ b/neo/bin/export.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+
+from neo.Settings import settings
+from neo.Implementations.Blockchains.LevelDB.LevelDBBlockchain import LevelDBBlockchain
+from neo.Core.Blockchain import Blockchain
+import argparse
+from tqdm import trange
+import binascii
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-m", "--mainnet", action="store_true", default=False,
+                        help="use MainNet instead of the default TestNet")
+    parser.add_argument("-c", "--config", action="store", help="Use a specific config file")
+
+    # Where to store stuff
+    parser.add_argument("--datadir", action="store",
+                        help="Absolute path to use for database directories")
+
+    parser.add_argument("-o", "--output", help="Where to save output file")
+
+    parser.add_argument("-t", "--totalblocks", help="Total blocks to export", type=int)
+
+    args = parser.parse_args()
+
+    if args.mainnet and args.config:
+        print("Cannot use both --config and --mainnet parameters, please use only one.")
+        exit(1)
+
+    # Setting the datadir must come before setting the network, else the wrong path is checked at net setup.
+    if args.datadir:
+        settings.set_data_dir(args.datadir)
+
+    # Setup depending on command line arguments. By default, the testnet settings are already loaded.
+    if args.config:
+        settings.setup(args.config)
+    elif args.mainnet:
+        settings.setup_mainnet()
+
+    file_path = None
+
+    if not args.output:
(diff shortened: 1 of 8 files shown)


neo/bin/export.py
```diff
@@ -0,0 +1,85 @@
...
+        raise Exception("Please specify an output path")
+    file_path = args.output
+
+    # Instantiate the blockchain and subscribe to notifications
+    blockchain = LevelDBBlockchain(settings.chain_leveldb_path)
+    Blockchain.RegisterBlockchain(blockchain)
+
+    chain = Blockchain.Default()
+
+    file = open(file_path, 'wb')
```
ixje: I'd like to suggest using `with open(path) as f:` because the current implementation will not close the file if an exception occurs.

neo/bin/import.py
```diff
@@ -0,0 +1,106 @@
...
+    file_path = None
+
+    if args.logevents:
+        settings.log_smart_contract_events = True
+
+    if not args.input:
+        raise Exception("Please specify an input path")
+    file_path = args.input
+
+    file = open(file_path, 'rb')
```
ixje: same comment as above (`with open() as f`)

localhuman:
Thanks for the feedback!

ixje:
@localhuman not sure if you want to update it anymore, but the `file.close()` statements are not necessary when using `with open`. That's done implicitly when leaving the `with` block

neo/bin/export.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # Setting the datadir must come before setting the network, else the wrong path is checked at net setup.
+    if args.datadir:
+        settings.set_data_dir(args.datadir)
+
+    # Setup depending on command line arguments. By default, the testnet settings are already loaded.
+    if args.config:
+        settings.setup(args.config)
+    elif args.mainnet:
+        settings.setup_mainnet()
```
metachris: Settings are not using testnet by default anymore, so I'd recommend adding `else: settings.setup_testnet()`

neo/bin/export.py
```diff
@@ -0,0 +1,81 @@
...
+
+    file_path = args.output
+
+    # Instantiate the blockchain and subscribe to notifications
+    blockchain = LevelDBBlockchain(settings.chain_leveldb_path)
+    Blockchain.RegisterBlockchain(blockchain)
+
+    chain = Blockchain.Default()
+
+    with open(file_path, 'wb') as file:
```
metachris: `file` should be avoided as variable name since it already exists, better perhaps `file_out`

neo/bin/import.py
```diff
@@ -0,0 +1,104 @@
...
+    # Setting the datadir must come before setting the network, else the wrong path is checked at net setup.
+    if args.datadir:
+        settings.set_data_dir(args.datadir)
+
+    # Setup depending on command line arguments. By default, the testnet settings are already loaded.
+    if args.config:
+        settings.setup(args.config)
+    elif args.mainnet:
+        settings.setup_mainnet()
+
```
metachris: same as in `export`, should use else:setup_testnet

neo/bin/import.py
```diff
@@ -0,0 +1,104 @@
...
+    if args.logevents:
+        settings.log_smart_contract_events = True
+
+    if not args.input:
+        raise Exception("Please specify an input path")
+    file_path = args.input
+
+    with open(file_path, 'rb') as file:
```
metachris: same as above, perhaps change name `file`

PR #453, opened 2018-06-03
prompt store history and log in custom datadir

metachris:

If user uses a custom data-dir with `--datadir`, np-prompt should store log and history there. Before it would just use `~/.neopython/`.

Removing temporary user-data paths in settings, and updating prompt.py slightly to use the correct base directory.

Did a small refactoring to Settings.py, to not have the temporary user paths anymore.

Note: I left the explicit setting of testnet in the examples because it illustrates what you should do where, helps the examples to make this explicit.

Manual tests

no (TestNet is still used by default)

- [ ] Did you add any tests?


localhuman:
Looks good to me! Would be interested how others feel!

brianlenz:
I like it 👍 Should the changelog mention that if you use a custom `--datadir` that your `np-prompt` history will reset?

metachris:
Good idea @brianlenz. Updated.

metachris:
Gonna merge this if no more comments or objections

Issue #428, opened 2018-05-21
Question: JSON-RPC and wallet


kornrunner:
Hi,

This actually isn't an issue, but rather a question on usage. I've tried to find it in docs and existing issues, but I've failed, so I apologize upfront if I've missed something.

I've initially used `neo-cli` and it's RPC API. I've got an application written in PHP that's using JSON RPC to communicate with neo node and do stuff. However, recently I've wanted to do something with smart contracts, and neo-cli isn't supporting them in JSON RPC API yet, so I've stumbled upon your project.

What I'm looking for is: how can I open my wallet when running `np-api-server`? I've managed to open the wallet by using `np-prompt`, but since I need RPC endpoint open as well for contracts, and I need my funds available for smart contrancts - how can I acheive this?

Thank you!

dethos:
I'm not sure this is possible at the moment. According to this issue, https://github.com/CityOfZion/neo-python/issues/189, some of the RPC methods that require the open wallet aren't implemented yet.

metachris:
Which RPC methods in particular would you like to use?

kornrunner:
I was hoping for `sendtoaddress` which I reckon I'd need unlocked wallet.

dethos:
I'm currently working on a PR that might solve this issue and allow us to add the missing methods, that depend on an open wallet. 

I should have the first draft ready in the next few days.

metachris:
@kornrunner is this helping you, and do you have any more reason for this issue to stay open?

kornrunner:
It looks like this is it. I'll still have to check it out, but I'll close this one and open a new one if anything is out of order.

Thank you very much!

PR #459, opened 2018-06-06
Add --wallet argument to the api server

dethos:
According to issue https://github.com/CityOfZion/neo-python/issues/189 there are still a few methods missing in the RPC API of `neo-python`. Most of them require an open wallet, a functionality that was missing in the current implementation. 

#428 was later created asking for one of those methods.

Added ``--wallet`` flag to the ``np-api-server`` command. The server can now open a wallet and use it. As a demonstration I also implemented the `listaddress` method, which uses the wallet.

Using a manual testing process plus adding a couple of automated tests.

Not really. Only that since a wallet can be opened, users should be aware that they should protect the RPC endpoint when used with the `--wallet` flag, in order to avoid unauthorized uses of the wallet.

- [x] Did you add any tests?

### DIFF
--- a/neo/api/JSONRPC/JsonRpcApi.py
+++ b/neo/api/JSONRPC/JsonRpcApi.py
@@ -341,18 +341,13 @@ class JsonRpcApi:
         return result
 
     def list_address(self):
-        """Get information about all the addresses present on the open wallet
-
-        The current implementation still does not fully comply with the spec,
-        since it seems there isn't a straight forward way of getting "haskey"
-        information.
-        """
+        """Get information about all the addresses present on the open wallet"""
         result = []
         for addrStr in self.wallet.Addresses:
             addr = self.wallet.GetAddress(addrStr)
             result.append({
                 "address": addrStr,
-                # "haskey": True,
+                "haskey": not addr.IsWatchOnly,
                 "label": None,
                 "watchonly": addr.IsWatchOnly,
             })
--- a/neo/api/JSONRPC/JsonRpcApi.py
+++ b/neo/api/JSONRPC/JsonRpcApi.py
@@ -245,6 +245,12 @@ class JsonRpcApi:
         elif method == "getpeers":
             return self.get_peers()
 
+        elif method == "listaddress":
+            if self.wallet:
+                return self.list_address()
+            else:
+                raise JsonRpcError(-400, "Access denied.")
+
         raise JsonRpcError.methodNotFound()
 
     def get_custom_error_payload(self, request_id, code, message):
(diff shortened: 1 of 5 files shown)


neo/bin/api_server.py
```diff
@@ -181,6 +189,20 @@ def main():
         else:
             print("Logging to stdout and stderr")
 
+    if args.wallet:
+        if not os.path.exists(args.wallet):
+            print("Wallet file not found")
+            return
+        passwd = prompt("[password]> ", is_password=True)
```
metachris: We could allow an environment variable for the password, so this could run more easily supervised (eg. systemd). For instance `NEO_PYTHON_JSONRPC_WALLET_PASSWORD`, and if set it would skip the manual prompt. Just an idea, what do you think?

metachris:
Looks good on first glance. Added one comment/question about whether we should allow to set the pwd with an env-var for easier running on a server.

PS: Changelog has a conflict, please merge the development branch into your branch again :)

neo/bin/api_server.py
dethos: Yes definitely. :+1:

metachris:
Last question -- do we need the `haskey` part of the json-rpc response for `listaddress`?

dethos:
@metachris not sure, but if we want to maintain the same interface as `neo-cli` nodes, users might be expecting item in the response.

metachris:
Going to merge this if no further comments @localhuman @ixje @brianlenz

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -332,3 +339,21 @@ def get_peers(self):
...
+        The current implementation still does not fully comply with the spec,
+        since it seems there isn't a straight forward way of getting "haskey"
+        information.
+        """
+        result = []
+        for addrStr in self.wallet.Addresses:
+            addr = self.wallet.GetAddress(addrStr)
+            result.append({
+                "address": addrStr,
+                # "haskey": True,
```
brianlenz: Could this just be:

```
"haskey": not addr.IsWatchOnly
```

?

brianlenz:
Looks good to me. I just had an idea for `haskey` that I commented on.

neo/api/JSONRPC/JsonRpcApi.py
dethos: @localhuman @metachris what do you think? 
Just need to make sure that information will be correct.

metachris: I think its a good idea

dethos:
Done. Applied the suggested changes.

PR #463, opened 2018-06-08
speed up import by re-using stream and block objects

ixje:
the new blockchain `import` was apparently slow. The changes introduced give a 2x improvement

re-use objects instead of re-creating them for each block to process.

manual testing

no

- [ ] Did you add any tests?
can I skip for this one? pretty please 🙏


localhuman:
👍

PR #464, opened 2018-06-11
Add RPC method: "getnewaddress"

dethos:
According to the issue https://github.com/CityOfZion/neo-python/issues/189 there are still a few methods missing in the API. This pull request implements `getnewaddress`.

When calling the method on a node with a open wallet, a new address is created.

Manually and with automated tests.
No

- [x] Did you add any tests?

### DIFF
--- a/neo/api/JSONRPC/test_json_rpc_api.py
+++ b/neo/api/JSONRPC/test_json_rpc_api.py
@@ -4,7 +4,6 @@ Run only thse tests:
     $ python -m unittest neo.api.JSONRPC.test_json_rpc_api
 """
 import json
-import pprint
 import binascii
 import os
 from klein.test.test_resource import requestMock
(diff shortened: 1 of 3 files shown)


neo/api/JSONRPC/test_json_rpc_api.py
```diff
@@ -516,3 +514,34 @@ def test_listaddress_with_wallet(self):
...
+        mock_req = mock_request(json.dumps(req).encode("utf-8"))
+        res = json.loads(self.app.home(mock_req))
+
+        error = res.get('error', {})
+
+        self.assertEqual(error.get('code', None), -400)
+        self.assertEqual(error.get('message', None), "Access denied.")
+
+    def test_getnewaddress_with_wallet(self):
+        test_wallet_path = "./listaddress.db3"
```
metachris: i think a slightly better approach would be using the Tempfile module to create a temp wallet file in a standard temp directory: https://docs.python.org/3/library/tempfile.html

metachris:
Looks good to me! 👍 Just one minor comment

dethos:
@metachris done. 
Note: the wallet is created on a temporary directory. Could not use the temporary file feature you suggest, since the test is using the `create` wallet method and and the file/path must not exist for it to work.

localhuman:
Great work!

PR #465, opened 2018-06-12
Add "getbalance" to the RPC api

dethos:
This PR adds the `getbalance` method to the JSON RPC API. (It is still missing according to issue #189)
I added the method, following the specification present on the [NEO docs](http://docs.neo.org/en-us/node/cli/2.7.4/api/getbalance.html).
Manually and with automated tests.
No

- [x] Did you add any tests?


localhuman:
Looks good! Needs some merge conflicts to be resolved

PR #473, opened 2018-06-19
return value of address in ImportMultiSigContractAddr()

HandsomeJeff:
Currently we are unable to directly get the address of a multisig contract. This is for ease-of-use for creating custom python programs. 

I added `return address`.

It's a simple, quality of life improvement

Just two words

- [x ] Did you add any tests?
- [x ] Did you run `make lint`?
- [x ] Did you run `make test`?
- [x ] Are you making a PR to a feature branch or development rather than master?

### DIFF
--- a/neo/Prompt/Commands/LoadSmartContract.py
+++ b/neo/Prompt/Commands/LoadSmartContract.py
@@ -235,7 +235,6 @@ def ImportMultiSigContractAddr(wallet, args):
         wallet.AddContract(verification_contract)
 
         print("Added multi-sig contract address %s to wallet" % address)
-        
         return address
 
     return 'Hello'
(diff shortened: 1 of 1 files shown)


neo/Prompt/Commands/LoadSmartContract.py
```diff
@@ -235,5 +235,7 @@ def ImportMultiSigContractAddr(wallet, args):
         wallet.AddContract(verification_contract)
 
         print("Added multi-sig contract address %s to wallet" % address)
+        
+        return address
 
     return 'Hello'
```
metachris: Might this `return 'Hello'` not be a problem for your use-case?

metachris:
Looks good. Please also add an entry to the changelog file

neo/Prompt/Commands/LoadSmartContract.py
HandsomeJeff: In my use-case, I'll be interacting with the newly created multisig contract address. So having ImportMultiSigContractAddr() return this address would be helpful.

localhuman:
looks like a codestyle issue is failing the build

HandsomeJeff:
Hi, is there anything I can do on my end?

metachris:
Yes, please run `make lint` and fix the issue.

Also please add a CHANGELOG entry to this PR.

localhuman:
Thanks!

PR #477, opened 2018-06-26
Dependency update: prompt-toolkit

metachris:

Updated the code to work with the new version of prompt-toolkit, which in v2.x had a few breaking changes, in particular around the Style and the prompt function signature: https://github.com/jonathanslenders/python-prompt-toolkit/blob/master/CHANGELOG. 

See also:
* https://github.com/jonathanslenders/python-prompt-toolkit/blob/deb7488b7b24814ae12518639a2fbe0b67c43ec3/examples/print-text/print-formatted-text.py
* https://github.com/jonathanslenders/python-prompt-toolkit/blob/deb7488b7b24814ae12518639a2fbe0b67c43ec3/examples/prompts/bottom-toolbar.py

Updating the code based on the prompt-toolkit examples.

The only thing I could not migrate was the bottom toolbar styling. It seems the new version doesn't allow foreground colors in the bottom toolbar anymore: https://github.com/CityOfZion/neo-python/blob/deps-prompt/neo/bin/prompt.py#L171 

It looks like this now:

screen shot 2018-06-26 at 12 16 23

Manual tests

No, except the changed bottom toolbar styling

- [ ] Did you add any tests?


localhuman:
Looks good! I'm ok with the different bottom toolbar styling-

localhuman:
Actually, after playing around with this for a bit I'm noticing some odd things such as: ` <prompt_toolkit.styles.style.Style object at 0x10fc36f60>` being printed after any command using `FormattedText`.  Will revert for now

PR #483, opened 2018-06-28
Feature prompt iargs

localhuman:

- Adds option to enter arguments for smart contract in an 'interactive' mode, which allows for much better parsing of input, activated by passing the `--i` flag when invoking.   Example:

```
neo> build mycontract.py test 0710 10 True False --i
[I 180627 19:04:00 BuildNRun:51] Saved output to ../nex-neo-vault/vault.avm 
[Param 0] String input: getBalances
[Param 1] Array input: ['AWeZnH735EavQJKbJPC5F8fxutBnJFhukW', [123, 'abcd']]
[I 180627 19:04:08 Invoke:548] Used 0.349 Gas 

Calling mycontract.py with arguments ['--i'] 
Test deploy invoke successful
Used total of 378 operations 
Result [{'type': 'Array', 'value': [{'type': 'Integer', 'value': '3'}, {'type': 'Integer', 'value': '3'}]}] 
Invoke TX gas cost: 0.0001 
-------------------------------------------------------------

neo> 
```

- Adds ability to *not* parse address strings such as AeV59NyZtgj5AMQ7vY6yhr2MRvcfFeLWSb when inputting to smart contract by passing the `--no-parse` flag

- Changes the structure of items dispatched in SmartContractEvents to use the `ContractParameter` interface for better type inference and variable usage.

- removes the `stack_item_to_py` method which was duplicating efforts of the existing standardized `ContractParameter` interface.

- added the `--i` interactive mode for invoking smart contracts.
- removed `stack_item_to_py` method usage with notify and log events.

- manual and automated tests

- Due to changes in the structure of `NotifyEvent`, this PR will not pass tests that are run from the `boa_test` module at the moment unless installing neo-boa from a branch within the `travis` config.  Once this PR is merged to development, `neo-boa` tests can be updated to use the new `NotifyEvent` structure and deployed with a new version.

- Any apps that use or subscribe to notify events will need to be updated.

- [x] Did you add any tests?


neo/Prompt/Utils.py
```diff
@@ -130,6 +133,13 @@ def get_from_addr(params):
     return params, from_addr
 
 
+def get_parse_addresses(params):
+    if '--no-parse' in params:
```
ixje: just a general thought on the name; should we have these kind of flags be self explanatory? e.g. '--attach-neo' is obvious what it does. '--no-addr-parse' or '--no-addr-convert' might be easier to understand without requiring long help instructions.

ixje:
1) The `build ... args` console feedback looks good now. 
2) I also like the idea that when input is wrong, it asks for it again. The only problem is that you can't exit that mode unless you provide a valid input. ctrl+c/z either kills the console or has no effect. This is especially true with the above example of a public key argument.
3) I just noticed that for the `array` type you can actually provide non-arrays (e.g. `123`). I think it should enforce it being an array instance or code can get executed that tries to access an index on a non-array input.

localhuman:
Thanks for the great feedback!

PR #486, opened 2018-06-29
Fix privnet check in settings

metachris:

Fixed the privnet check in settings -- if no privnet is running, now gives proper error message.

n/a

manual test

no

- [ ] Did you add any tests?


localhuman:
👍

Issue #445, opened 2018-05-31
Multisig Fails With NEP-5 Tokens


DaveOnBlocks:
### Current behavior

Sending a NEP-5 token from a multi-sig address results in a failure message e.g.:
`
send TST AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y 1 --from-addr=APWjDP3SgSFGrjh6RwYYh4TJaJbDrcjg65
`
Results in: 
```
Used 0.379 Gas
could not transfer tokens
```

### Expected behavior

-A transaction to sign should be generated

### How to reproduce

-Create a multisig address
-Transfer tokens to address
-Try to move tokens out of address 

### Your environment

Let us know in what environment you're running into the issue:

- OS: docker privnet (Linux 856da81ed65a 4.9.87-linuxkit-aufs)
- neo-python version: v0.6.9
- Python version: 3.6.3

PR #491, opened 2018-07-01
Fix sending NEP5 tokens from multisig address

ixje:
Fix https://github.com/CityOfZion/neo-python/issues/445

transaction attributes were parsed but not added to Invocation transactions. 

Manual testing on privatenet

see own review

- [ ] Did you add any tests?

### DIFF
--- a/neo/Core/TX/Transaction.py
+++ b/neo/Core/TX/Transaction.py
@@ -712,10 +712,7 @@ class Transaction(InventoryMixin):
                 if type(attr.Data) is UInt160:
                     hashes.add(attr.Data)
                 else:
-                    if len(attr.Data) == 40:
-                        hashes.add(UInt160(data=binascii.unhexlify(attr.Data)))
-                    else:
-                        hashes.add(UInt160(data=attr.Data))
+                    hashes.add(UInt160(data=attr.Data))
 
         for key, group in groupby(self.outputs, lambda p: p.AssetId):
             asset = GetBlockchain().GetAssetState(key.ToBytes())
--- a/neo/Core/TX/Transaction.py
+++ b/neo/Core/TX/Transaction.py
@@ -712,7 +712,10 @@ class Transaction(InventoryMixin):
                 if type(attr.Data) is UInt160:
                     hashes.add(attr.Data)
                 else:
-                    hashes.add(UInt160(data=attr.Data))
+                    if len(attr.Data) == 40:
+                        hashes.add(UInt160(data=binascii.unhexlify(attr.Data)))
+                    else:
+                        hashes.add(UInt160(data=attr.Data))
 
         for key, group in groupby(self.outputs, lambda p: p.AssetId):
             asset = GetBlockchain().GetAssetState(key.ToBytes())
--- a/neo/Prompt/Commands/Invoke.py
+++ b/neo/Prompt/Commands/Invoke.py
@@ -336,19 +336,18 @@ def test_invoke(script, wallet, outputs, withdrawal_tx=None,
             wallet_tx.Gas = tx_gas
             # reset the wallet outputs
             wallet_tx.outputs = outputs
-            wallet_tx.Attributes = []
+            wallet_tx.Attributes = [] if invoke_attrs is None else invoke_attrs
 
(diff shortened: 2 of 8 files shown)


neo/Core/TX/Transaction.py
```diff
@@ -712,7 +712,10 @@ def GetScriptHashesForVerifying(self):
                 if type(attr.Data) is UInt160:
                     hashes.add(attr.Data)
                 else:
-                    hashes.add(UInt160(data=attr.Data))
+                    if len(attr.Data) == 40:
```
ixje: previously this branch expected a `bytes` object. Imagine we add a `--tx-attr` of type `script` (0x20), then we would have to add something like: 
> --tx-attr={"usage":32,"data": b'#\xba\'\x03\xc52c\xe8\xd6\xe5"\xdc2 39\xdc\xd8\xee\xe9'}

this will fail to parse correctly because there's a space in the above data value, even though it translates to a valid script_hash. This little branch here allows to provide a data input on the console as follows:
> --tx-attr={"usage":32,"data": b'23ba2703c53263e8d6e522dc32203339dcd8eee9'}

neo/Prompt/Commands/Invoke.py
```diff
@@ -336,19 +336,18 @@ def test_invoke(script, wallet, outputs, withdrawal_tx=None,
             wallet_tx.Gas = tx_gas
             # reset the wallet outputs
             wallet_tx.outputs = outputs
-            wallet_tx.Attributes = []
+            wallet_tx.Attributes = [] if invoke_attrs is None else invoke_attrs
```
ixje: Currently I'm only re-applying the attributes that were passed to the function initially. I'm a little bit in doubt whether we should also restore one of the following attributes.

this: https://github.com/ixje/neo-python/blob/99adf951a4798f344fef02acd884f6a7ee0edac8/neo/Prompt/Commands/Invoke.py#L268
or this one
https://github.com/ixje/neo-python/blob/99adf951a4798f344fef02acd884f6a7ee0edac8/neo/Prompt/Commands/Invoke.py#L284

feedback appreciated why we should or should not.

neo/Core/TX/Transaction.py
ixje: I looked into this suggestion. Assuming you were referring to this part
https://github.com/CityOfZion/neo-python/blob/f570ce99302033b6276e338ee5043449b118ba70/neo/Prompt/Utils.py#L133-L139

or this part a little bit deeper
https://github.com/CityOfZion/neo-python/blob/f570ce99302033b6276e338ee5043449b118ba70/neo/Prompt/Utils.py#L162-L168
then we would pretty much be duplicating this loop
https://github.com/CityOfZion/neo-python/blob/f570ce99302033b6276e338ee5043449b118ba70/neo/Core/TX/Transaction.py#L710-L715

as we would still have to determine that the attribute usage is of type `Script` and then that it's not already in the right format. I think the current solution makes most sense, but I'm still open for an example what you might have in mind if you think otherwise.

ixje:
@localhuman I believe it should be done now.

neo/Core/TX/Transaction.py
localhuman: I was thinking of something like this:
```    
        datum = obj['data']
        if type(datum) is str:
            datum = datum.encode('utf-8')
        if len(datum) == 40:
            datum = binascii.unhexlify(datum)
        usage = obj['usage']
        return TransactionAttribute(usage=usage, data=datum)
```

ixje: Aha. I understand your view now. I do still think that would be suboptimal because any TX attribute of length 40 will be unhexlified. This means also `Description`, `Remark` types and others that happen to be exactly 40 characters. The current suggested solution is 100% just for script hashes, and that's all I (we?) can foresee it applies for now.

localhuman: How about something like this: 
```        datum = obj['data']
        if type(datum) is str:
            datum = datum.encode('utf-8')
        usage = obj['usage']
        if usage == script and len(datum) == 40:
            datum = binascii.unhexlify(datum)

        return TransactionAttribute(usage=usage, data=datum)```

PR #496, opened 2018-07-05
Update README.rst

jseagrave21:

Updates which Ubuntu Windows users are directed to install.
Simplifies the instructions for creating and activating a venv.
Gives instructions for first time users of neo-python.

Gives the address for Ubuntu 18.04.
Directs users to create an alias for python and pip.
Directs users to uninstall pycryto and pycryptodome and install pycryptodome 3.6.1 specifically because simply using pip install pycryptodome will install pycryptodome 3.6.3 which is incompatible with neo-python.

I ran it. I was finally able to run neo-python for the first time.

no

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -194,7 +194,7 @@ could lead to version conflicts.
     git clone https://github.com/CityOfZion/neo-python.git
     cd neo-python
     
-    #setup alias for python and pip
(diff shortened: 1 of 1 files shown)


localhuman:
Congrats on your first PR! 🎉

PR #499, opened 2018-07-08
Readme Update 2

jseagrave21:
Adds the instructions to install pycrypto-2.6.1 and fixes spelling errors.

It seems like my update to the readme included some spelling errors. Also, based on other people's comments and conversation in Discord, after uninstalling pycrypto, you should reinstall pycrypto-2.6.1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -231,7 +231,7 @@ could lead to version conflicts.
 
   ::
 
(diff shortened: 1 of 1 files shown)


jseagrave21:
Okay, I updated the command to what I think you are saying. And yes, this version seems to be what neo-python is looking for. See the below excerpt from the python chat.

image

ixje:
ah that makes sense. Looks good to me.

PR #509, opened 2018-07-14
Fix size calculations of Serializable classes

### LINKED ISSUE #427: Fix Size calculations for JSON-RPC
### Current behavior

- What is the problem?

Splitting out from #418.

The JSON returned by RPC calls does not include the `size` property for `Block` and `Transaction`:

https://github.com/CityOfZion/neo-python/blob/master/neo/Core/BlockBase.py#L215
https://github.com/CityOfZion/neo-python/blob/master/neo/Core/TX/Transaction.py#L592

The `Size()` functions don't properly align with the C# implementation:

https://github.com/CityOfZion/neo-python/blob/master/neo/Core/BlockBase.py#L91
vs.
https://github.com/neo-project/neo/blob/master/neo/Core/BlockBase.cs#L69

https://github.com/CityOfZion/neo-python/blob/master/neo/Core/TX/Transaction.py#L354
vs.
https://github.com/neo-project/neo/blob/master/neo/Core/Transaction.cs#L119

The root of the issue is that neo-python is using `sys.getsizeof`, which is not accurate.

See https://github.com/CityOfZion/neo-python/pull/418#issuecomment-389803377 for more details. @ixje's PR #423 will help with size calculations.

### Expected behavior

- What should be happening?

`size` should be calculated properly and included in the JSON.

ixje:
Closes #427 

Re-implement all size calculations 

Some tests coverings a lot of the classes and a whole lot of manual comparing by debugging the C# client vs results of python.

no

- [X] Did you add any tests?


localhuman:
Great work! 💯

PR #514, opened 2018-07-19
add prompt command to split VIN to multiple VOUT

localhuman:

- adds the ability to split one VIN of global assets ( neo / gas ) into an arbitrary amount of them.  Example: `wallet split Ab8RGQEWetkhVqXjPHeGN9LJdbhaFLyUXz neo 0 100` would split a neo VIN containing 1000 neo into 100 pieces of 10.

- added prompt command

- manual testing

- no

- [x] Did you add any tests?


localhuman:
Now with tests!

Issue #492, opened 2018-07-04
node stops syncing, no errors


j4ys0n:
### Current behavior

- What is the problem? Node stops syncing new blocks. If I stop/start the node, it continues syncing, but it will sit at the same block for days if that's not done.
- Include a screenshot if it makes sense.

```
[I 180704 00:36:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:36:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:36:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:37:01 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:37:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:37:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:37:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:38:01 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:38:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:38:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:38:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:39:01 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:39:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:39:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:39:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:40:01 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:40:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:40:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:40:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:41:01 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:41:16 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:41:31 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:41:46 threading:864] [MainNet] Block 2278690 / 2464138
[I 180704 00:42:01 threading:864] [MainNet] Block 2278690 / 2464138
[...]

PR #520, opened 2018-07-22
Improve peer connection maintenance in NodeLeader

hal0x2328:

Even after the recent fixes in the development branch to fix the issues in #492, maintaining connection to the peer pool on a long-term basis is still an issue for neo-python in some cases. Often neo-python gets into a state where the peer count will drop to just a few nodes, and no new peers are added after that. Some of the blame may lie on remote nodes incorrectly reporting dead peers as active, a situation neo-python doesn't seem to currently deal well with.

A multi-pronged approach seems to work to improve overall stability of the peer count
:
1. After the initial protocol handshake with a new peer, immediately request its peer list instead of waiting for the timer

2. Track connection failures in a Twisted callback and maintain a session blacklist of peers that have either refused connections or have timed out during the initial connection attempt, to avoid wasting resources attempting to connect to known-dead peers

3. Only add a new remote address to the ADDRS list after a successful connection has been made

4. When new peers are received from a connected node, check the blacklist before attempting a connection to them

5. Instead of reverting back to the seed nodes in PeerCheckLoop if the amount has fallen below 2 peers, instead attempt to reconnect to all unconnected-but-known-good peers up to CONNECTED_PEER_MAX

6. Increased timeout to 120 in Twisted connectTCP, as it seems to be a collective timeout from the beginning of the connection queue start instead of an individual per-connection timeout, resulting in connections timing out very early when a larger CONNECTED_PEER_MAX causes Twisted to open many connections in sequence

Manual testing/observation of seed list on MainNet/TestNet over 24 hours

No

- [x] Did you add any tests?
- [ ] Did you add an entry to `CHANGELOG.rst`? (seems redundant due to other NodeLeader changes recorded in this cycle)

### DIFF
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -110,7 +110,6 @@ class NeoNode(Protocol):
 
     def connectionLost(self, reason=None):
         """Callback handler from twisted when a connection was lost."""
-        self.endpoint = self.transport.getPeer()
         if self.block_loop:
             self.block_loop.stop()
             self.block_loop = None
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -110,6 +110,7 @@ class NeoNode(Protocol):
 
     def connectionLost(self, reason=None):
         """Callback handler from twisted when a connection was lost."""
+        self.endpoint = self.transport.getPeer()
         if self.block_loop:
             self.block_loop.stop()
             self.block_loop = None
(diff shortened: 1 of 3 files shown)


localhuman:
This is great! Will check it out...

localhuman:
Have been running this for a couple of days with no issues.  Lets test it out further in development!

metachris:
Very nice! thanks

PR #521, opened 2018-07-22
Add Support for Python 3.7

jseagrave21:
This update gives specific instructions for installing and running with Python3.7 with no prior dependencies installed.

Note: I uninstalled Ubuntu and re-installed running Windows 10. Everything runs smoothly.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,13 +113,20 @@ OSX
 Ubuntu/Debian 16.10+
 ^^^^^^^^^^^^^^^^^^^^
 
-Ubuntu starting at 16.10 supports Python 3.6 in the official
-repositories, and you can just install Python 3.6 and all the system
-dependencies like this:
+Ubuntu starting at 16.10 supports Python 3.6+ in the official repositories. We recommend
+using Python 3.7. 
+
+First, ensure Ubuntu is fully up-to-date with this:
 
 ::
 
-    apt-get install python3.6 python3.6-dev python3.6-venv python3-pip libleveldb-dev libssl-dev g++
+   sudo apt-get update && sudo apt-get upgrade
+
+You can install Python 3.7 and all the system dependencies like this:
+
+::
+
+   sudo apt-get install python3.7 python3.7-dev python3.7-venv python3-pip libleveldb-dev libssl-dev g++
 
 Older Ubuntu versions (eg. 16.04)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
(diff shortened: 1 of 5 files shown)


jseagrave21:
This pull is in response to the Python 3.7 issue: https://github.com/CityOfZion/neo-python/issues/518

localhuman:
Going to check the install on osx, otherwise thanks for the contribution!

README.rst
```diff
@@ -113,13 +113,20 @@ OSX
 Ubuntu/Debian 16.10+
 ^^^^^^^^^^^^^^^^^^^^
 
-Ubuntu starting at 16.10 supports Python 3.6 in the official
-repositories, and you can just install Python 3.6 and all the system
-dependencies like this:
+Ubuntu starting at 16.10 supports Python 3.6+ in the official repositories. We recommend
+using Python 3.7.
```
metachris: I'm not sure we want to recommend Python 3.7. Perhaps just remove that sentence.

metachris:
@localhuman - sure that we want to have all documentation use Python 3.7 by default? @jseagrave21  is there a good reason for that? I think supporting 3.7 is great, just not sure we should tell everyone to use 3.7. If we do, we should probably also update the tests to use 3.7.

jseagrave21:
@metachris neo-python seems faster to me running on Python 3.7 (best technical answer I have). Since I was able to install and run neo-python using Python 3.7, I figured the next logical step was to update the readme for newcomers, and to show we support the latest Python. If a recommendation for Python 3.7 is too much, I understand.

If we need to update the tests also, I could possibly start doing that. It seems like something I should be able to figure out.

README.rst
jseagrave21: I added this because the rest of my instructions were for Python 3.7. The whole sentence could be removed.

metachris:
I see! I'm on board with it, just thinking of implications, like if we recommend it that we should perhaps also switch the tests to it?

jseagrave21:
That sounds good to me.

jseagrave21:
I updated travis.yml in a local venv and all tests seem to run properly.
I also updated some other docs to show consistency.
I have not updated documentation for older versions of Ubuntu.

What do you think?

metachris:
Like this Travis doesn't seem to like it: https://travis-ci.org/CityOfZion/neo-python/builds/407480192

Check out this: https://github.com/travis-ci/travis-ci/issues/9831 -- perhaps you need to add another line or two to make it work.

localhuman:
I would like to support 3.7 but don't want to make it the default yet.  I think it would make more sense to add support for 3.7 for a month or so while the community tries it out and if there are no problems we will make it the recommended Python version.

jseagrave21:
Makes sense. I have removed mention of a recommendation but you could also close this PR and I will keep the branch open so the docs are still ready to go if needed. I am enjoying using 3.7 because it seems that tests and operation in general are much faster, but like I said earlier, I created this PR because it is something I can do and seemed logical. On the other hand, it seems like Travis is not happy, and I don't know how to fix that lol.

localhuman:
Sounds good, thanks for the work!

PR #542, opened 2018-08-09
Fix typos

tomascarvalho:
Fixes two typos on print messages

Fixed the typos

N/A

No

- [ ] Did you add any tests?


ixje:
thanks!

PR #544, opened 2018-08-10
update requirements for doc creation

ixje:
the requirements for just creating documentation were heavily outdated. 

created a fresh install/venv, `make doc`'ed + `pip install` until all missing packages were resolved.

`make doc`
no

- [ ] Did you add any tests?


localhuman:
Last time I checked, the Read the Docs server that builds the docs can only be done with Python 3.5, so I know upgrading neo-boa ( which won't work with 3.5 ) will break the docs server build.

ixje:
Oh didn't know that. I only tested locally. Any way to test without pushing here?
I think there's a good chance python 3.6 is now supported, see:
* https://github.com/rtfd/readthedocs.org/pull/3817
* https://github.com/rtfd/readthedocs.org/pull/3814#discussion_r175231510

localhuman:
That would be great if they're supporting 3.6 venv now!  I can switch the build to the development branch to see if it works after merging this-

ixje:
I'll leave it up to you to take further then 👍

Issue #546, opened 2018-08-11
Docker: Using less resources with Alpine Linux image instead of ubuntu


nunojusto:
### Current behavior

- What is the problem?
Why use Ubuntu container if we can use a lighter Linux distro like Alpine

### Expected behavior

- What should be happening?
Confirmed less memory used (see image attached) and also with less services/packages installed on Linux we have less potential vulnerabilities
 (see image attached)
![selection_999 534](https://user-images.githubusercontent.com/21068187/43986186-c638c892-9d05-11e8-9e7a-836f37964d24.png)

###
Conclusion:
I've all done and running with Alpine. Just waiting for a PR approval.

PR #547, opened 2018-08-11
Changed container to Alpine Linux instead of Ubuntu

nunojusto:
#546 

Rebuild current Dockerfile to use Alpine Linux image

Already tested and it's working locally

No


nunojusto:
Hi, metachris.
I know that in CoZ is kind of standard to use ubuntu and i agree with the argument of being more easy to play with package manager. But in the end is just a change aversion cause it's not so different to use apk.
There are advantages to use a lighter distro: Faster, smaller, more secure.
Some of them are already visible and I posted in my screenshot (inside the issue).

Here's a good article advocating for Alpine instead of any other.
https://nickjanetakis.com/blog/the-3-biggest-wins-when-using-alpine-as-a-base-docker-image

i25959341:
This is awesome. I used Kubernetes for containers deployment and using Alpine would make the deployment process easier to build and more secure for containers that use neo-python. 

I think Alpine shouldn't replace the ubuntu images but should be provided as an option for DevOps like us to use, as alpine is great for production. And the Ubuntu version should be used for the development stage.

metachris:
Let's start with this being a second Dockerfile besides the Ubuntu one and see where it is developing. Could you please rename the alpine file to `Dockerfile.alpine` or something similar?

nunojusto:
Sure, I willl do that

metachris:
Oh, and keep the original Dockerfile in there please :)

nunojusto:
I hope, now that's all done

metachris:
It seems there are still some merge conflict artifacts left over: https://github.com/CityOfZion/neo-python/pull/547/files#diff-1ec61223727a10e5e4c74c68336809fc

nunojusto:
Clean

nunojusto:
ok now?

metachris:
Thanks :+1:

Issue #440, opened 2018-05-30
Different result in sys_fee and net_fee between neo-python rpc node and neo-cli node


apisit:
### Current behavior
A transaction that contains sys_fee from neo-python node is different from the one returning from neo-cli node.

- What is the problem?
the `sys_fee` and `net_fee` fields
- Include a screenshot if it makes sense.

here is the one from neo-cli node
<img width="1147" alt="screen shot 2018-05-30 at 9 43 17 am" src="https://user-images.githubusercontent.com/1012916/40692679-2041a242-63ee-11e8-9a4d-2b1ca99a1951.png">

this is from neo-python node
<img width="1156" alt="screen shot 2018-05-30 at 9 43 55 am" src="https://user-images.githubusercontent.com/1012916/40692745-89585dac-63ee-11e8-8021-9fbd11f4f41c.png">

### Expected behavior
returns correct `sys_fee` and `net_fee` 

- What should be happening?
- Include a screenshot if it makes sense.

### How to reproduce
Try fetching this transaction from the neo-python node
```
curl --request POST \
  --url http://node2.ams2.bridgeprotocol.io:10332/ \
  --header 'content-type: application/json' \
  --data '{
  "jsonrpc": "2.0",
  "method": "getrawtransaction",
  "params": ["0xcd0cf97e0679e2e387307c80c077203f35d8f155caca778c720d75a447f2c76a",1],
  "id": 1
}'
```

and then compare it with 
```
curl --request POST \
  --url http://seed2.o3node.org:10332/ \
  --header 'content-type: application/json' \
  --data '{
  "jsonrpc": "2.0",
  "method": "getrawtransaction",
  "params": ["0xcd0cf97e0679e2e387307c80c077203f35d8f155caca778c720d75a447f2c76a",1],
  "id": 1
}'
```

- Please explain the steps to reproduce the problem. If you are having an issue with your machine or build tools, the issue belongs on another repository as that is outside of the scope of this project.

### Your environment
This does not apply

Let us know in what environment you're running into the issue:

- OS: (Windows version/Linux distro/OSX version) 
- neo-python version: (hint: use `np-prompt --version`)
- Python version: (hint: `python -V`)

ixje:
just verified on `/NEO-PYTHON:0.7.3-dev/` and the issue is still there

PR #549, opened 2018-08-11
Development

hal0x2328:
GAS price calculation for syscalls in neo-python is out-of-sync with the core code in neo-project. neo-python was calculating a GAS price of 0.1 GAS for `GetTime` which is not present in the Neo core GAS price calculation. 

Additionally, a GAS fee of 0.1 for the API call `GetTransactionHeight` was added to core but was not present in neo-python. 

Finally, a new prefix `System` was added to the core code in addition to `Neo.` and `Antshares.` which was not handled in neo-python.

**How did you solve this problem?*
I brought the GAS price calculation code in ApplicationEngine.py back into sync with the ApplicationEngine.cs in the Neo core code.

I ran an audit comparing the unchanged neo-python code with neo-cli transaction GAS usage/state results, identifying specific transactions that were inconsistent, the first one of which was 94ac4c0e5d8d9f16246ca5eba63ab5b6b5fa6f38952a8707b9c0de8d772d9a20.

After applying the changes, I ran the audit again, and the GAS calculations for previously inconsistent transactions including the one above are now correct in the modified neo-python instance.

No

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,7 @@ All notable changes to this project are documented in this file.
 ------------------------
 - Adds support for ``IsPayable`` flag in prompt.
 - Fix Block header problems with ``block_import.py`` script
+- Sync GAS price calculations with current Neo core
 
 
 [0.7.6] 2018-08-02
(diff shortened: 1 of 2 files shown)


ixje:
💯 
Is there perhaps any relationship with https://github.com/CityOfZion/neo-python/issues/440?

hal0x2328:
> Is there perhaps any relationship with #440?

No, it does not appear to be related.

PR #550, opened 2018-08-11
Fix sys_fee calculation

ixje:
Fix https://github.com/CityOfZion/neo-python/issues/440

bring sys_fee calculation inline with C# (specifically `IssueTransaction` for #440)
Also looked into all other transaction types and brought them inline where applicable

run the tests from #440

no

- [ ] Did you add any tests?


ixje:
I kind of saw that and was thinking, no tests failing so they're probably not used anymore. Anyway, lesson learned; don't rely on tests alone

ixje:
don't merge, I noticed the bug is back with this last update. Will have to figure out why actually loading the sys_fee breaks it again. will do that later on sunday

PR #555, opened 2018-08-17
Docker login to use --password-stdin for increased security

nunojusto:
For security reasons Docker login must use --password-stdin.
See: https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin

Changed the line to pipe the password to docker login command

Totally tested by many people

No

- [ ] Did you add any tests?


localhuman:
@metachris this look ok?

nunojusto:
Well, maybe it's not so important if you are using it locally on you own pc. But in a shared server it's important. And also, this removes the warning message that docker outputs on Travis. Look at logs

PR #559, opened 2018-08-19
Add Seedlist.rst to Docs and update Basicusage.rst for API Servers

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**
This adds the seedlist.rst file to the docs folder. It follows the neo-cli file closely. I used this when building my API server. Adds specific instruction for Windows WSL (Ubuntu) users.

Updates Basicusage.rst for API Servers with a Port Description and special instructions for Windows WSL (Ubuntu) users.

PR #561, opened 2018-08-20
Create Seedlist.rst and update Basicusage.rst for API Servers

jseagrave21:
Adds Seedlist.rst. Follows neo-cli but is neo-python specific. I referenced this when building my RPC server.
Updates Basicusage.rst with port description and specific instructions for Windows WSL (Ubuntu) users.

*Note this PR was re-created from this previous [PR](https://github.com/CityOfZion/neo-python/pull/559) for better comparison and merging


jseagrave21:
@ixje Okay, I recreated it. Thank you for the feedback.

jseagrave21:
I edited the API section of Basicusage.rst to give better instructions for running an API server. I did not include an exhaustive list of API commands as seen in the [NEO API Reference](http://docs.neo.org/en-us/node/cli/2.7.6/api.html) or example of how to query. If this would be useful for neo-python servers, and not redundant, I would be happy to add it as well.

localhuman:
Great work!

jseagrave21:
Thank you! Please let me know if more documentation would be helpful. I don't want to be redundant but I would also like it to be easy for someone to read and understand Neo-python instead of overloading Discord.

PR #536, opened 2018-08-08
Add test for np-prompt using pexpect

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**
Addresses issue #490 
Adds test_prompt.py for testing np-prompt with pexpect

**How did you solve this problem?**
Learned about pexpect, trial and error

**How did you make sure your solution works?**
Using unittest

**Are there any special changes in the code that we should be aware of?**
Yes, requires installation of pexpect

**Please check the following, if applicable:**

- [x] Did you add any tests?
- [x] Did you run `make lint`?
- [x] Did you run `make test`?
- [x] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

Issue #490, opened 2018-06-29
Idea: Add a test that checks np-prompt with expect


metachris:
Idea: We could add a test that checks that np-prompt/prompt.py is working and can open a wallet, using expect.

PR #563, opened 2018-08-21
Add test for np-prompt using pexpect

jseagrave21:
Addresses issue #490
Adds test_prompt.py for testing np-prompt with pexpect
Re-created from https://github.com/CityOfZion/neo-python/pull/536 for to solve compatibility problems

Learned about pexpect, trial and error

Using unittest

Yes, requires installation of pexpect

- [x] Did you add any tests?


jseagrave21:
@metachris @ixje  here is the new PR

Issue #60, opened 2017-11-06
[Feature] Add Claim functionality


localhuman:
- Add the ability to monitor unclaimed coins in a wallet
- Add the ability to claim unclaimed coins in a wallet
- Add tests to verify this functionality

localhuman:
closed in https://github.com/CityOfZion/neo-python/pull/75

PR #568, opened 2018-08-23
Update test_prompt.py

jseagrave21:
Updates test_prompt to spawn prompt.py directly instead of using the shortcut `np-prompt`. Does not change the functionality of the test.

This PR was created to achieve a passing neo-python-core build to merge PR [#60](https://github.com/CityOfZion/neo-python-core/pull/60).
cc @localhuman 

Discussion with @localhuman and tested it personally.

Tested personally with unittest.

No

- [x] Did you add any tests?


localhuman:
Thanks!!

Issue #564, opened 2018-08-22
Implement recent changes in GAS costs


localhuman:
These recent changes have been merged into neo core.  They will need to be implemented in neo-python:

- https://github.com/neo-project/neo/pull/349
- https://github.com/neo-project/neo/pull/348

PR #569, opened 2018-08-23
Update ApplicationEngine.py

jseagrave21:
Add appropriate gas cost for `VERIFY`
Add appropriate GAS cost for `CHECKMULTISIG` for an array

To meet the requirement of issue #564 .

Interpreting

Individual testing

No

- [ ] Did you add any tests?


jseagrave21:
Obviously this is not ready to merge. I will keep trying to figure out why Travis is failing. If I can't, I will close the PR.

One question I have is if my approach for determining if the item is an array using `isinstance` is valid. I am not sure what this data would look like.

neo/SmartContract/ApplicationEngine.py
```diff
@@ -292,10 +292,19 @@ def GetPrice(self):
             return 20
         elif opcode == CHECKSIG:
```
localhuman: you could do `elif opcode in [CHECKSIG,VERIFY]:`

neo/SmartContract/ApplicationEngine.py
```diff
@@ -292,10 +292,19 @@ def GetPrice(self):
...
+        elif opcode == VERIFY:
+            return 100
         elif opcode == CHECKMULTISIG:
             if self.EvaluationStack.Count == 0:
                 return 1
-            n = self.EvaluationStack.Peek().GetBigInteger()
+            
+            item = self.EvaluationStack.Peek()
+
+            array = isinstance(item, list)
```
localhuman: You'll need to check for the interop `Array` type, not the built in `list` type.
```
if isintance(item,Array):
   n = item.Count
```

jseagrave21: I can't believe it is as easy as just checking if it is an Array lol.

jseagrave21:
@localhuman Thank you for the feedback! I was toying with the idea for checking for `Array` type, but I couldn't find any reference on Google, so I figured it couldn't be that easy. lol.

neo/SmartContract/ApplicationEngine.py
```diff
@@ -290,12 +290,18 @@ def GetPrice(self):
...
             return 100
         elif opcode == CHECKMULTISIG:
             if self.EvaluationStack.Count == 0:
                 return 1
-            n = self.EvaluationStack.Peek().GetBigInteger()
+            
+            item = self.EvaluationStack.Peek()
+
+            if isinstance(item, Array):
+                n = item.count
```
localhuman: Should be `item.Count`

localhuman:
Hmm, seems like there's something else going on here causing the tests to fail.  Will take a little deeper look...

jseagrave21:
This was the same set of errors I was getting when I originally added `VERIFY`. I couldn't quickly identify why, which is why I originally commented about the Travis failures. I look also when I get back home, if you haven't figured it out by then.

jseagrave21:
@localhuman Good find. I didn't think to import `VERIFY` since I didn't see it in the neo-project PR.

jseagrave21:
@localhuman Thank you for working with me on this PR!

PR #573, opened 2018-08-25
Add new API methods to return free or priority txs in a specific block or the rawmempool

jseagrave21:
This PR adds methods to return free or priority txs in a specific block or the rawmempool.
Specifically, it adds:
- `getblockfreetxonly`
- `getblockprioritytxonly`
- `getrawmempoolfreetxonly`
- `getrawmempoolprioritytxonly`

This PR helps address issue #189 .

Trial and Error

I first built scripts to query the mainnet. They are shown below.
**NOTE:** The default functionality for these scripts is to return the free txs. If you want to return the priority txs just change the 'if' statement to `not in`.

Getblockfreetxonly:
```
import requests

url = "<the IP of the RPC node you want to query>"
body = {"jsonrpc": "2.0","method": "getblock","params": [2656128, 1],"id": 2}

r = requests.post(url, json=body)
req = r.json()
length = len(req['result']['tx']) - 1

p = 0
tx = []
while p <= length:
    if "'net_fee': '0'" in str(req['result']['tx'][p]) and "MinerTransaction" not in str(req['result']['tx'][p]):
        tx.append(req['result']['tx'][p])
        p += 1
    else:
        p += 1

print(free_tx)
```
Getrawmempoolfreetxonly:
**NOTE:** If the rawmempool is large this can take awhile, so consider testing by reducing the first while loop. For example, while `p <= length // 100`
```
import requests

url = "<the IP of the RPC node you want to query>"
body = {"jsonrpc": "2.0","method": "getrawmempool","params": [],"id": 2}

r = requests.post(url, json=body)
req = r.json()
length = len(req['result']) - 1

p = 0
transactions = []
while p <= length:
    d = requests.post(url, json={"jsonrpc": "2.0","method": "getrawtransaction","params": [str(req['result'][p]), 1],"id": 2})
    data = d.json()
    transactions.append(data['result'])
    p += 1
transactions_length = len(transactions) - 1

p = 0
tx = []
while p <= transactions_length:
    if "'net_fee': '0'" in str(transactions[p]):
        tx.append(transactions[p])
        p += 1
    else:
        p += 1

print(tx)
```

Then I made some tests. I was able to prove that I migrated the 'getblock...' methods successfully. However, I am not sure about the 'getrawmempool...' methods because there is no mempool on individual tests.

No.

- [x] Did you add any tests?


jseagrave21:
Not sure why the travis build terminated. Looks like maybe it timed out. Is there a way to restart it? cc @ixje

jseagrave21:
From my personal testing, the transactions are being extracted from the normal `getrawmempool` process; however, these transactions are not able to be found using the core processes from `getrawtransaction` as seen here https://travis-ci.org/CityOfZion/neo-python/builds/421638694#L1483

I am having trouble understanding why this is the case because this is exactly how my live RPC query scripts work and those do function correctly. cc @ixje @localhuman @metachris

jseagrave21:
Still seeing similar problems where tx is not able to be found: https://travis-ci.org/CityOfZion/neo-python/builds/421704371#L1427

jseagrave21:
@localhuman @metachris  I think the reason the tx cannot be found is because new instances of the Blockchain are being created during `make test` while the tx hashes stay in the mempool. So, when the current blockchain is called with `Blockchain.Default()` and the first hash in the mempool is queried, it isn't found with `GetTransaction` which is indicated by these notifications with the mempool shown first:
```['0x16370d94ea789379bb751beaf53d70570b4a2aac942bb34f2b6a00c27431c2da', '0x548f2823eaef7b01ee2217c6025e535320d96fd61d6b2c9a95c59944a4de0ea0', '0x34feece805c88d5f40e45615909c8beac5a49fa2e368809c94f4e13173d900b2', '0xf1fdb9052797f3e1d5d8f9c9f6347f836a58c384a5728de7d49a35c33a2207a7', '0x467e506995ff054a98f8334715b20cb548b295065fde5d4450624baea6d93c2c', '0x5c0042cb59f62b7d871741815194197338d50c711a6c4941294d035e81874a4f', '0xa0c6af41340d9ad4ea0d0a696175d7d76c0991c5a0a158d27b5e39347aade350', '0xb8c2a1c6f06cc168fc4f8a84e435e53d37640401556418f4652a6989f5585352', '0xdadc8d06726c83e96081fd71eead0da315c825def75f72e5fea3b0d575732497', '0xaffd3bdabe0971d9df0fde632593fa7834b1aa5ad3a52a1f55418cd8c074d136', '0xb1f5d3f9c8e94711b80999b4ea9a356beae8b627f6ae49016e62672c6dbd2183', '0xc08210d8385c3a63cf7769c04f1e1f4a2167bc5badc43e9d68db5b3d41ea68d0', '0xe950a280c063284fe6abdc7bd4beafbf8e3dca8ec3187a41d12142a46cd2a425', '0x9dbfc7856cebb1a0e7fce7e7a23cd35f6964ab4c4cd07f4a20e25b04f39567d0', '0x3463bfca0df077d44bc9a0259b73bfb7ec79c7ebd6f764eb890c125a73d45221']
[I 180904 23:46:42 LevelDBBlockchain:411] Could not find transaction for hash b'16370d94ea789379bb751beaf53d70570b4a2aac942bb34f2b6a00c27431c2da'
None -1
{'jsonrpc': '2.0', 'id': '2', 'error': {'code': -100, 'message': 'Unknown Transaction'}}```

The "None -1" comes directly from `GetTransaction` as does the Notification, and the message is raised in the method function body.

I think this method would work on a node with unconfirmed transactions. However, I cannot test this theory because my RPC node never has any unconfirmed transactions when it is running.

Could you please advise on a path forward, or if you are interested in these methods at all?

PR #579, opened 2018-08-27
Update pexpect

jseagrave21:
Make pexpect compatible with Python 3.7

will solve build failure seen here: https://travis-ci.org/CityOfZion/neo-python/builds/420864567#L1656

and help solve this issue #518 

A similar issue can be seen here which was solved with pexpect 4.6.0: https://github.com/pexpect/pexpect/issues/453


localhuman:
💯 I think with this Python 3.7 is now officially supported!

PR #581, opened 2018-08-29
Feature max testops

localhuman:
- Fixes vulnerability to RPC invoke functionality that can send node into unclosed loop during 'test' invokes.

- Limit amount of 'free' operations in application engine to 500k

- Added a test

- No

- [x] Did you add any tests?


vncoelho:
Great, @localhuman, step by "Ants" step.
Also thanks to @FabioRick for the devil insight :dagger:

ixje:
@localhuman was going to ask/suggest 

-  using`logger.debug` instead of `logger.error` like the other checks done at the same code location, and if you could name  
- if there should perhaps be a `teardownClass` where we revert `settings.set_loglevel(DEBUG)` back to `INFO`
- if you could rename the test case name `test_invalid_array_index` to something that actually describes the `max_free_ops` test as this seems like a copy/paste from another test.

At least this last one might be worth another push? as that can become really confusing if it fails at some point.

PR #524, opened 2018-07-24
Add Support for Peewee 3.6.4

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**
Add support for Peewee 3.6.4 (latest)
See issue #262 

**How did you solve this problem?**
Trial and error. And learning about the new capabilities of Peewee 3, mainly Blobfields.
See https://github.com/coleifer/peewee/pull/1682

**How did you make sure your solution works?**
I tested on privatenet. And Travis passes.

**Are there any special changes in the code that we should be aware of?**
Adds implementation of BlobField, which is new in Peewee 3+

**Please check the following, if applicable:**

- [ ] Did you add any tests?
No, tested added, because support for Peewee 3.6.4 should pass all current tests.

- [x] Did you run `make lint`?
- [x] Did you run `make test`?
- [x] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

PR #584, opened 2018-08-30
Bugfix open newly created wallet

localhuman:
- Wallets created after the PeeWee updates have a `MigrationState` of `b'1'` instead of `'1'` as previously, which made it impossible to open a newly created wallet. 

- This change coerces the value to an integer and compares `MigrationState` to `1` so that both versions will work.

- manual testing

- I would not be surprised if there are other similar bugs related to the wallet.

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,6 +11,8 @@ All notable changes to this project are documented in this file.
 - Update pexpect to 4.6.0 to be compatible with Python 3.7
 - Accept incoming node connections, configurable via protocol config file setting
 - Fixes vulnerability to RPC invoke functionality that can send node into unclosed loop during 'test' invokes
+- Fix issue with opening recently created wallets
+
 
 [0.7.7] 2018-08-23
 ------------------
--- a/neo/Wallets/Wallet.py
+++ b/neo/Wallets/Wallet.py
@@ -111,8 +111,7 @@ class Wallet:
                 raise Exception("Password hash not found in database")
 
             hkey = hashlib.sha256(passwordKey).digest()
-
-            if self.LoadStoredData('MigrationState') != '1':
+            if int(self.LoadStoredData('MigrationState')) != 1:
                 raise Exception("This wallet is currently vulnerable. Please "
                                 "execute the \"reencrypt_wallet.py\" script "
                                 "on this wallet before continuing")


jseagrave21:
This is because of my peewee PR. I remember having a problem with this. I removed it after @ixje asked me about it here:
https://github.com/CityOfZion/neo-python/pull/524/commits/525b1066b5b0cd7328a9c1f79ffbdf3b57dc779a

https://github.com/CityOfZion/neo-python/pull/524#pullrequestreview-145369629

jseagrave21:
I am sorry if I introduced more issues 😢

localhuman:
No worries! I think better tests could have caught the issue earlier.  Mostly I just wanted to open my wallet 😆  .  I'll keep this PR open as I look for anything similar.

jseagrave21:
Okay. I tried to include some snippets of the change to help. Mostly, the code use to prefer to store everything as a string because peewee didn't mind and would decode or not as needed. With peewee3, they added `BlobField` and basically took away the original functionality of `CharField` and `StringField`.

jseagrave21:
@localhuman during testing I am seeing these notifications:
```
neo.api.JSONRPC.test_json_rpc_api.JsonRpcApiTestCase) ... Could not save stored data <Model: Key> instance matching query does not exist:
SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS "t1" WHERE ("t1"."Name" = ?) LIMIT ? OFFSET ?
Params: ['PasswordHash', 1, 0]
Could not save stored data <Model: Key> instance matching query does not exist:
SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS "t1" WHERE ("t1"."Name" = ?) LIMIT ? OFFSET ?
Params: ['IV', 1, 0]
Could not save stored data <Model: Key> instance matching query does not exist:
SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS "t1" WHERE ("t1"."Name" = ?) LIMIT ? OFFSET ?
Params: ['MasterKey', 1, 0]
Could not save stored data <Model: Key> instance matching query does not exist:
SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS "t1" WHERE ("t1"."Name" = ?) LIMIT ? OFFSET ?
Params: ['MigrationState', 1, 0]
Could not save stored data <Model: Key> instance matching query does not exist:
SQL: SELECT "t1"."Id", "t1"."Name", "t1"."Value" FROM "key" AS "t1" WHERE ("t1"."Name" = ?) LIMIT ? OFFSET ?
Params: ['Height', 1, 0]
```
I am wondering if this is another issue with peewee? Since the "Could not save stored data" announcement comes from 
```
def SaveStoredData(self, key, value):
        k = None
        try:
            k = Key.get(Name=key)
            k.Value = value
            k.save()
        except Exception as e:
            print("Could not save stored data %s " % e)
```
What do you think?

PR #586, opened 2018-08-30
fix for import_blocks.py block hash caching issue

hal0x2328:

import_blocks.py reuses a block object when each new block is read from the chain.acc file, but once a block hash for an block object is requested for the first time, it is cached, so future deserializations of new blocks into the existing object will be associated with the wrong hash.

I added code to `Deserialize` in neo/Core/BlockBase.py to set the private `__hash` variable to `None` before deserializing a new block.

Tested import_blocks.py  - before the change, any imported block would have no transaction results since they were all stored with the wrong hash. After the change, blocks returned the correct list of transactions.

No

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ All notable changes to this project are documented in this file.
 - Accept incoming node connections, configurable via protocol config file setting
 - Fixes vulnerability to RPC invoke functionality that can send node into unclosed loop during 'test' invokes
 - Fix issue with opening recently created wallets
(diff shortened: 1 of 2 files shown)


ixje:
Is this PR https://github.com/CityOfZion/neo-python/pull/537 for the same issue not working? 🤔

hal0x2328:
#537 fixed a related issue with the block headers, this PR fixes the list of transactions of a block being missing.

localhuman:
Great work, thanks!

PR #587, opened 2018-08-31
Add `getaddresstxs` RPC method

jseagrave21:
Adds `getaddresstxs` RPC method
Keep in mind the user must designate how many tx he/she is looking for (e.g. the last 5 txs).
This is because crawling the entire blockchain for every tx for an address takes a long time.

This helps address issue #189 

Trial and Error

Unittest

No

- [x] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,6 +12,7 @@ All notable changes to this project are documented in this file.
 - Accept incoming node connections, configurable via protocol config file setting
 - Fixes vulnerability to RPC invoke functionality that can send node into unclosed loop during 'test' invokes
 - Fix issue with opening recently created wallets
+- Add `getaddresstxs` RPC method
 
 
 [0.7.7] 2018-08-23
--- a/neo/api/JSONRPC/JsonRpcApi.py
+++ b/neo/api/JSONRPC/JsonRpcApi.py
@@ -153,7 +153,7 @@ class JsonRpcApi:
 
                     for info in req[p]['vout']:
                         address = info['address']
-        
+
                         if address == params[0]:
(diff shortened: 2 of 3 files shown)


jseagrave21:
@localhuman @metachris There might be other more preferable ways to implement this. Please let me know what you think.

hal0x2328:
@jseagrave21 Mine wasn't a call to a node, it was a call to the scan.nel.group block explorer which had already done the hard work of mapping the address to the list of transactions. 

Unless I don't understand what you're doing, I think implementing this as an RPC method in a node that hasn't already indexed the address in a wallet db or something similar would put a tremendous load on the RPC node as it walked the entire blockchain just to get info on a single address.

It would be fine to have a branch with this capability for a privately-run node for your own dApp or research purposes but I wouldn't want to have this code enabled on a public RPC node for just anyone to query.

jseagrave21:
@hal0x2328 Okay, if I restricted its load on the RPC node at the expense of the user (more time) by inputting a `time.sleep(0.1)` during the loop would it be better? Or forced the user to provide more inputs (bound the problem further)?

jseagrave21:
Wow, okay. Nevermind. When I tested it using unittest, it seemed pretty quick. Probably because the chain is a lot smaller and a lot less txs. I am trying it on my local node and asked for my last 5 txs and after 15 min of working, it isn't finished. And inputting a `time.sleep(0.1)` in the loop, doesn't seem to make the node accessible during that time (probably just my misunderstanding of what that code actually does). I will close the PR but leave the branch alive, in case anyone wants to look at the script.

If anyone sees a way for me to salvage this PR so I am actually contributing something, please let me know.

Issue #522, opened 2018-07-23
Add support for incoming P2P connections


ixje:
### Current behavior
Afaik; currently only `neo-python` can setup new connections to a node, but a node cannot connect to `neo-python`.

to privnet/C# node
```
Eriks-Air: erik$ nc 192.168.1.100 20333
��version'��L��U[mO��U
                      /NEO:2.7.4/�^C
```
to neo-python
```
Eriks-Air: erik$ nc 192.168.1.100 20338
Eriks-Air:mitmproxy erik$
```

### Expected behavior
neo-python should also accept incoming connections.

### How to reproduce

point any node to `neo-python` or try to netcat to it (see above).

jseagrave21:
@ixje Hey, I am testing the new dev branch as an RPC node and am receiving a lot of error messages. This occurs the most frequently:
```
Traceback (most recent call last):
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Network/NeoNode.py", line 213, in CheckDataReceived
    self.MessageReceived(message)
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Network/NeoNode.py", line 257, in MessageReceived
    self.HandleGetHeadersMessageReceived(m.Payload)
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Network/NeoNode.py", line 510, in HandleGetHeadersMessageReceived
    self.SendSerializedMessage(Message('headers', HeadersPayload(headers=headers)))
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Network/Message.py", line 44, in __init__
    payload = binascii.unhexlify(Helper.ToArray(payload))
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Core/Helper.py", line 77, in ToArray
    value.Serialize(writer)
  File "/mnt/c/users/seagrave/neo/neo-python-dev/neo/Network/Payloads/HeadersPayload.py", line 44, in Serialize
    writer.WriteSerializableArray(self.Headers)
  File "/mnt/c/users/seagrave/neo/neo-python-dev/venv/lib/python3.6/site-packages/neocore/IO/BinaryWriter.py", line 408, in WriteSerializableArray
    item.Serialize(self)
AttributeError: 'NoneType' object has no attribute 'Serialize'
Traceback (most recent call last):
[...]

PR #589, opened 2018-09-01
resolve simultaneous port usage for #522

ixje:
Resolve simultaneous connecting from and to the same port. Part of addressing https://github.com/CityOfZion/neo-python/issues/522

changed from using the `endpoint` API to lower level `reactor` API. While the original code works in standalone mode, it seems to conflict when used in conjunction with `connectTCP`
https://github.com/CityOfZion/neo-python/blob/f003f0a66cdd10cef91f30a6cda5549844ebd013/neo/Network/NodeLeader.py#L151

manual by running the node, then typing `nodes` to verify that we connected to our seeds, then run `nc localhost 20333` and see a `version` reply.
```
Eriks-Air:~ erik$ nc 127.0.0.1 20333
Anttversion2
�7��[mO��x�/NEO-PYTHON:0.7.8-dev/
```

no

- [ ] Did you add any tests?

no new changelog entry, just polishing old PR

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,7 @@ All notable changes to this project are documented in this file.
 - Prefix ``vin`` JSON output format to match C#
 - Update ``neo-boa`` to v0.5.0 for Python 3.7 compatibility
 - Update pexpect to 4.6.0 to be compatible with Python 3.7
-- Accept incoming node connections, configurable via protocol config file setting
+- Accept incoming node connections, configurable via protocol config file setting (default: OFF)
 - Fixes vulnerability to RPC invoke functionality that can send node into unclosed loop during 'test' invokes
 
 [0.7.7] 2018-08-23
--- a/neo/Network/NodeLeader.py
+++ b/neo/Network/NodeLeader.py
@@ -130,8 +130,7 @@ class NodeLeader:
         self.peer_check_loop.start(240, now=False)
 
         if settings.ACCEPT_INCOMING_PEERS:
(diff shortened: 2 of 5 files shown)


neo/Network/test_node_leader.py
```diff
@@ -141,6 +139,7 @@ def mock_send_msg(node, message):
         with patch('twisted.internet.reactor.connectTCP', mock_connect_tcp):
             with patch('twisted.internet.reactor.callLater', mock_call_later):
                 with patch('neo.Network.NeoNode.NeoNode.SendSerializedMessage', mock_send_msg):
+                    leader.Shutdown()
```
ixje: this was the only place that seemed to work. I tried placing it in the `teardown` and just as final statement for tests using the leader instance, but no luck. Don't know why, but this worked

ixje:
ugh, it builds fine locally just not with the `pexpect` tests on travis. probably has something to do with timeouts. Will try to restart the build somewhere early in the morning when there's hopefully more resources available or something like that :/

jseagrave21:
@ixje Hey, I when I reduced the timeout from their default, I was just trying to speed up the test. You can always remove the `timeout=...` and it should default to 30 sec. Then the test should pass.

localhuman:
I think it makes sense to turn it off by default.

ixje:
@localhuman we're good now 💦

jseagrave21:
@ixje With these changes, will P2P still function, if I turn it on? Also, is there any additional benefits (to the node, to the network) by turning on P2P?

PR #588, opened 2018-09-01
Add `sendfrom` and `sendmany` RPC methods

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**
- Adds `sendfrom` RPC method and tests
- Adds `sendmany` RPC method and tests (original sendmany PR from here: https://github.com/CityOfZion/neo-python/pull/596)
- Integrates both methods with `sendtoaddress` RPC method

Helps address issue #189 and https://github.com/CityOfZion/neo-python-rpc/issues/12

**How did you solve this problem?**
Worked from https://github.com/CityOfZion/neo-python/pull/556 and adjusted accordingly

NOTE: This PR relies on `BlockchainFixtureTestCase` through `WalletFixtureTestCase`

**How did you make sure your solution works?**
Unittest

**Are there any special changes in the code that we should be aware of?**
No

**Please check the following, if applicable:**

- [x] Did you add any tests?
- [x] Did you run `make lint`?
- [x] Did you run `make test`?
- [x] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

Issue #478, opened 2018-06-26
Reduce test fixtures size to speed up tests


ixje:
### Current behavior
travis-ci takes > 15 minutes to run all tests. One reason is that the tests rely on a fixture database that's > 500MB in size. The fixture database is basically a zipped up testnet blockchain folder. This big size is the result of having testcases deployed on the regular TestNet, which then have many unnecessary blocks added to the fixtures in between different tests created on that network.

We believe that the fixture size can be easily reduced to < 50 MB by recreating all tests on a special private TestNet. This testnet should be based on [neo-privatenet-docker](https://github.com/CityOfZion/neo-privatenet-docker)
The image should only be spun up when adding new tests that requires events or interactions on a full blockchain (e.g. smart contract storage tests) to keep the database to the minimal size necessary.

TODO:
- [x] create an inventory of all tests to be ported to the private testnet (hint: look at what derives from: [BlockchainFixtureTestCase](https://github.com/CityOfZion/neo-python/blob/master/neo/Utils/BlockchainFixtureTestCase.
- [x] migrate all tests to a new private testnet docker image (based on neo-privatenet-docker)
- [x] make testnet docker image publicly available
- [x] create new fixture file, get it on CoZ AWS (ask @localhuman), update all test cases.
- [x] create PR to `neo-python`

dauTT:
Hi ixje, I would be happy to help you with this issue. Shall I start with the creation of the inventory? Question: how exactly do you want to migrate all tests using docker? A concrete example would help here as I have basic knowledge of docker. Thanks.

ixje:
Hi @dauTT that would be great!

To explain a bit more; almost all of the current tests are created on the official TestNet. Which means that there are smart contracts deployed on the official TestNet that have been interacted with in the past and we then confirm with our tests that we still get the same results. 

We want to re-create those smart contracts (as the sources are likely lost @localhuman ?), but then deploy them on a new private network created with the `neo-private-docker` image. Have a look at https://github.com/CityOfZion/neo-privatenet-docker to see what that is about.

* So the first step would be to try re-create the smart contracts belonging to the specific tests that are currently in `neo-python`. You can test and validate your new contracts either on the testnet or using a `neo-private-docker`.
[...]

PR #596, opened 2018-09-06
Add `sendmany` RPC method

jseagrave21:
Adds `sendmany` RPC method and tests

Help address Issue #189 

NOTE: This PR relies on `BlockchainFixtureTestCase` through `WalletFixtureTestCase`

Trial and Error. Also worked from neo-project docs

Unittest

No

- [x] Did you add any tests?


ixje:
While it looks good now, i think we should wait with merging, because these tests will break as soon as we switch to using #478 and remove `fixtures_v7` that the `WalletFixtureTestCase` used relies on. I'd like to see #478 get used rather sooner that later and the update accordingly.

jseagrave21:
Okay, that sounds great!

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -409,3 +419,60 @@ def list_address(self):
                 "watchonly": addr.IsWatchOnly,
             })
         return result
+
+    def send_many(self, params):
+        if len(params) not in [1, 2, 3] or type(params[0]) is not list:
```
ixje: `type(params[0])` will throw a list index out of bounds if `params` is `None` because the first check will fail and then it will still try this because of the `or`

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -409,3 +419,60 @@ def list_address(self):
...
+        data = standard_contract.Data
+        tx.Attributes = [
+            TransactionAttribute(usage=TransactionAttributeUsage.Script,
+                                 data=data)
+        ]
+        context = ContractParametersContext(
+            tx, isMultiSig=signer_contract.IsMultiSigContract
+        )
+        self.wallet.Sign(context)
+        if context.Completed:
```
ixje: please add a test for both branches

neo/api/JSONRPC/JsonRpcApi.py
jseagrave21: @ixje Fixed. Please see code as it appears here:
https://github.com/jseagrave21/neo-python/blob/47fa65d6bc3d446da8a63fa2cfb57ab1ee599d77/neo/api/JSONRPC/JsonRpcApi.py#L560

neo/api/JSONRPC/JsonRpcApi.py
jseagrave21: @ixje Done. Please see code as it appears here:
https://github.com/jseagrave21/neo-python/blob/47fa65d6bc3d446da8a63fa2cfb57ab1ee599d77/neo/api/JSONRPC/test_json_rpc_api.py#L1442

jseagrave21:
@ixje I have taken your comment here https://github.com/CityOfZion/neo-python/pull/610#pullrequestreview-158233700 and merged `sendmany` into `sendfrom`, since they also share a considerable amount of code. Next, I plan on configuring `sendmany` in Prompt to use the foundation code from JsonRpcApi. I thought JsonRpcApi should be the foundation code since `sendmany` was originally a RPC method, not a Prompt method.

The combined PR for `sendfrom` and `sendmany` is located here: https://github.com/CityOfZion/neo-python/pull/588

PR #598, opened 2018-09-07
Add `getnodestate` RPC method

jseagrave21:
Adds `getnodestate` RPC method

Follows the `state` method from Prompt.py. This is very useful for finding out the real-time status of the node.

Additionally, the data could be fed to programs. Example below:
```
{
    "jsonrpc": "2.0",
    "id": "2",
    "result": {
        "Progress": [
            758986,
            "/",
            1383999
        ],
        "Block-cache length": 0,
        "Blocks since program start": 0,
        "Time elapsed (minutes)": 9.75e-06,
        "Blocks per min": 0,
        "TPS": 0
    }
}
```

Helps address Issue #189 

Trial and Error.

Unittest. Also, ran  it on my local RPC node to verify "Block-cache-length", "Blocks since program start", "Blocks per min", and "TPS".

No.

- [x] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.7.9-dev] in progress
 -----------------------
--
+- Add `getnodestate` RPC method
 
 
 [0.7.8] 2018-09-06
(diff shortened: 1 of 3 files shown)


jseagrave21:
@ixje Could you let me know if you think this useful  and if you like the format (specifically, the format for "Progress")  and if there was any other info which would be useful from a node? I was trying to make each part a possible input to someone's program but the json formatting is a bit odd for the [height, "/", headerheight]

Issue #606, opened 2018-09-13
Sending 0 NEO results in "Unhandled error in Deferred"


jseagrave21:
### Current behavior
As shown:
```
neo> send neo AHJxoC87MMGJHw353SMzZQLLuhR5QZ5Jps 0
sending with fee: 0.0
[Password]> ***
[I 180912 23:38:16 Transaction:617] Verifying transaction: b'115b0d3cab492b9dffe87ed8726c09c6d76d4053768ea42973f7440f3c8ddd46'
Relayed Tx: 115b0d3cab492b9dffe87ed8726c09c6d76d4053768ea42973f7440f3c8ddd46
neo> Unhandled error in Deferred:

Unhandled error in Deferred:

Unhandled error in Deferred:

Unhandled error in Deferred:
```

### Expected behavior
This should not result in an "unhandled" error.

### How to reproduce

Try sending 0 NEO to a valid address.

### Your environment

- OS: Window 10 Pro
- neo-python version: v0.7.8-dev
- Python version: 3.7.0

PR #607, opened 2018-09-13
Ensure non-zero send value in prompt.py

jseagrave21:
Ensures a user cannot send 0 of an asset

Addresses Issue #606 

NOTE: This PR's test relies on `BlockchainFixtureTestCase` through `WalletFixtureTestCase`

Trial and Error

testing on privatenet and unittest

No

- [x] Did you add any tests?


neo/Prompt/Commands/Send.py
```diff
@@ -31,6 +31,10 @@ def construct_and_send(prompter, wallet, arguments, prompt_password=True):
         address_to = get_arg(arguments, 1)
         amount = get_arg(arguments, 2)
 
+        if int(float(amount)) == 0:
```
ixje: just `float(amount)` is enough, no need to `int()` as well
```
>>> float("0.0") == 0
True
>>> float("0") == 0
True
```

neo/Prompt/Commands/Send.py
```diff
@@ -31,6 +31,10 @@ def construct_and_send(prompter, wallet, arguments, prompt_password=True):
...
+            print("invalid amount format")
```
ixje: can we make the message more direct like; "amount cannot be 0" or something along those lines. As this message might suggest "0" is not valid and should be 0.0 or something.

neo/Prompt/Commands/Send.py
jseagrave21: Okay, thank you for the feedback! I really appreciate it!

neo/Prompt/Commands/Send.py
jseagrave21: Absolutely. Thank you for the feedback!

ixje:
Do not merge until #478 is finished and merged

ixje:
great! Yes I like your suggestion but you can put that in a separate PR.

PR #610, opened 2018-09-14
Adds `sendmany` feature to Prompt.py

jseagrave21:
Adds `sendmany` feature to Prompt.py including options for `from-addr` and `change-addr` and all previous supported features from `send` command (e.g. support for multi-sig wallets and user attributes)

Trial and Error

Quite a bit of personal testing on privatenet and implemented a few tests.

NOTE: A couple of tests rely on `WalletFixtureTestCase`

No

- [x] Did you add any tests?


jseagrave21:
@ixje This PR is ready for another review. Please let me know if I am on the right track.

neo/Prompt/Utils.py
```diff
@@ -131,6 +131,38 @@ def get_from_addr(params):
...
 
 
+def get_change_addr(params):
+    to_remove = []
+    change_addr = None
+    for item in params:
+        if '--change-addr=' in item:
+            to_remove.append(item)
+            try:
+                change_addr = item.replace('--change-addr=', '')
```
ixje: if the return of this functions expects at best 1 `change_addr` then we do not need the `to_remove` list and loop over it. We can  do something like:
```
for item in params:
   change_addr = None
   if '--change-addr' in item:
      params.remove(item)
      change_addr = item.replace('--change-addr=','')
      return params, change_addr
```
This way we stop the moment we found the a change address and avoid unnecessary looping over the remaining params. 

What is the `Try/except` catching exactly? It's generally bad practice to use `Exception` (==catch all) and we should prefer being more specific (like `ValueError`or `TypeError`)

neo/Prompt/Utils.py
```diff
@@ -131,6 +131,38 @@ def get_from_addr(params):
...
+            except Exception as e:
+                pass
+    for item in to_remove:
+        params.remove(item)
+
+    return params, change_addr
+
+
+def get_outgoing(params):
```
ixje: same comments as for `get_change_addr`
1) if we only return 1 `outgoing` we don't need to push 1 item in a list. 
2) try to make the exception specific

neo/bin/prompt.py
```diff
@@ -594,7 +595,18 @@ def show_wallet(self, arguments):
             print("Wallet: '{}' is an invalid parameter".format(item))
 
     def do_send(self, arguments):
-        construct_and_send(self, self.Wallet, arguments)
+        try:
+            contract_tx, scripthash_from, fee, owners, user_tx_attributes = construct_send_basic(self, self.Wallet, arguments)
+        except Exception as e:
```
ixje: what specific errors are we catching?

neo/bin/prompt.py
```diff
@@ -594,7 +595,18 @@ def show_wallet(self, arguments):
...
+            return
+        process_transaction(self, self.Wallet, contract_tx=contract_tx, scripthash_from=scripthash_from, fee=fee, owners=owners, user_tx_attributes=user_tx_attributes)
+
+    def do_send_many(self, arguments):
+        try:
+            contract_tx, scripthash_from, scripthash_change, fee, owners, user_tx_attributes = construct_send_many(self, self.Wallet, arguments)
+        except Exception as e:
```
ixje: same as above

neo/Prompt/Commands/Send.py
```diff
@@ -14,81 +15,165 @@
...
+    fee = Fixed8.Zero()
+    if priority_fee is not None:
+        fee = priority_fee
+        if fee < Fixed8.Zero():
+            print("fee cannot be negative")
             return False
 
+    output = TransactionOutput(AssetId=assetId, Value=f8amount, script_hash=scripthash_to)
+    contract_tx = ContractTransaction(outputs=[output])
+    return contract_tx, scripthash_from, fee, owners, user_tx_attributes
```
ixje: the return signature of a function must be concise. Meaning you can't have success return a 5 argument tuple while failure returns a single boolean. This will exceptions when you try to use the command as follows:
 ```
contract_tx, scripthash_from, fee, owners, user_tx_attributes = construct_send_basic(prompt, wallet, bad_args)
```
`bad_args` will cause the function to returns  `False` and then you'll get an exception saying something like (from the top of my head):
```
TypeError: 'bool' object is not iterable
```
because it's expecting an tuple (which is iterable) to unpack 5 values from and instead gets a boolean

File: neo/Prompt/Commands/Send.py [hunk 6]
```diff
@@ -14,81 +15,165 @@
...
 
         fee = Fixed8.Zero()
         if priority_fee is not None:
             fee = priority_fee
+            if fee < Fixed8.Zero():
+                print("fee cannot be negative")
+                return False
 
         print("sending with fee: %s " % fee.ToString())
+    return contract_tx, scripthash_from, scripthash_change, fee, owners, user_tx_attributes
```
ixje: same remark on function return signature as before

neo/Prompt/Utils.py
jseagrave21: The format for `get_change_addr` followed the example of the pre-existing `get_from_addr`. I will work on improving these utility functions.

neo/Prompt/Utils.py
jseagrave21: I think I addressed this issue.

jseagrave21:
@ixje Okay, this PR is ready for another review

neo/bin/prompt.py
```diff
@@ -142,6 +142,7 @@ class PromptInterface:
                 'wallet split {addr} {asset} {unspent index} {divide into number of vins}',
                 'wallet close',
                 'send {assetId or name} {address} {amount} (--from-addr={addr}) (--fee={priority_fee})',
+                'sendmany (--outgoing={number outgoing}) (--change-addr={addr}) (--from-addr={addr}) (--fee={priority_fee})',
```
ixje: we have to think of a clearer way to explain this function. I had to read the code to understand that `outgoing` is a number that collects the `outputs_array` of the [C# description](http://docs.neo.org/en-us/node/cli/2.7.6/api/sendmany.html). For command arguments placing something between `()` parentheses means it's an optional argument. I'm pretty sure that's not the case here. I also can't get it to work:

```
neo> sendmany 0 neo neo
Asset id not found
neo> sendmany 1 neo neo
Asset id not found
neo> sendmany 1 neo
Not enough arguments
neo> sendmany 1 neo wut
Asset id not found
neo> sendmany
Not enough arguments
neo> sendmany --outgoing=0 neo bla
Asset id not found
```

neo/Prompt/Commands/Send.py
```diff
@@ -14,85 +15,155 @@
...
+    return [contract_tx, scripthash_from, scripthash_change, fee, owners, user_tx_attributes]
 
-        output = TransactionOutput(AssetId=assetId, Value=f8amount, script_hash=scripthash_to)
-        tx = ContractTransaction(outputs=[output])
 
-        ttx = wallet.MakeTransaction(tx=tx,
-                                     change_address=None,
-                                     fee=fee,
-                                     from_addr=scripthash_from)
+def process_transaction(prompter, wallet, contract_tx, scripthash_from=None, scripthash_change=None, fee=None, owners=None, user_tx_attributes=None):
```
ixje: can you please remove the `prompter` argument. It's not used. Same goes for all other functions in the `Send` file, they have `prompter` argument that's not used

neo/bin/prompt.py
jseagrave21: --outgoing is the only required command because it sets up how many iterations (transactions) prompt will ask for. I will update with what might be a better description. The test `send_many_good_simple` illustrates how the function should work https://github.com/jseagrave21/neo-python/blob/ee6c37c9aa2b4ef3c39977e469f5057d96dc1dce/neo/Prompt/Commands/tests/test_send_command.py#L336

jseagrave21:
@ixje Okay, this PR is ready for another review

jseagrave21:
@ixje Okay, I think this PR is ready for another review. Below is what I envision a complex implementation of `sendmany` should look like in prompt:
```
neo> sendmany 2 --change-addr=AWygZ1B5c3GDiLL6u5bHSVU45Ya1SVGX9P --from-addr=AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y --fee=0.
005
Outgoing Number  1
Asset to send: neo
Address to: APRgMZHZubii29UXF9uFa6sohrsYupNAvx
Amount to send: 500
Outgoing Number  2
Asset to send: gas
Address to: APRgMZHZubii29UXF9uFa6sohrsYupNAvx
Amount to send: 100
sending with fee: 0.005
[Password]> ***
[I 181009 16:39:06 Transaction:617] Verifying transaction: b'fef0bdb63d6e16141b839ef730f09519e47657aba1acce3acbbb922cd9a33140'
Relayed Tx: fef0bdb63d6e16141b839ef730f09519e47657aba1acce3acbbb922cd9a33140

neo> tx fef0bdb63d6e16141b839ef730f09519e47657aba1acce3acbbb922cd9a33140
{
    "txid": "0xfef0bdb63d6e16141b839ef730f09519e47657aba1acce3acbbb922cd9a33140",
    "size": 437,
    "type": "ContractTransaction",
    "version": 0,
    "attributes": [
        {
            "usage": 32,
            "data": "23ba2703c53263e8d6e522dc32203339dcd8eee9"
        }
    ],
    "vout": [
        {
            "n": 0,
            "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
            "value": "500",
            "address": "APRgMZHZubii29UXF9uFa6sohrsYupNAvx"
        },
        {
            "n": 1,
            "value": "100",
            "address": "APRgMZHZubii29UXF9uFa6sohrsYupNAvx"
        },
        {
            "n": 2,
            "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
            "value": "99999390",
            "address": "AWygZ1B5c3GDiLL6u5bHSVU45Ya1SVGX9P"
        },
        {
            "n": 3,
            "value": "25903.995",
            "address": "AWygZ1B5c3GDiLL6u5bHSVU45Ya1SVGX9P"
        }
    ],
    "vin": [
        {
            "txid": "0x205d2a897b8adc1e1401a04c20a018eff5cf5324833f37c839aeb8a96f62fcd4",
            "vout": 2
        },
        {
            "txid": "0x205d2a897b8adc1e1401a04c20a018eff5cf5324833f37c839aeb8a96f62fcd4",
            "vout": 3
        }
    ],
    "sys_fee": "0",
    "net_fee": "0.005",
    "scripts": [
        {
            "invocation": "40690317762c605c1ee412bf5c3ce61dddf55c61aa47532babdf67c6a75582a4e371900f22eed299a5f7aad08c2258b3368207bee7fb7616369f944169b5208fbf",
        }
    ],
    "height": 9426,
    "unspents": [
        {
            "n": 0,
            "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
            "value": "500",
            "address": "APRgMZHZubii29UXF9uFa6sohrsYupNAvx"
        },
        {
            "n": 1,
            "value": "100",
            "address": "APRgMZHZubii29UXF9uFa6sohrsYupNAvx"
        },
        {
            "n": 2,
            "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
            "value": "99999390",
            "address": "AWygZ1B5c3GDiLL6u5bHSVU45Ya1SVGX9P"
        },
        {
            "n": 3,
            "value": "25903.995",
            "address": "AWygZ1B5c3GDiLL6u5bHSVU45Ya1SVGX9P"
        }
    ]
}
```

neo/Prompt/Commands/tests/test_send_command.py
```diff
@@ -182,21 +299,219 @@ def test_13_utilst_bad_type(self):
...
+            wallet = self.GetWallet1(recreate=True)
+            args = ["2"]
+
+            framework = construct_send_many(wallet, args)
+            res = process_transaction(wallet, contract_tx=framework[0], scripthash_from=framework[1], scripthash_change=framework[2], fee=framework[3], owners=framework[4], user_tx_attributes=framework[5])
+
+            self.assertTrue(res)  # verify successful tx
+
+            json_res = res.ToJson()
+            self.assertEqual(self.watch_addr_str, json_res['vout'][0]['address'])  # verify correct address_to
```
ixje: Is it just me going crazy or have i seen this before somewhere? It feels like I'm reviewing something I've done before in another PR. Note this is not a request to change something, I think I just seem to have lost track

neo/Prompt/Commands/tests/test_send_command.py
```diff
@@ -38,129 +39,245 @@ def GetWallet1(cls, recreate=False):
...
+            wallet = self.GetWallet1(recreate=True)
+            args = ['gas', self.watch_addr_str, '2', '--tx-attr={"usage":241,"data":"This is a remark"}']
 
-        self.assertTrue(res)
+            framework = construct_send_basic(wallet, args)
+            res = process_transaction(wallet, contract_tx=framework[0], scripthash_from=framework[1], fee=framework[2], owners=framework[3], user_tx_attributes=framework[4])
 
-        self.assertEqual(1, len(res.Attributes))
+            self.assertTrue(res)
+            self.assertEqual(2, len(res.Attributes))
```
ixje: can you please add a comment above saying:
```
By default the script_hash of the transaction sender is added to the TransactionAttribute list, therefore the Attributes length is `count` + 1
```
We only have to do this once, but it's something that raises eyebrows if you read this code without it. Because it looks like you add just 1 attribute and then test for 2

neo/Prompt/Utils.py
```diff
@@ -85,13 +85,14 @@ def get_asset_id(wallet, asset_str):
 
 
 def get_asset_amount(amount, assetId):
-    f8amount = Fixed8.TryParse(amount)
+    f8amount = Fixed8.TryParse(amount, require_positive=True)
     if f8amount is None:
         print("invalid amount format")
```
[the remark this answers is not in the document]
jseagrave21: Since this is the utility function I will leave this message and remove the message in `Send.py`.

neo/Prompt/Commands/tests/test_send_command.py
jseagrave21: Probably. I derived these tests from my learning experience from implementing `sendmany` for JsonRpcApi

neo/Prompt/Utils.py
```diff
@@ -85,13 +85,14 @@ def get_asset_id(wallet, asset_str):
...
+        return False
 
     elif f8amount.value % pow(10, 8 - Blockchain.Default().GetAssetState(assetId.ToBytes()).Precision) != 0:
         print("incorrect amount precision")
```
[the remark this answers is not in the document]
jseagrave21: Since this is the utility function I will leave this message and remove the message in `Send.py.`

jseagrave21:
Okay, in addition to your requested changes I also sent the `print("insufficient funds")` to `logger.debug()` because I noticed an error is already raised, which provided an even better description including the assetId.

ixje:
Thanks! 🥇

ixje:
Can you please squash your 46 commits into a single one? If you didn't like the SourceTree blog I linked you previously then you might like a command line one: [how to squash](https://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit)

Issue #605, opened 2018-09-13
Discrepancies with multiple addresses in a wallet


jseagrave21:
### Current behavior
Using `account` to display an address's balance hides other balances within the wallet and makes them inaccessible. Here is what I see:
```
Wallet {
    "path": "neo-privnet.wallet",
    "addresses": [
        {
            "version": 0,
            "script_hash": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
            "frozen": false,
            "votes": [],
            "balances": {
                "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b": "99999900.0",
                "0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7": "74674.9999"
            },
            "is_watch_only": false
        },
        {
            "version": 0,
            "script_hash": "AZGLHgV4sBsAQLYioetrJpTeSFNLsGzJdR",
            "frozen": false,
            "votes": [],
            "balances": {
                "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b": "100.0",
                "0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7": "4.995"
            },
            "is_watch_only": false
        },
        {
            "address": "AHJxoC87MMGJHw353SMzZQLLuhR5QZ5Jps",
            "script_hash": "10d951cc11ce0d3b562acd0c0bbb183707fff64c",
            "tokens": null
        }
    ],
    "height": 9572,
    "percent_synced": 100,
    "synced_balances": [
        "[NEO]: 99999900.0 ",
        "[NEOGas]: 74629.9999 "
    ],
    "public_keys": [
        {
            "Address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
            "Public Key": "031a6c6fbbdf02ca351745fa86b9ba5a9452d785ac4f7fc2b7548ca2a46c4fcf4a"
        },
        {
            "Address": "AZGLHgV4sBsAQLYioetrJpTeSFNLsGzJdR",
            "Public Key": "0322bea896b923fb37a80dcb79d22061af85751b7140878987110b988e9a3e353f"
        },
        {
            "Address": "AHJxoC87MMGJHw353SMzZQLLuhR5QZ5Jps",
[...]

Issue #578, opened 2018-08-26
`getaccountstate` returns the wrong balance of NEO


jseagrave21:
### Current behavior
`getaccounstate` returns the wrong balance of NEO. NEO-cli returns the correct balance.
GAS balance seems to be correct

PM me on Discord if you want a specific address to query (I only figured this out because I queried one of my addresses)

### Expected behavior

`getaccountstate` should return the correct balance of both NEO and GAS

### How to reproduce

You can use this script in Python to test if the balances match:

```
import requests

url = "http://108.252.121.18:10332"
body = {"jsonrpc": "2.0","method": "getaccountstate","params": ["<address>"],"id": 2}

req = requests.post(url, json=body)
req = req.json()
data = req['result']

print('Python Data: ', data)

url = "http://18.197.197.10:10332"
body = {"jsonrpc": "2.0","method": "getaccountstate","params": ["<address>"],"id": 2}

req = requests.post(url, json=body)
req = req.json()
data = req['result']

print('NEO-cli: ', data)
```

### Your environment

This is based on live RPC calls

localhuman:
You'll need to specify an address with a mismatch.

jseagrave21:
@localhuman I PM'd you.

wy:
NEO and GAS are based on UTXO so only problem I can think of is the 'dead fork' incident.

@jseagrave21 if you know the block number when the transfer happened, then you can use getblockhash to compare. They should mismatch for faulty vs correct nodes.

jseagrave21:
@wy I spoke with @localhuman and am going to try to import the address as watch only so I can compare transaction history that way. I check the most recent transactions and it is not that.

jseagrave21:
Although I haven't confirmed completely, I think this issue is caused by missing block information for some RPC nodes. For example, block 2669088 is missing from one RPC server I queried.

I queried 3 different RPC nodes running NEO-Python and got 2 similar responses and 1 different and none of them were correct.

I queried 2 different RPC nodes running NEO-Cli and got 2 similar responses and both were correct.

jseagrave21:
When querying http://108.252.121.18:10332 about 2669088, I get the following:

for `getblockhash`, I receive: {'jsonrpc': '2.0', 'id': 2, 'error': {'code': -32603, 'message': "'>=' not supported between instances of 'str' and 'int'"}}

[...]

PR #611, opened 2018-09-17
Persist reset fix

localhuman:
- Balance mismatches in neo/gas were being caused by the resetting of storages during failed smart contract invocations.  

- `DBCollection` now has a method `MarkForReset` which caches a change-set that can be reverted back to in the case of a failed smart contract invocation.

- Manually verify balances that were previously incorrect

- Currently re-syncing mainnet Chain and Notification databases, will add another PR once those are complete.

- [ ] Did you add any tests?


jseagrave21:
@localhuman This looks great! It looks like it should address #578 but will it also fix the issues I am seeing in #605 ?

localhuman:
Not sure about #605.  I frequently use wallets with multiple addresses and have not seen that behavior.

PR #614, opened 2018-09-18
Add `gettxhistory` RPC method

jseagrave21:
Adds `gettxhistory` RPC method which returns a list of every tx in the associated wallet in JSON format, including block_index and blocktime

This PR helps address issue #189 

Trial and Error.

Unittest

No

- [x] Did you add any tests?


jseagrave21:
I added block_index and blocktime to this method to help a user sort and filter the results. I am thinking about sorting through transactions when filing taxes. I think this method will be very helpful. Please let me know if you have any thoughts or recommendations. cc @localhuman @ixje

localhuman:
I would prefer that any non standard RPC methods ( methods not implemented in `neo-cli` RPC ) to not be in the core RPC implementation.  Could you make an extended version of `JsonRpcApi` with this additional method ( along with #598 and #573 ).  We could then add a switch in `np-api-server` to optionally start the extended JsonRpcApi.

PR #615, opened 2018-09-18
Fix unhandled deferred errors and update seedlist with new NGD nodes

localhuman:
- many times when running `np-prompt` we see `Unhandled error in deferred`.  This PR aims to solve that by turning all tasks handed to `reactor` into objects that we can add an `errback` to.

- Also, update mainnet seedlist to new list of NGD nodes

- manual testing

- After this is merged, there may still be cases where `Unhandled error in deferred` could show up.  In this case these errors should be reported as issues so we can fix them.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Network/NeoNode.py
+++ b/neo/Network/NeoNode.py
@@ -4,7 +4,8 @@ import traceback
 from logzero import logger
 from twisted.internet.protocol import Protocol
 from twisted.internet import error as twisted_error
-from twisted.internet import reactor, task
+from twisted.internet import reactor, task, threads
+
 from neo.Core.Blockchain import Blockchain as BC
 from neocore.IO.BinaryReader import BinaryReader
 from neo.Network.Message import Message
@@ -213,7 +214,6 @@ class NeoNode(Protocol):
             self.MessageReceived(message)
 
         except Exception as e:
-            traceback.print_exc()
             self.Log('Error: Could not extract message: %s ' % e)
             return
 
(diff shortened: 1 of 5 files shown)


neo/Network/NodeLeader.py
```diff
@@ -203,6 +211,9 @@ def RemoveConnectedPeer(self, peer):
         if peer in self.Peers:
             self.Peers.remove(peer)
 
+    def onSetupConnectionErr(self, err):
+        print("On setup connectioin error! %s" % err)
```
ixje: typo `connectioioioiooion`
I'm afraid this might spam prints in console when clients don't respond and we keep retrying. should this perhaps be a Log/Debug print instead?

localhuman:
Sorry, my `i` key is doing very odd things lately :/

Anyways, after a few more hours I've now seen an unhandled error again.

jseagrave21:
This is great! I was wondering if I should be seeing all those errors. Lol

ixje:
@localhuman did it not create any log/debug print? I believe that if an exception is raised that is not handled by the deferred (printing is not necessarily handling)

neo/bin/prompt.py
```diff
@@ -1150,8 +1152,9 @@ def main():
     cli = PromptInterface(fn_prompt_history)
 
     # Run things
-    #    reactor.suggestThreadPoolSize(15)
-    reactor.callInThread(cli.run)
+
+    d = reactor.callInThread(cli.run)
```
ixje: do we want to add an `errBack` here incase some exception happens that we don't handle?

neo/Network/NeoNode.py
```diff
@@ -256,19 +258,18 @@ def MessageReceived(self, m):
         elif m.Command == 'getheaders':
             self.HandleGetHeadersMessageReceived(m.Payload)
         elif m.Command == 'headers':
-            headerResultDeferred = threads.deferToThread(self.HandleBlockHeadersReceived, m.Payload)
-            headerResultDeferred.addErrback(self.onThreadDeferredErr)
+            reactor.callFromThread(self.HandleBlockHeadersReceived, m.Payload)
```
ixje: I think this might have to be `callInThread` because `self.HandleBlockHeadersReceived` is not an already running thread.

`callFromThread` is described as:

> From another thread, execute a function in the reactor thread.
> Use callFromThread to call reactor APIs from outside the reactor thread. For ex‐
> ample, use callFromThread when:
> • writing data out through a transport from another thread
> • invoking a custom log observer from another thread
> • stopping the reactor from another thread

whereas `callInThread`says:

> Execute a blocking function in its own thread
which is what I believe we're after.

localhuman:
I think the problem was using `stop` on `task.LoopingCall` rather than cancelling the deferred object.  Running these changes I'm not seeing the issue anymore, but will continue for a while longer to make sure.

Issue #609, opened 2018-09-14
change the default testmode of the run function of the smartcontract virtual machine


edwardz246003:
In the smartcontract virtual machine, the default testmode of run function is True. This will make the RPC invokescript with infinite gas. It should be change to false.

  
https://github.com/CityOfZion/neo-python/blob/978f77922247d2d85c63dfe8db66e856fcb56582/neo/SmartContract/ApplicationEngine.py#L409

ixje:
thanks!

ixje:
fixed by https://github.com/CityOfZion/neo-python/pull/581
specifically:
https://github.com/CityOfZion/neo-python/blob/978f77922247d2d85c63dfe8db66e856fcb56582/neo/SmartContract/ApplicationEngine.py#L239-L242

PR #621, opened 2018-09-19
Disable infinite gas via RPC

ixje:
address issue described in #609 

change testmode to default to false

make test

no

- [ ] Did you add any tests?


localhuman:
This is already fixed by https://github.com/CityOfZion/neo-python/pull/581

ixje:
ah yes, good point!

Issue #608, opened 2018-09-14
Disparity between local and RPC responses for `getaccountstate`


jseagrave21:
### Current behavior
When querying neo-privnet.wallet (AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y) via RPC is get this following correct response, which includes the address's script hash.
```
{
    "jsonrpc": "2.0",
    "id": 2,
    "result": {
        "version": 0,
        "script_hash": "0xe9eed8dc39332032dc22e5d6e86332c50327ba23",
        "frozen": false,
        "votes": [],
        "balances": [
            {
                "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                "value": "99999850"
            },
            {
                "asset": "0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7",
                "value": "74649.9949"
            }
        ]
    }
}
```

But, when I use `account` in prompt.py, which calls `getaccountstate`, I get the following, which does not return the address's script hash.
```
{
    "version": 0,
    "script_hash": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
    "frozen": false,
    "votes": [],
    "balances": {
        "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b": "99999850.0",
        "0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7": "74649.9949"
    }
}
```
Consequently, since `getaccountstate` is called when using the `wallet` command, the address script hashes aren't displayed there either.

Additionally, a user cannot actually use a script hash to query `getaccountstate` from prompt.py or via RPC. I think that `getaccountstate` should be able to accept script hashes because that is what the function params call for.

### Expected behavior
Using the `account` command should return the script_hash of the address as intended.
And, I think that `getaccountstate` should be able to accept script hashes because that is what the function params call for (i.e.  `def GetAccountState(self, script_hash, print_all_accounts=False):`)

### How to reproduce
Query any address using `getaccountstate` via RPC and compare to `account` in prompt.py
Try to request an account state using an address's script hash.

### Your environment

- OS: Windows 10 Pro
- neo-python version: v0.7.8-dev
- Python version: Python 3.7.0

jseagrave21:
I am working on this now.

PR #627, opened 2018-09-21
Fix AccountState

jseagrave21:
This PR addresses Issue #608 

Primarily it fixes 'script_hash' so that output is actually the address string's scripthash

Secondarily it:
- Adds an 'address' key, so the address is also displayed so `account` and `wallet` are more user-friendly
- Adds a test for `ToJson` in AccountState
- Replaces the variable `script_hash` in `GetAccountState` in LevelDBBlockchain with `address` because the input is an address not a scripthash
- Adds a test for `GetAccountState` in LevelDBBlockchain

- Trial and Error
- discussion with @ixje concerning the intended functionality of `getaccountstate` and `script_hash`
- @hal0x2328 helped me understand that the privatenet consensus nodes running in docker are running neo-cli, not neo-python.

Unittest

No

- [x] Did you add any tests?


jseagrave21:
@localhuman @ixje This PR is ready for another review.

PR #633, opened 2018-09-22
Add `dumpprivkey` RPC method

jseagrave21:
Adds the `dumpprivkey` RPC method

This PR helps address Issue https://github.com/CityOfZion/neo-python/issues/189

Trial and Error

Unittest

no

- [x] Did you add any tests?


jseagrave21:
@dauTT is working on implementing the basic HTTP authentication via Twisted. I was able to implement this method into a local version of my `ExtendedJsonRpc`.

jseagrave21:
@dauTT I would prefer **not** to include the username credential, since this is more sensitive data being passed. Plus, it should make user interaction easier (less input required).

Also, how would this request look if using `curl` or something similar?

dauTT:
@jseagrave21 ,

with `curl` the two examples above look as flow:
1. **`geblockcount`**: 
```
curl -X POST http://0.0.0.0:1033 -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 5, "method": "getblockcount", "params": [] }'
```
   **result**:
```
{"jsonrpc": "2.0", "id": 5, "result": 12895}
```

2. **`dumpprivkey`**: 
```
curl -POST -u joe http://0.0.0.0:1033 -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "id": 5, "method": "dumpprivkey", "params": ["AJQ6FoaSXDFzA6wLnyZ1nFN7SGSN2oNTc3"] }'
Enter host password for user 'joe':
```
(The password is not visible while you are entering it.)
**result**:
```
{"jsonrpc": "2.0", "id": 5, "result": "Ky94Rq8rb1z8UzTthYmy1ApbZa9xsKTvQCiuGUZJZbaDJZdkvLRV"}
```

jseagrave21:
@dauTT okay, that looks great! A user could use something like `requests` to automate the login process.

@ixje I am leaning towards not requiring a username. What do you think?

Also, now that `ExtendedJsonRpc` is merged I will move this method to it and update the PR.

ixje:
All regular methods should behave as before, meaning without having to specify any additional arguments or whatsoever. If I understand you correctly, that's now the case so that is good. Only this dangerous method should require extra action and I'm fine with having username/pw as there should be a barrier to using this.

Does `basic-http` from `twisted` support a (exponential) backoff timer? Meaning whenever they fail to authenticate, they have to wait a certain amount of time before they can try again. if they fail again, the wait time goes up drastically. The reason for that is we want to avoid the credentials getting brute forced over rpc.

dauTT:
@ixje , 

I can implement such backoff timer. It shouldn't be a problem. @jseagrave21  are you also happy with @ixje proposal? (Authentication will use username + password + backoff timer)

jseagrave21:
@dauTT Yes, I guess the only question I have is how does the wallet/server owner set the wallet "username"?

dauTT:
The username can be entered in a similar way we enter the wallet password or if you don't mind we can pass it as a argument.

jseagrave21:
@dauTT okay, do you have a branch or PR to my branch I could use to integrate your code?

dauTT:
The code is not ready yet. I need to do some clean up and refactoring. Once done I will share it with you.

jseagrave21:
@dauTT it is supposed to be in `ExtendedJsonRpcApi`. That is where @ixje
wanted it.

jseagrave21:
@dauTT Awesome! I will see what I can do.

jseagrave21:
@dauTT After my initial look I am wondering how hard will it be if we want to add basic HTTP authentication to other normal RPC methods like `sendtoaddress`, `sendfrom`, or `sendmany`?

dauTT:
Very easy, we just need to move them to ExtendedJsonRPCAuth, without thinking. :smile:

jseagrave21:
@dauTT I saw that was an option but the original idea of extended methods
was that those were not in parody with neo-cli. @ixje is this option
acceptable?

Secondly, @dauTT could you describe a little more how to address making the
method asynchronous? I will admit I may be out of my league but I'd like to
try.

dauTT:
Yes, also in that case it should not be a problem. with the `AbstractJsonRpcApi` class at our disposal we can partition our methods as we want:

**with option ``--extended-rpc``**:
- `ExtendedJsonRpcApi` class will contain methods which do not required authentication
- `ExtendedJsonRpcApiAuth` class will contain methods which required authentication (e.g, `dumpprivkey`)

**without any option**:
- `JsonRpcApi` class will contain methods which do not required authentication
- `JsonRpcApiAuth` class will contain methods which required authentication. (e.g `sendtoaddress`, `sendfrom`, or `sendmany`?)

In regards  to adding the asynchronous feature, maybe you can get some good idea by reading this book, [Twisted_Network_Programming_Essentials.pdf](https://zodml.org/sites/default/files/Twisted_Network_Programming_Essentials.pdf). In particular:
- **ch3. Writing Asynchronous Code with Deferreds.**
- **Example 9-2. db_checker.py** (pag 87)

jseagrave21:
@dauTT Okay, thank you. Sounds good!

ixje:
I do not know yet how you solved the basic authentication, but an idea could be to make a decorator. That way you do not need the additional `ExtendedJsonRpcApiAuth` and `JsonRpcApiAuth` classes, but can simply apply the decorators to whatever extended function we feel necessary. e.g.
```
    @require_basic_auth
    def dump_priv_key(self, params):
        if not params or params[0] == '':
            raise JsonRpcError(-100, "Missing argument")
```

dauTT:
@ixje ,

I thought about the `decorator` solution (#1) before making my proposal and I came to the conclusion that pursuing this route will require a lot of refactoring and we will need at least one additional entry point (e.g `@app.route('/auth')`) beside the [default one](https://github.com/CityOfZion/neo-python/blob/master/neo/api/JSONRPC/JsonRpcApi.py#L119): 

```
    @app.route('/')
    @json_response
    @cors_header
    def home(self, request):
        ....
``` 
So I give up this idea.

Another potential solution (#2) which come out of the box is to have two apps in each classes:
```
class JsonRpcApi:
    app = Klein()
    app_auth = Klein()
    ...
```

```
class ExtendedJsonRpcApi:
    app = Klein()
    app_auth = Klein()
```
It is definitely not elegant but it will work.

The [original solution](https://github.com/CityOfZion/neo-python/pull/633#issuecomment-429653412) (#3) has 4 classes but one one app. So there is a trade off in each solutions :
#1 (decorator): elegant solution, refactoring intensive, at least 2 entries points.
#2 (2 apps for each classes): out of the box solution, not elegant.
#3 (4 classes and one app): out of the box solution, not elegant.

ixje:
I think a second entry point might just be what's needed. In a way, the more barriers the better. I'll see if I can have a discussion about this regarding the C# implementation and protection

dauTT:
@ixje ,

I thought again. :smile: 

we can also have one 5) solution with only one  **entry point**, less headache, no decorator and only two classes :`JsonRpcApi`, `ExtendedJsonRpcApi`. This is also easy to implement.


Adding new methods which required authentication will be very simple. The following is a woking code which need to be refine. But you get an idea of what I am talking about

```
class JsonRpcApi:
    app = Klein()
    port = None

    def __init__(self, port, wallet=None):
        self.port = port
        self.wallet = wallet
        self.method_required_auth = ['sendtoaddress', 'sendfrom',  'sendmany']
     ....

    @app.route('/')
    @json_response
    @cors_header
    def home(self, request):
        # {"jsonrpc": "2.0", "id": 5, "method": "getblockcount", "params": []}
        # or multiple requests in 1 transaction
        # [{"jsonrpc": "2.0", "id": 1, "method": "getblock", "params": [10], {"jsonrpc": "2.0", "id": 2, "method": "getblock", "params": [10,1]}
        request_id = None

        try:

            content = json.loads(request.content.read().decode("utf-8"))
            auth = request.getHeader('Authorization')

            if content['method'] in self.method_required_auth:
                if auth and auth.split(' ')[0] == 'Basic':
                    decodeddata = base64.decodestring(auth.split(' ')[1].encode("utf-8"))
                    if decodeddata.decode("utf-8").split(':') == ['username', 'password']:

                        return self.get_data(content)
                else: 
                    request.setResponseCode(401)
                    request.setHeader('WWW-Authenticate', 'Basic realm="realmname"')

                    error = JsonRpcError.AuthError()
                    return self.get_custom_error_payload(request_id, error.code, error.message)

            if isinstance(content, list):
                result = []
                for body in content:
                    result.append(self.get_data(body))
                return result
    
            # otherwise it's a single request
            return self.get_data(content)
                    
                    
        except JSONDecodeError as e:
            error = JsonRpcError.parseError()
            return self.get_custom_error_payload(request_id, error.code, error.message)
...

class ExtendedJsonRpcApi(JsonRpcApi):
    """
    Extended JSON-RPC API Methods
    """

    def __init__(self, port, wallet=None):
        self.start_height = Blockchain.Default().Height
        self.start_dt = datetime.datetime.utcnow()
        super(ExtendedJsonRpcApi, self).__init__(port, wallet)

        self.method_required_auth.append('dumpprivkey')
    ....

```

jseagrave21:
@dauTT That is really cool! So you don't have to use Twisted at all?! Nice!

ixje:
your new solution sounds great!

dauTT:
@ixje ,

**could you please review the following code: [link](https://github.com/dauTT/neo-python/tree/feature-authentication)**?

The solution 5 (basic HTTP auth + backoff timer+one entry point) has been implemented 
i)  **without** the asynchronuos feature (commit a6de15a15faca2f3470a3231c51f8562b836c438) 
ii) **with** the asynchronuos feature (commit 05329adb18f672b9f423a5c8322992136ac96e27 ) 

**This is my assesment**: 

- solution 5 ii) is nice but will cause us to change all the test cases related to `json rpc` calls and will add the extra complexity to deal with `Deffered` objects. 

- solution 5 i) is easy to understand but we **can't keep the backoff timer feature** because when it kicks in, it will block all the incoming requests. 

Please let me know if you want to pursuit solution 5 ii) or 5 i)

ixje:
With the merging of #756 I'm going to close this and suggest to implement this as plugin in https://github.com/ixje/neopython-extended-rpc-server/ .

I'm still open to review your code. Just let me know.

jseagrave21:
@ixje @dauTT I'd like t propose opening a separate issue for implementing basic HTTP auth as I think this functionality is very important and useful for more than just `dumpprivkey`. Also, it seems as though @dauTT has already made significant progress and I'd like to encourage completion.

Issue #585, opened 2018-08-30
Implement NEP8


localhuman:
neo-python should be ready to support NEP8 as soon as it is on Testnet/Mainnet.  

Relevant changes in neo-vm here: https://github.com/neo-project/neo-vm/pull/39

ixje:
what is NEP8? I see many open PR's for NEPs that are not officially numbered yet, so I'm not sure which one is going to be NEP8.

localhuman:
Updated issue with changes.

deanpress:
@ixje  debugged my smart contract and noticed the C# compiler already utilizes NEP-8. So many contracts built in C# can not be invoked in neo-python atm.

ixje:
I'm going to look into this on Monday.

localhuman:
FWIW I don't believe NEP8 is on Mainnet CN yet.

ixje:
also requires https://github.com/neo-project/neo/commit/0de8b8867326405265e3daa24128269e33d914de

PR #635, opened 2018-09-25
Nep8

localhuman:

This addresses #585, and supersedes #603.  It adds in NEP8 functionality from here: https://github.com/CityOfZion/neo-boa/pull/94

Also makes usage of NEP8 compiler features configurable via `protocol.json` and from the prompt, with the default value being `False` until NEP8 is on mainnet.  From the prompt, you can use `config compiler-nep8 on|off`

- @ixje did the majority of the work in #603 

- tests

- no.  But thanks to @ixje for taking the initiative and getting this important feature implemented!

- [x] Did you add any tests?


neo/Settings.py
```diff
@@ -203,6 +205,9 @@ def setup(self, config_file):
         if 'ServiceEnabled' in config:
             self.SERVICE_ENABLED = bool(config['ServiceEnabled'])
 
+        if 'CompilerNep8' in config:
+            self.COMPILER_NEP_8 = bool(config['CompilerNep8'])
```
ixje: I'd like to suggest to at least use `config.get('CompilerNep8', False)` because the `AcceptIncomingPeers` update showed that quite some users just copy/replace their configs with their old one after they upgraded. Then this call will fail because the key is not there. Ideally we might want to have a utility function that tries to get the key, if not uses a specified default value but also warns the user that the key was not found and is now using the "x" as the default.

localhuman:
🥇

Issue #626, opened 2018-09-20
Add better Log message filtering


ixje:
_Rationale:_
Logging is a great way to get a quick idea what goes on under the hood when experiencing unexpected behaviour. This however assumes that log messages are descriptive and easy to consume. Currently messages are split into the default levels of INFO, DEBUG and ERROR. 

When enabling for example debug logging via the command `config debug on` you get DEBUG logs of the whole application. Several components can be very active (e.g. the network logging) and verbose causing you to miss the logs you're actually interested in. At this point it is an all or nothing setting and more control is desired.

_Goal:_
Get more control over the various levels of log information and filter on the desired messages. For example we'd like to split/filter messages on the various components such as; `network`,` VM` and `database` amongst others. We currently use the excellent [logzero](https://logzero.readthedocs.io/en/latest/#i-logzero-setup-logger) module which is build on top of the buildin `Logging` module. The latter supports [Filters](https://docs.python.org/3/library/logging.html#logging.Filter) which might be the solution here, but we're open to other approaches (a rationale is appreciated!)

PR #651, opened 2018-10-07
Feature log output control

ixje:
Addresses #626 

- Changed logging from a single logger instance for `neo-python` wide logging, to 1 logger per component (i.e. `network`, `db`, `vm` and a `generic` default for all others). 
- Introduced a `LogManager` class that manages these loggers and allows them to be configured individually i.e. set `network` to `INFO` level, while having `vm` on `DEBUG`. At the same time the setup allows configuring the `stdout` of the CLI to certain levels, while having file logging output at other levels (previously not possible as `logzero` uses the same levels for all).

write tests

Yes, `logzero` was mostly replaced by the `LogManager`. Note that I kept the `LogFormatter` from `logzero` so the look and feel is the same.

Why not use `logzero`? 
I actually intended to keep `logzero` but it has some unexpected defaults/behaviour as described [here](https://github.com/metachris/logzero/pull/61#discussion_r151943040). 

In summary, different handlers always have the same levels, `propagate` on loggers is set to `False` from the default `True`. Looking at the time period these PRs and some issues related on the repository have been open/not addressed, made me believe the intention is probably to keep the behaviour as is. For `neo-python` it did not match the intended features/requirements I had in mind, being:

- turn logging on or off for independent components 
- set independent logging levels for components
- set independent logging levels for stdout (`StreamHandler`) and other handlers like `RotatingFileHandler`

Finally, I'm aware that `neo-python-core` uses `logzero` at some places. These will not be filtered by the logmanager. I believe this is not an issue, because as some point these log messages should be either removed or replaced by exceptions such that they can be properly responded to.

- [X] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -3,10 +3,15 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
-
 [0.8.2] In Progress
 -------------------
 - Add log output control via the new ``config output_level`` command. The old ``config debug`` command is removed.
+- Add ExtendedJsonRpcApi, Add ``getnodestate`` RPC extended method, Add ``gettxhistory`` RPC extended method
+- Fix return types of ``claimGas`` function.
+- Update compiler version ``v0.5.4``
+- Adds ``sendmany`` feature to prompt.py, integrates with ``send`` feature, and adds provisions for sending with a negative fee and bad from_address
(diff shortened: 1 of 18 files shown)


ixje:
close in favour of #691 because this PR is stuck due to travis not recognising updates

PR #652, opened 2018-10-07
Fix a couple of tests 

ixje:
There are a couple of tests that didn't work as intended under certain conditions, had uncaught exceptions or otherwise. This cleans up a couple of them

1) This test was passing but internally had an exception that shows up in logs
https://github.com/CityOfZion/neo-python/blob/773e57e72b24a00f6d6d0f9b6fe701e3bb0e7d31/neo/Prompt/Commands/tests/test_send_command.py#L91

2) This test would cause a `404` error on downloading, while actually expected it kept triggering me to look into why. 
https://github.com/CityOfZion/neo-python/blob/master/neo/Prompt/Commands/tests/test_bootstrap.py#L24

3) This test would fail if your configured theme is `light` instead of the default `dark`
https://github.com/CityOfZion/neo-python/blob/773e57e72b24a00f6d6d0f9b6fe701e3bb0e7d31/neo/test_preferences.py#L22
4) one test used the deprecated `assertEquals` instead of `assertEqual` which kept throwing a warning
5) the `BlockchainFixtureTestCase` and `WalletFixtureTestCase` relied on running the tests from the root folder. My IDE (PyCharm) executes from whatever file I press `run test` and for tests using either of the before mentioned cases it would fail to find wallets due to the relative path construction e.g.
https://github.com/CityOfZion/neo-python/blob/773e57e72b24a00f6d6d0f9b6fe701e3bb0e7d31/neo/Utils/WalletFixtureTestCase.py#L14-L15

1. catch exception
2. don't assume default theme, but actually read what theme is configured
3. added a clear notification indicating the test shows an error
4. use `assertEqual`
5. detect root folder and create absolute paths.

`make test`

no

- [ ] Did you add any tests?

### DIFF
--- a/neo/Core/TX/test_transactions.py
+++ b/neo/Core/TX/test_transactions.py
@@ -27,7 +27,6 @@ class TransactionTestCase(NeoTestCase):
     tx_vout = []
 
     def test_tx_deserialize(self):
-
         ms = MemoryStream(self.tx_raw_hex)
 
         reader = BinaryReader(ms)
@@ -59,7 +58,6 @@ class TransactionTestCase(NeoTestCase):
     ctx_id = b'4feb0081f9425cab84269127bef0a871a84d4408f09923d17ebb257cd231b362'
 
     def test_contract_tx_deserialize(self):
-
         ms = MemoryStream(binascii.unhexlify(self.ctx_raw))
 
         reader = BinaryReader(ms)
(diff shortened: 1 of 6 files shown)


jseagrave21:
Good fixes. I also caught the Prompt.Commands.Send fix. It is included in https://github.com/CityOfZion/neo-python/pull/610

jseagrave21:
@ixje Hey, should there have been an update to the changelog?

ixje:
I initially chose not to do it because it was 2 minor fixes, then extended it with 2 more. I'm ok without a changelog on these. Most will not notice it as they've been there for such a long time.

PR #657, opened 2018-10-09
Fix Neo.Contract.GetIsPayable method not found error

jhwinter:
Corresponds to https://github.com/CityOfZion/neo-boa/pull/103
Changed [Neo.Contract.IsPayable](https://github.com/CityOfZion/neo-python/blob/773e57e72b24a00f6d6d0f9b6fe701e3bb0e7d31/neo/SmartContract/StateReader.py#L151) to `Neo.Contract.GetIsPayable`
Added tests in neo-boa and neo-python 
n/a

- [x] Did you add any tests?


ixje:
Hi, the correct name according to the [official C# implementation](https://github.com/neo-project/neo/blob/61db36a6294831b1cb67d7336c18b15eeb65749b/neo/SmartContract/NeoService.cs#L85) is `Neo.Contract.IsPayable`

ixje:
Will close for now to avoid accidental merging, but we can keep the discussion open. If you think my reference is wrong please let me know and we can re-assess.

PR #661, opened 2018-10-09
Add guideline for adding tests to the neo-privnet-unittest image

dauTT:

- Add option (`-u`) to run neo-python in combination with the `neo-privnet-unittest` image.
- Add guideline for adding unit-tests to the neo-privnet-unittest image.

manual testing.
running : `make docs`

This image [neo-privnet-unittest](https://hub.docker.com/r/dautt/neo-privnet-unittest/), need to be move under `COZ` domain. Once this is done we need to update the link in `tests.rst` file.

- [ ] Did you add any tests?


ixje:
Note to self: push docker image to coz account

docs/source/tests.rst
```diff
@@ -35,3 +35,119 @@ If you are adding tests or altering functionality, it might be faster to only ru
...
 
   python -m unittest neo/test_settings.py
+
+  
+
+Guideline
+-----
+
+The following guideline is relevant to all the ``unit-tests`` which are depending either on ``BlockChainFixtureTestCase`` or ``WalletFixtureTestCase``. 
+Such tests rely on fixtures which are produced by a ``neo-privnet-unittest`` image. The image is currently located here: `link <https://hub.docker.com/r/dautt/neo-privnet-unittest/>`_)
```
ixje: Please remove the `)` at the end of the line (as we have no matching `(`

docs/source/tests.rst
```diff
@@ -35,3 +35,119 @@ If you are adding tests or altering functionality, it might be faster to only ru
...
+.. code-block:: sh
+
+	neo-test-coz.wallet	(pwd = coz)							
+	neo-test1-bc.wallet	(pwd = 1234567890) 
+	neo-test2-bc.wallet	(pwd = 1234567890)                                                                                                                   	                  
+	neo-test1-w.wallet	(pwd = 1234567890)						
+	neo-test2-w.wallet	(pwd = 1234567890) 
+	neo-test3-w.wallet	(pwd = 1234567890)
+	
+(If you want to add new tests (BC) then use BC wallets. Otherwise use W wallets for new tests (W))
```
ixje: This line is kind of hard to understand. If you read it carefully it says:
If I want to add new tests, use BC wallets. (Otherwise == )If you _do not_  want to add new test, then use W wallets for new tests. 

I think we need to make it clearer when to choose which wallet. When you do so, can you please refer to the wallets as
> use the `-bc.wallet` wallets
> use the `-w.wallet` wallets

that makes it a bit more obvious for new readers that the BC/W abbreviation is part of the file name.

docs/source/tests.rst
```diff
@@ -35,3 +35,119 @@ If you are adding tests or altering functionality, it might be faster to only ru
 .. code-block:: sh
 
   python -m unittest neo/test_settings.py
+
+  
+
+Guideline
```
ixje: Please change the title to: `Fixtures Guidelines`
to make it more explicit what the next section is about.

dauTT:
Thanks @ixje  for reviewing the PR. I have implemented your feedback.

docs/source/tests.rst
ixje: You marked this as resolved but I don't think it is fully addressed. The text is nearly identical besides changing 
> then use BC wallets.

to

> then use the `-bc-wallet` wallets

That change is good but only 1 part of the request. The second part is that the description is still ambiguous and we need to make a clear description for users such that they how to decide between using `-bc.wallet` wallets and when to use `-w.wallet` wallets. 
Please also drop the `(BC)` and `(W)` as they have no added value when we already refer to `-bc.wallet` and `-w.wallet` (which have a more direct reference to the files). Thanks!

dauTT: Okay, I have changed the ambiguous sentence so that it is super clear. Let us hope it works fine now. :smile:

PR #664, opened 2018-10-13
Add documention for API servers and RPC methods

jseagrave21:
- Adds detailed documentation for API server operation and RPC methods
- Updates formatting and fixes typo

Trial and Error

`make docs`

No

- [ ] Did you add any tests?


docs/source/basicusage.rst
```diff
@@ -1,7 +1,10 @@
 Basic Usage
------------
+===========
 
-There are two main ways to use neo-python: ``np-prompt`` and running just the node with custom
+Introduction
+------------
+
+There are two main ways to interface with neo-python: ``np-prompt`` and running just the node with custom
```
ixje: ``np-prompt`` equals just running the node regardless of custom changes to the code. ``np-prompt`` is the same as executing `python neo/bin/prompt.py`. Having said that this addition is confusing to me.

docs/source/basicusage.rst
```diff
@@ -67,45 +71,55 @@ Example notifications plus help with all available arguments:
...
-    --wallet WALLET       Open wallet. Will allow you to use methods that
-                          require an open wallet
-    --host HOST           Hostname ( for example 127.0.0.1)
+      -h, --help            show this help message and exit
+      --datadir DATADIR     Absolute path to use for database directories
+      --maxpeers MAXPEERS   Max peers to use for P2P Joining
+      --wallet WALLET       Open wallet. Will allow you to use methods that
+                            require an open wallet
+      --host HOST           Hostname ( for example 127.0.0.1)
+      --extended-rpc        Use extended json-rpc api
```
ixje: This is not the most optimal place to comment, but it will also affect this document. 
> Use extended json-rpc api

this description is not self-explanatory, what does "extended" mean for somebody that just typed `np-prompt -h`. It would be great if we could update this to something like:
> Enable unofficial RPC API extensions

docs/source/basicusage.rst
```diff
@@ -67,45 +71,55 @@ Example notifications plus help with all available arguments:
...
+      --syslog              Log to syslog instead of to log file ('user' is the
+                            default facility)
+      --syslog-local [0-7]  Log to a local syslog facility instead of 'user'.
+                            Value must be between 0 and 7 (e.g. 0 for 'local0').
+      --disable-stderr      Disable stderr logger
+
+Running Extended Json-RPC Api Server
+""""""""""""""""""""""""""""""""""""
 
+Using the ``--extended-rpc`` arg will start your server with additional methods from ExtendedJsonRpcApi in addition to the normal JsonRpcApi methods. All original functionality is retained with the added functionality from ExtendedJsonRpcApi.
```
ixje: As pointed out in my previous comment on the console argument, if you've not read the `ExtendedJsonRpcApi` code then this will be a somewhat vague description. The second line also repeats what was already said in the first, just using different wording. 

I think it would help the reader if we point out that this flag enables additional unofficial RPC methods. Saying that will immediately make it clear
1) that the default methods are officially supported methods
2) that the extra methods are unofficial 
3) that they `additional` , thus retaining the other normal functionality. (somewhat implied already be `extended`)
4) why there is a separation.

docs/source/basicusage.rst
```diff
@@ -123,6 +137,7 @@ If you want an external program to access your API server, an open firewall port
      - 10332 
      - 20332
 
+For P2P and WebSocket information see `NEO-Python Seedlist <https://neo-python.readthedocs.io/en/latest/Seedlist.html>`_.
```
ixje: neo-python does not use or support websockets. Can you please remove mentioning `websockets` as it might raise questions or false expectations

jseagrave21:
Originally I was just going to write it for the extended methods but I figured it would be good to write a copy for the default methods also because we do not maintain parody at this moment with Neo CLI also our outputs look different in some cases like getaccountstate then Neo CLI and some of our input or allowable inputs are different than Neo CLI

docs/source/basicusage.rst
jseagrave21: I edited the original line: "There are two main ways to use neo-python: np-prompt and running just the node with custom code," because I thought 'use' was confusing and interface was a better term. I added an "Introduction" header to help with the formatting for the whole document. Could you give an example of how you would recommend using/interfacing with neo-python?

docs/source/basicusage.rst
jseagrave21: https://github.com/jseagrave21/neo-python/blob/96e70141d6b5c4d3961b120a499bb2686f96fa85/neo/bin/api_server.py#L132

docs/source/basicusage.rst
jseagrave21: I removed this section because updating the `help` for the `--extended-rpc` flag eliminates the confusion and the need for extra explanation I think.

docs/source/basicusage.rst
jseagrave21: done.

docs/source/basicusage.rst
ixje: i'll withdraw my comment. Having read the full `basicusage` file I see why this is described as `interfacing in 2 ways`.

ixje:
All good. Just waiting for a response from @localhuman @metachris on the documentation duplication https://github.com/CityOfZion/neo-python/pull/664#pullrequestreview-165225000

localhuman:
Looks good to me!

jseagrave21:
Now that https://github.com/CityOfZion/neo-python/pull/634 is merged I am going to add it to the list.

jseagrave21:
Okay, it should be good to go.

ixje:
@localhuman I'm cannot tell in what direction "looks good to me" points. 
Where I tagged you and Chris is with my idea: 
> I'm leaning towards referring to the offical NEO CLI documentation for the default RPC methods and only describing the extended ones in this documentation. 

but your response could equally mean that the commits by @jseagrave21 look good and they can go in, which would be the opposite. Can you be explicit in what you mean with "looks good to me"?

jseagrave21:
Since https://github.com/CityOfZion/neo-python/pull/714 was merged, I will update the example.

jseagrave21:
I also added REST server documentation because other than reading the code and querying a REST server, I haven't seen any documentation for it.

jseagrave21:
@localhuman @ixje I have updated the documentation for #712 and #719. Please  let me know what you think.

jseagrave21:
With `ExtendedJsonRpcApi` transitioned to plugins is this PR still salvageable? What documentation would we like to commit to maintaining?

ixje:
I think `ExtendedJsonRelated` documentation should be moved close to the source. So have a readme for new function and what output it produces together with e.g. the rpc-plugin. 

I'm still of the opinion that we should not duplicate documentation of the basic commands. NEO is maintaining it, we try to adhere to the same format, so linking to their docs is easier. Otherwise we're taking up a huge task were I'd rather use that time and energy to focus on improving other parts of `neo-python`.

jseagrave21:
@ixje would you like to keep the REST documentation or examples of inputs for JsonRpcApi? Perhaps some of the fixes for typos and then just delete the rest? I can pair it down.

PR #665, opened 2018-10-13
Add raw tx building examples

ixje:
A common question in the #python support channel is how to build raw transactions alá neon-js that can be send over RPC using the `sendrawtransaction` method.

Add a commented file to `\examples\` with 2 examples 1) building a ContractTransaction 2) building an InvocationTransaction

n/a

no

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.2] In Progress
 -------------------
+- Add raw transaction building examples in ``\examples\`` folder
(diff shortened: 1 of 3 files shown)


ixje:
@localhuman is setting the `Gas` attribute for InvocationTransactions mandatory?

localhuman:
I don't think so, unless it will cost more than the 10 free gas.

One comment, you can import `address_to_scripthash` from `neocore.bin.cli` I think rather than adding it in the example.

ixje:
That's what I thought (regarding the GAS question). As for the helper method, I was already surprised I couldn't find a utility function, didn't think of looking inside the neocore `cli`. I'll now stick with my current version as it also does the `UInt160` conversion I need for the later functions, but thanks for pointing it out 👍

PR #679, opened 2018-10-24
Fix getpeers RPC not matching maxpeers

### LINKED ISSUE #678: RPC getpeers return value exceeds maxpeers setting
### Current behavior
As reported here: https://github.com/CityOfZion/neo-python/issues/352#issuecomment-431966346
The `getpeers` command get return a list of peers that exceeds the `maxpeers` configuration setting. 

### Expected behavior
getpeers should not exceeds maxpeers.

### Your environment

Let us know in what environment you're running into the issue:

- OS: Debian 9.5.0 
- neo-python version: waiting for user input
- Python version: 3.7.0

ixje:
Fix #678 

if a peer disconnects or loses connection for whatever reason, then also remove it from the `ADDRS` list, on top of the `Peers` list.

wrote a test pre-fix, failed, fix, works

no

- [X] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.2] In Progress
 -------------------
+- Fix discrepancy between `getpeers` RPC call and the `maxpeers` setting
 - Update ``CreateAddress`` functionality and tests
 - Add VM sanity checks for operations on ``BigInteger``'s
 - Add raw transaction building examples in ``\examples\`` folder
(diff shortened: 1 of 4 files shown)


jseagrave21:
Good work! I was just sitting down to take a look at this lol.

ThomasLobker:
Unfortunately this introduces a new problem. Peers with lost connection are now not reconnecting anymore, as they would have when calling [PeerCheckLoop](https://github.com/CityOfZion/neo-python/blob/development/neo/Network/NodeLeader.py#L221). This is because PeerCheckLoop is comparing `self.Peers` with `self.ADDRS` and when the peer with the lost connection is removed from both, then it will not reconnect anymore. This is an issue when you are using a fixed set of peers. When the last peer disconnects, the synchonization stops completely.

ixje:
reverted in #693

PR #680, opened 2018-10-25
Update process_transaction

jseagrave21:
- Adds logic in case the tx fails to relay
- Maintains parody with `process_transaction` from `Send.py`
  - https://github.com/CityOfZion/neo-python/blob/fd5e1d928cd3860839f93b7a9c757779e1b82ada/neo/Prompt/Commands/Send.py#L206

**-update-**

- Maintains parody with neo-cli RPC method
- Ensures a tx sent via RPC is saved to the wallet

trial and error

unittest

no

- [x] Did you add any tests?


neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -555,7 +555,11 @@ def process_transaction(self, contract_tx, fee=None, address_from=None, change_a
...
         if context.Completed:
             tx.scripts = context.GetScripts()
-            NodeLeader.Instance().Relay(tx)
-            return tx.ToJson()
+            relayed = NodeLeader.Instance().Relay(tx)
+            if relayed:
+                self.wallet.SaveTransaction(tx)
+                return tx.ToJson()
+            else:
+                raise JsonRpcError(-32001, "Could not relay tx %s " % tx.Hash.ToString())
```
ixje: Unfortunately accepting this change would make the behaviour of calls like `sendtoaddress` deviate from the C# implementation, which is not what we want. See:
https://github.com/neo-project/neo/blob/dbe688b7b1d64975128b8aefb5c3ee692df07cd5/neo/Network/RPC/RpcServer.cs#L525-L532

jseagrave21:
@ixje I will try to suggest the change to C# but I may fail because I don't know C#. Perhaps a good compromise would be to just include `self.wallet.SaveTransaction(tx)` because I think this was mising from the original implementation https://github.com/neo-project/neo/blob/dbe688b7b1d64975128b8aefb5c3ee692df07cd5/neo/Network/RPC/RpcServer.cs#L528

jseagrave21:
@ixje I looked at the neo-project RPC and I just don't feel comfortable tinkering since I am not set up to test the code and I don't really know how to code in C#. I still think that this line https://github.com/CityOfZion/neo-python/pull/680/files#diff-f9e9d3e0fbf0fb287940a8d009a379b2R560 is equivalent to this line https://github.com/neo-project/neo/blob/dbe688b7b1d64975128b8aefb5c3ee692df07cd5/neo/Network/RPC/RpcServer.cs#L528 and should be added. What do you think?

jseagrave21:
Based on my reading of the purpose of `self.wallet.SaveTransaction(tx)` here 
https://github.com/CityOfZion/neo-python/blob/773e57e72b24a00f6d6d0f9b6fe701e3bb0e7d31/neo/Wallets/Wallet.py#L1102
I really think we should at least add this line.

ixje:
Sounds reasonable. I hope @localhuman can shed a quick light here because I'm far from familiar with that part of the `neo-python` code and I think he can probably do that in a minute or so

jseagrave21:
Based on our discussion, I will update the PR accordingly.

jseagrave21:
Okay @localhuman this is ready for your review. A brief recap: Originally, this PR was to bring the RPC method to parody with `process_transaction` from `Send.py`. However, this would take the RPC method out of parody with neo-cli. So, I have just updated `process_transaction` enough to bring it to parody with the neo-cli RPC method. Also, I  believe this is an important piece of code to include for anyone who is sending tx via RPC. cc @ixje

ixje:
👍

PR #681, opened 2018-10-25
Add transfertokens Extended RPC method

jseagrave21:
- Adds the `transfertokens` RPC method and tests
  - allows the user to easily transfer NEP5 tokens

Trial and Error

Unittest and personal testing on privatnet

No

- [x] Did you add any tests?


jseagrave21:
Another option I could use instead of adding a `verbose` input would be to make all the print statements into exceptions. Then these exceptions would actually be visible via RPC response, I think. This would mean updating the tests but that would be too hard.

What do you think @ixje @localhuman ?

ixje:
With the merging of #756 I'm going to close this and suggest to implement this as plugin in https://github.com/ixje/neopython-extended-rpc-server/ .

PR #685, opened 2018-10-26
Update Readme and Prompt.py help

jseagrave21:
- Update Readme documentation for Bootstrapping and wallet commands
  - The bootstrapping instructions update is specifically to address a user who was boostrapping for the first time and was unaware that the notifications db should also be bootstrapped.
- Update Prompt.py `help` to correctly show optional inputs

Trial and Error

Checked preview

No

- [ ] Did you add any tests?


neo/bin/prompt.py
```diff
@@ -141,8 +141,8 @@ class PromptInterface:
                 'wallet unspent (neo/gas)',
                 'wallet split {addr} {asset} {unspent index} {divide into number of vins}',
                 'wallet close',
-                'send {assetId or name} {address} {amount} (--from-addr={addr}) (--fee={priority_fee})',
-                'sendmany {number of outgoing tx} (--change-addr={addr}) (--from-addr={addr}) (--fee={priority_fee})',
+                'send {assetId or name} {address} {amount} (--from-addr={addr}) (--fee={priority_fee}) (--owners=[{addr}, {addr}, ...]) (--tx-attr=[{"usage": <value>,"data":"<remark>"},{"usage":<value>,"data":"<remark>"}, ...])',
```
ixje: can you please use 1 example input for `owners` and `tx-attr` i.e.`(--owners=[{addr}, ...])` 
I think that should be sufficient, the command is already really long so I'd like to try and keep it to a minimum

README.rst
```diff
@@ -285,37 +285,69 @@ Let's query for a block in the current server by hash or by block index:
...
+    wallet create_addr {number of addresses}
+    wallet delete_addr {addr}
+    
+    export wif {address}
+    import wif {wif}
+    
+    export nep2 {address}
+    import nep2 {nep2_encrypted_key}
+    
+    send {assetId or name} {address} {amount} (--from-addr={addr}) (--fee={priority_fee}) (--owners=[{addr}, {addr}, ...]) (--tx-attr=[{"usage": <value>,"data":"<remark>"},{"usage":<value>,"data":"<remark>"}, ...])
```
ixje: update to new output according to feedback below

PR #690, opened 2018-10-29
Add exception for ContractParameterType.Void input

jseagrave21:
- Adds exception for ContractParameterType.Void input based on my experience with `build`/`test` (shown below) and discussion with @ixje and @ThomasLobker 

```
[Param 2] Void input: 0xFF
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: 0xff
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: ff
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: void
Could not parse param as Void : Unknown param type Void
[Param 2] Void input:
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: None
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: []
Could not parse param as Void : Unknown param type Void
[Param 2] Void input: 0
Could not parse param as Void : Unknown param type Void
```
- And, improves `get_param` test coverage

Trial and Error

Unittest

No

- [x] Did you add any tests?


localhuman:
I'm not sure if this is the correct approach.  It would make more sense to disallow creating contracts that take void as an input parameter, since that makes no sense at all.

ixje:
@ThomasLobker that's what I would expect. As @localhuman pointed out (and I said the same on Discord) it doesn't make sense to have a void input parameter.

jseagrave21:
Okay, sounds good. I have the fix manually tested. Now, I am just writing tests for it.

jseagrave21:
@localhuman @ixje @ThomasLobker Okay, this PR is ready for another review.

neo/Prompt/Commands/LoadSmartContract.py
```diff
@@ -64,6 +64,10 @@ def LoadContract(args):
     path = args[0]
     params = parse_param(args[1], ignore_int=True, prefer_hex=False)
 
+    for i in str(params):
+        if i == "f":
```
ixje: I'm honestly surprised to see that this works. the `void` type is `0xFF` (or 255 in decimal) and not `f` (16 in decimal)

neo/Prompt/Commands/tests/test_loadsmartcontract.py
```diff
@@ -0,0 +1,99 @@
...
+        # test good contract
+        with mock.patch("builtins.open", mock.mock_open(read_data="path.avm")):
+
+            args = ["path.avm", "070710", "01", "True", "True", "True"]
+
+            res = LoadContract(args)
+
+        self.assertTrue(res)
+
+        # test bad contract
```
ixje: I believe we're testing for a "file not found" right? If so, can you please update the comment to clarify that. At first I was looking at the params wondering what's bad about the contract or its usage. Just a minor thing, but will help if we ever need to review it again.

CHANGELOG.rst
```diff
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.3-dev] in progress
 -----------------------
--
+- Remove ``ContractParameterType.Void`` (``ff``) as an optional input; improve ``get_param`` test coverage; improve ``loadcontract`` and ``gatherloadedcontractparams`` test coverage
```
ixje: Kind of similar remark as I made in the "TX mempool PR". I'll quote
> Can we make this message a "what was solved" instead of a "how" it was solved?

The message described here is really useful for the weekly contributions gists, but from a Changelog perspective not ideal. Imagine reading this line without knowing any context other than it being `neo-python` related, would you know what it does?

PR #699, opened 2018-11-01
Make ExtendedRPC interface plugin based only

ixje:
Currently unofficial RPC extensions are PR'ed to be added to the `ExtendedJsonRpcApi` class. This means they need to be reviewed and approved by those with review/merge rights on the `neo-python` repository, as well as eventually be maintained by the latter group in case an author disappears or becomes unavailable. This creates a maintenance problem. 

Because we consider the `ExtendedJsonRpcApi` as valuable to the community we changed the implementation to be plugin based. This has several benefits:
- `neo-python` maintainers do not become the maintainers of the user extensions.
- plugin authors have increased control and responsibility over their own extensions without being limited or delayed by `neo-python` maintainers.
- user extensions survive `neo-python` version updates

- existing tests still pass after porting them over.

- [X] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.3-dev] in progress
 -----------------------
--
+- Make the ExtendedRPC interface plugin based only.
 
 
 [0.8.2] 2018-10-31
--- a/neo/api/JSONRPC/ExtendedJsonRpcApi.py
+++ b/neo/api/JSONRPC/ExtendedJsonRpcApi.py
@@ -1,8 +1,38 @@
-from neo.Core.Blockchain import Blockchain
-from neo.api.JSONRPC.JsonRpcApi import JsonRpcApi, JsonRpcError
-from neo.Implementations.Wallets.peewee.UserWallet import UserWallet
-from neocore.UInt256 import UInt256
-import datetime
+import os
+import inspect
+import pkgutil
+from importlib import import_module
+from neo.api.JSONRPC.JsonRpcApi import JsonRpcApi
+from neo.api.JSONRPC.ExtendedRpcCommand import ExtendedRpcCommand
+import neo.api.JSONRPC.extended_commands as rpc_plugins
+
+plugins = []
+
+
+def iter_namespace(ns_pkg):
+    return pkgutil.iter_modules(ns_pkg.__path__, ns_pkg.__name__ + ".")
+
+
+# search in the plugins dir
+for (importer, _name, ispkg) in iter_namespace(rpc_plugins):
+    # only for plugins in packages
+    if ispkg:
+        # load the package
+        m = importer.find_module(_name).load_module(_name)
+        # and iterate over its contents
+        for (_, mod_name, _) in iter_namespace(m):
+            try:
+                imported_module = import_module(mod_name)
+
+                for i in dir(imported_module):
+                    attribute = getattr(imported_module, i)
+
+                    # to find a class that extends ExtendedRpcCommand
+                    if inspect.isclass(attribute) and issubclass(attribute, ExtendedRpcCommand):
+                        plugins.append(attribute)
+
+            except ImportError:
+                pass
 
 
 class ExtendedJsonRpcApi(JsonRpcApi):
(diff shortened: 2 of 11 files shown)


jseagrave21:
Will creating extended RPC methods still be something I can list as a contribution?

jseagrave21:
This looks great! One thing I haven't figured out how to do when contributing is create new folders. Since each new command will require a new folder, is there a way for me to create a new folder using GH online? Or could you recommend a different way?

dauTT:
Nice @ixje!,

It would be good idea to add your disclaimers to the python docs so that everyone is aware of the implication of using user extensions commands:

> - neo-python maintainers do not become the maintainers of the user extensions.
> - plugin authors have increased control and responsibility over their own extensions without being limited or delayed by neo-python maintainers.
> - user extensions survive neo-python version updates

jseagrave21:
Related to @dauTT 's [comment](https://github.com/CityOfZion/neo-python/pull/699#issuecomment-435199900), should there be instructions for contributors who would like to add new methods, specifically to include a label and their GH username so users can reach out to them if issues arise? For example, at the top of [cmd_getnodestate](https://github.com/CityOfZion/neo-python/pull/699/files#diff-6e3500a3ea7058c8a8305c3afdf93d8b) there would be a title
```
"""
command: "getnodestate"
use: Returns the status of the node
author: @jseagrave21 
implementation: see https://github.com/CityOfZion/neo-python/blob/e5c90e79f2f131259729d4ed2179f2b0ee0a4a93/docs/source/Api_Server.rst#getnodestate
"""

**Comment by localhuman** [comment]
Great work!  Just would like to provide a little feedback.

I think this plugin system is perhaps too specific in its implementation, in that you can only extend the RPC interface by writing extra commands.   My ideal system would resemble pretty closely how Django works with plugins, extensions and such, so that extended functionality is installed via pip or github and configured via a settings file.  

So a default settings file would look like: 
```
RPC_SERVER='neo.api.JSONRPC.JsonRpcApi.JsonRpcApi'

NOTIFICATION_SERVER='neo.api.REST.RestApi.RestApi'

PROMPT_COMMANDS = [
  'neo.Prompt.Commands.*',
  'my.Project.Namespace.Custom.*',
]

```

A custom JsonRPC settings file could look like that but have 
```
RPC_SERVER='my.project.JsonRPC.ExtendedJsonRpcApi'
```

Then in the `np-api-server` if started with RPC, the script would do a startup that looks like this: 
```
import importlib
rpc_module = importlib.import_module(settings.RPC_SERVER)
rpc_class = getattr(module, class_name)
rpc_server = rpc_class()
```

With this setup we can remove `ExtendedJsonRPCApi` from the neo-python repo, ( and perhaps remove a lot of current stuff in the repo ) into their own more manageable repos.

Then if someone wants the extended functionality, they can do `pip install neo-python-rpc-extended` or something which will install the proper package.  After that they can update their settings file and upon startup it will work.

This idea isn't fully flushed out, but there is a lot of thought behind how Django works, and it works pretty well.  I think an approach closer to this would provide the flexibility and extensibility we're looking for.

**Comment by ixje** [comment]
First off, thanks all for the input! This is valuable ❤️ 
@dauTT @jseagrave21 I agree more words need to be shared on how to set it up,  the implications and where to post bugs/suggestions. Will keep this in mind when updating. 

@localhuman 
**My initial thoughts on plugins**
I was/am of the opinion that different parts of the system have different interface needs and therefore can be addressed separately. Meaning the CLI and Network policies have different needs from the RPC server and can therefore have their own plugin interface (abstract base class) and plugin folder/loading. 

For what it's worth, the current approach still supports pip/github installation through [namespace-packages](https://setuptools.readthedocs.io/en/latest/pkg_resources.html?#namespace-package-support). If we continue in that direction I'll make it more obvious how to do that in a readme.

**The Django approach**
To be fair I'm not a Django expert, used it just a couple of times, had good experiences but not when it comes to extensions (as in; not working, hard to setup, error prone). I am however open minded when it comes to this and would like to discuss it a bit, as surely such a big project doesn't use it without good reasons. 

First thing that comes to mind though is that Django is far more mature and you probably set it up once and then don't need to update it for months before the next release. `neo-python` on the other hand arguably justifies a bi-weekly or monthly update at this point and then the following is one thing I dislike 
> After that they can update their settings file and upon startup it will work.

To me this reads as 

1. before each neo-python update you need to save your settings, then restore it when done updating. 

2. for each plugin you install you need to manually edit a/the global settings file. 

Is that a correct interpretation? if not please elaborate where my understanding deviates.

**The Django approach - example scenario**
If the above is correct, then if I wanted to install 2 RPC plugins from different authors I would have to do something like the following, right?
1. `pip install ExtendedJsonRpcApi`
2. `pip install rpc_plugin1`
3. `pip install rpc_plugin2`
4. update settings file to change the `RPC_SERVER` and add `RPC_COMMANDS` 
```
RPC_SERVER='neo.api.JSONRPC.JsonRpcApi.ExtendedJsonRpcApi'

RPC_COMMANDS = [
 'neo.bin.api.JSONRPC.default_cmds.*',
 'rpc_plugin1.cmds.*,
 'rpc_plugin2.cmds.*
]
```

My vision is; `pip install <plugin_name>` or "drag and drop" a plugin (if it's not publicly shared) and it just works from here on into eternity (sorta ;-)) without ever having to touch anything again. 

**in summary**
Again, I'm open to all sorts of feedback and discussion to get a good understanding of the different possibilities and requirements. For the current approach we could for example also just create a single `neo.Plugins` folder and categorise for different component plugins below that. Or do like flask and use a prefix `flask_<name>`.

localhuman:
Why would you need to install extra RPC plugins other than an Extended one?  The system works as currently without having to specify any commands.  If someone wants to create an Extended RPC Command Plugin based system, then they can make that and you can install as necessary.

What I would not like to do is support and need to provide interfaces that plugins need to adhere to and plugins can only be used if that hook exists.  I've already wasted many hours on the NEO C# plugin system trying to be able to do what I want to do and had my PR turned down because the only way you can 'plugin' to the system is via predefined 'plugin' paths and interfaces.

ixje:
From your reaction it seems like I misinterpreted this
> A custom JsonRPC settings file could look like that but have
> RPC_SERVER='my.project.JsonRPC.ExtendedJsonRpcApi'
> With this setup we can remove ExtendedJsonRPCApi from the neo-python repo

I'll drop this approach and have a look how to create something where plugins can plugin anywhere and do anything

Issue #703, opened 2018-11-03
Confirmed tx not purged from mempool


jseagrave21:
### Current behavior
My local mainnet RPC node shows one unconfirmed tx. When I query with `getrawmempool` I receive
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": [
        "0xf9916aa7159ff385b0b4e45686a4b7ccf3bcc934a60464b127a5148589aa3c18"
    ]
}
```

When I query that hash with `getrawtransaction` I receive
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xf9916aa7159ff385b0b4e45686a4b7ccf3bcc934a60464b127a5148589aa3c18",
        "size": 262,
        "type": "ContractTransaction",
        "version": 0,
        "attributes": [],
        "vout": [
            {
                "n": 0,
                "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                "value": "1",
                "address": "ANgA8bHVmvm6vvQJS24Fn9s5xegzzMviZp"
            },
            {
                "n": 1,
                "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                "value": "1",
                "address": "Ad7JugpBnsJnVPsG1Hvc7He65tQDNPzJYm"
            }
        ],
        "vin": [
            {
                "txid": "0x757f487d6fdece6b4a49b65ead8b1451cad7536ba504bbb8bd4b2de00ecda26c",
                "vout": 0
            }
        ],
        "sys_fee": "0",
        "net_fee": "0",
        "scripts": [
            {
                "invocation": "40568eb1c107d4a59e706a540e312125bca5d59afd6fa99ce0689e21311f844a481368fcdb1d0006a11156edffc203e18dffce112b9bf02074bba1572245c0ebae",
                "verification": "2103b6086a4347c8c49ee170f7631b86c9cf073c8b34574c703ea6e575f3744f318cac"
            }
        ],
        "blockhash": "0x7a610dff2a15047766e39b1e107a102c652f329dcc8c0983175278b2ef770ead",
        "confirmations": 7263,
        "blocktime": 1541076671
    }
}
```

### Expected behavior
[...]

PR #704, opened 2018-11-05
Add method to remove tx from mempool if found on the blockchain

jseagrave21:
- Resolves https://github.com/CityOfZion/neo-python/issues/703
- Adds `MempoolCheckLoop` and `RemoveTransaction` to remove any tx from the mempool found on the blockchain

Trial and Error

Unittest

No

- [x] Did you add any tests?


jseagrave21:
@localhuman This is implemented as we discussed in Slack.

neo/Network/NodeLeader.py
```diff
@@ -354,9 +354,47 @@ def AddTransaction(self, tx):
...
+        Checks the Mempool and removes any tx found on the Blockchain
+        Implemented to resolve https://github.com/CityOfZion/neo-python/issues/703
+        """
+        txs = []
+        keys = self.MemPool.keys()
+        for i in keys:
+            tx = self.GetTransaction(i)
+            txs.append(tx)
+
+        for tx in txs:
```
ixje: We can skip this
```python
        txs = []
        keys = self.MemPool.keys()
        for i in keys:
            tx = self.GetTransaction(i)
            txs.append(tx)

        for tx in txs:
```
and directly do
```python
for tx in self.MemPool.values():
```

neo/Network/test_node_leader.py
```diff
@@ -231,8 +235,45 @@ def test_inventory_received(self):
...
+
+        res = []
+        for tx in wallet.GetTransactions():
+                json = tx.ToJson()
+                res.append(json)
+
+        txid = res[0]['txid']
+        tx, height = Blockchain.Default().GetTransaction(txid[2:])
+
+        tx_bytes = tx.Hash.ToBytes()
```
ixje: I think we can skip the json conversion steps and simplify it to
```python
existing_tx = None
for tx in wallet.GetTransactions():
   existing_tx = tx
   break

self.assertNotEquals(None, existing_tx)
tx_bytes = tx.Hash.ToBytes()
```

neo/Network/test_node_leader.py
```diff
@@ -231,8 +235,45 @@ def test_inventory_received(self):
...
+
+    def test_mempool_check_loop(self):
+        # delete any tx in the mempool
+        txs = []
+        keys = NodeLeader.Instance().MemPool.keys()
+        for i in keys:
+            tx = NodeLeader.Instance().GetTransaction(i)
+            txs.append(tx)
+
+        for tx in txs:
```
ixje: same idea as previous comment. We can use `...MemPool.values()` instead of building a new list

neo/Network/NodeLeader.py
```diff
@@ -354,9 +354,47 @@ def AddTransaction(self, tx):
...
+            res = self.RemoveTransaction(tx)
+            if res:
+                logger.info("found tx 0x%s on the blockchain ...removed" % tx.Hash)
```
ixje: can you please change this to `debug` level. I don't want people to start screaming in #python that they keep seeing transactions being deleted from their blockchain. Oh and perhaps change the message that it's more obvious that the TX is removed from the mempool. If you'd see this line in some log file then at first sight you might actually think it's being removed from the blockchain (not that that's possible, but you get the idea)

CHANGELOG.rst
```diff
@@ -5,7 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.3-dev] in progress
 -----------------------
--
+- Add ``MempoolCheckLoop`` and ``RemoveTransaction`` to NodeLeader.py to resolve `#703 <https://github.com/CityOfZion/neo-python/issues/703>`_
```
ixje: Can we make this message a "what was solved" instead of a "how" it was solved? i.e.
> Fix confirmed TX not being purged from mempool 

That way somebody who reads the changelog doesn't have to click the issue link to know what was resolved. If you add the link to this PR, then they could lookup "how" it is resolved if they're interested.

neo/Network/test_node_leader.py
jseagrave21: Ran into this error as discussed in Slack.
```
======================================================================
ERROR: test_mempool_check_loop (neo.Network.test_node_leader.LeaderTestCase)
----------------------------------------------------------------------
  File "/mnt/c/users/jseag/neo/neo-python-dev/neo/Network/test_node_leader.py", line 212, in test_mempool_check_loop
    NodeLeader.Instance().MempoolCheckLoop()
  File "/mnt/c/users/jseag/neo/neo-python-dev/neo/Network/NodeLeader.py", line 389, in MempoolCheckLoop
    for tx in self.MemPool.values():
RuntimeError: dictionary changed size during iteration

----------------------------------------------------------------------
```
Decided to build a new list from `.values()`

PR #708, opened 2018-11-07
Silence - change logging levels on certain parts

ixje:
When people deploy faulty smart contracts on the blockchain or commit transactions that talk to them in the wrong way then errors can be thrown in the console. There's a good group of users that look at this log output and think their `neo-python` install is malfunctioning or broken. To list some examples:
> [W 181107 08:37:51 SmartContractEvent:278] Writing Payload value amount greater than ulong long is not allowed. Setting to ulong long max

> [E 181107 08:37:47 ApplicationEngine:233] NOT ENOUGH ITEMS TO CONCAT

> [I 181107 08:41:51 InteropService:190] Trying to get big integer Array [..omitted..]

> [I 181107 08:41:58 LevelDBBlockchain:383] Could not find transaction for hash b'insert hash'

I changed the log levels on the above and some others to `DEBUG` such that they don't show to the user by default. 

make test

no

- [ ] Did you add any tests?


localhuman:
Great addition!

Issue #713, opened 2018-11-08
getaccountstate returns different schema


xxedocxx:
### Current behavior
Balances returns in object.

### Expected behavior
It is supposed to return in array. Because neo-cli does.

**From Neo python**
balances:
{
 '0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7': '14.00251105',
 '0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b': '2.0' 
}
**From Neo cli**
balances: [ [Object], [Object] ] } }

### Your environment

Let us know in what environment you're running into the issue:

- OS: (Windows version/Linux distro/OSX version) 
- neo-python version: NEO-PYTHON:0.8.3-dev
- Python version:

dethos:
Since the objective of `neo-python` is supposed to be a full port of `neo-cli` and as stated on [previous comments ](https://github.com/CityOfZion/neo-python/pull/664#issuecomment-431571907) people should be able to "seamlessly switch" from one to another, it must receive the same inputs and return the same outputs, so I do think this is something that this should be fixed.

jseagrave21:
@dethos the only problem with your argument is that your example is for a RPC method which is planned to not be included in the default JsonRpcApi like it is in neo-cli (see @ixje 's review here https://github.com/CityOfZion/neo-python/pull/633#issuecomment-423770436)

In my opinion, neo-python is striving to be the best version of neo-cli. Just my opinion.

dethos:
The comment is on a topic related to `dumpprivkey`, but the observation about maintaining a similar interface to the user is general. Nevertheless lets wait to see what other people think.

xxedocxx:
It already has broke some dapps who uses the rpc api and relying on nodes from http://monitor.cityofzion.io/

jseagrave21:
@xxedocxx I didn't know that. Sounds like this fix is definitely necessary then. I guess in my perfect world I would have suggested neo-cli update their RPC method to match neo-python instead of the other way around lol.

ixje:
> Since the objective of `neo-python` is supposed to be a full port of `neo-cli` and as stated on [previous comments ](https://github.com/CityOfZion/neo-python/pull/664#issuecomment-431571907) people should be able to "seamlessly switch" from one to another, it must receive the same inputs and return the same outputs, so I do think this is something that this should be fixed.

[...]

PR #714, opened 2018-11-09
Fix response schema of the "getaccountstate" RPC method 

dethos:
This pull request tries to fix https://github.com/CityOfZion/neo-python/issues/713

Changed the JSON serialization of the AccountState instances

Tested manually and updated the automatic test

No, I do think I fixed it everywhere, but something might be missing.

- [ ] Did you add any tests?


ixje:
Thanks! 👍

PR #715, opened 2018-11-10
Added bad addresses to getpeers RPC endpoint

dethos:
This pull request completes the `getpeers` RPC functionality. At the time of the implementation, `neo-python` didn't track bad peers so it wasn't able to provide that information, since then that functionality was added so we must update the RPC method.

Added `bad` peers based on the `DEAD_ADDRS` list on the `NodeLeader`.

Updated the `getpeers` test

No

- [ ] Did you add any tests?


jseagrave21:
Nice!

ixje:
Thanks!

PR #712, opened 2018-11-08
Add GET & OPTIONS request functionality to JSON-RPC servers

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**
@xxedocxx (@Ed#2854 on Discord) brought up the issue that GET requests do not work and he DM'ed me saying he would like to see this functionality added. I have also noticed this and desired the same, so I could query RPC nodes from my phone's web browser.

@xxedocxx contributed the initial ability to parse GET request URI's into dictionary format using `furl`. It was great teamwork!

**NOTE:** this implementation requires that the "params" field not be last in GET requests (or there are parse errors)
**-edit-**
- this implementation does support having "params" as the last field

**NOTE:** also, GET requests do not support multiple requests in 1 transaction (tested; maintains parody with neo-cli)

**How did you solve this problem?**
trial, error, and teamwork

**How did you make sure your solution works?**
lots of personal testing and unittest

**Are there any special changes in the code that we should be aware of?**
Adds `furl` to requirements.tx

**Please check the following, if applicable:**

- [x] Did you add any tests?
- [x] Did you run `make lint`?
- [x] Did you run `make test`?
- [x] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

PR #716, opened 2018-11-10
Improve `parse_and_sign` and `ImportMultiSigContractAddr`

jseagrave21:
- Improves `parse_and_sign` and `ImportMultiSigContractAddr` functionality
- Add test coverage

**NOTE** Required an update to the test fixtures:

[notif_fixtures_v9.tar.gz](https://github.com/CityOfZion/neo-python/files/2568398/notif_fixtures_v9.tar.gz)
[fixtures_v9.tar.gz](https://github.com/CityOfZion/neo-python/files/2568399/fixtures_v9.tar.gz)

Trial and Error

Unittest
**NOTE:** Due to permission errors I was not able to run `make test` so I am not sure how the changes in the fixtures (mainly the small advancement in height) has affected other tests. I was thinking it might work after the fixtures are uploaded to AWS.

Updated the unittest fixtures

- [x] Did you add any tests?


neo/Prompt/Commands/LoadSmartContract.py
```diff
@@ -224,16 +224,10 @@ def ImportMultiSigContractAddr(wallet, args):
...
-
-        verification_contract = Contract.CreateMultiSigContract(pubkey_script_hash, int(m), publicKeys)
-
-        address = verification_contract.Address
-
-        wallet.AddContract(verification_contract)
-
-        print("Added multi-sig contract address %s to wallet" % address)
-        return address
+    pubkey_script_hash = Crypto.ToScriptHash(pubkey, unhex=True)
```
ixje: What would also improve this code is:
- catching the possible exception thrown by `Contract.CreateMultiSigcontract` when `m` != `len(publicKeys)`

neo/Prompt/Commands/Send.py
```diff
@@ -231,7 +231,7 @@ def parse_and_sign(wallet, jsn):
         context = ContractParametersContext.FromJson(jsn)
         if context is None:
             print("Failed to parse JSON")
-            return
+            return False
```
ixje: It is a bad practice to mix return types. Now we get `boolean` or `Transaction`.  I know this is already happening throughout the code, but let's not add more as it will lead to maintainability problems. You'll see that the general style used is: 
```
x = create_object()
if x :
   # do something
else:
   # could not create object
```
`None` will also test for `False` in your test cases so that shouldn't be a problem

neo/Prompt/Commands/LoadSmartContract.py
ixje: Hehe, this indeed solves it. I should have been more explicit because if you look in the code you'll see it does more than just that
https://github.com/CityOfZion/neo-python/blob/e002fa580388aedb6b83ef8fd8a4a2867b858813/neo/SmartContract/Contract.py#L91-L98

In stead of replicating that logic I'd rather see a: try/except around
`verification_contract = Contract.CreateMultiSigContract(pubkey_script_hash, int(m), publicKeys)`

While you're at it, can you please change the `raise Exception`'s to `raise ValueError`'s ? Thanks!

jseagrave21: I tried for quite awhile, but I couldn't figure out how to mock the `len(publicKeys) > 1024` test. I don't think it is worth creating a list with a length that is actually 1025. Let me know if you have any more recommendations.

ixje:
I've uploaded the files but it doesn't help. 
> [I 181115 11:46:03 BlockchainFixtureTestCase:70] downloading fixture notification database from https://s3.us-east-2.amazonaws.com/cityofzion/fixtures/notif_fixtures_v9.tar.gz. this may take a while
E[I 181115 11:46:04 LevelDBBlockchain:114] leveldb unavailable, you may already be running this process: b'IO error: lock /home/travis/.neopython/fixtures/test_chain/LOCK: already held by process' 

Initially I thought this was because you created the fixtures while the level DB was open, so I manually removed the `LOCK` file from the archives, but it still seems to fail. My only recommendation is to get it working locally with `make test` first. 99 out of 100 if it doesn't build locally it will also not build on github.

jseagrave21:
Okay, well I followed the documentation in `tests.py` to create the fixtures, which you can see weren't perfect to begin with. If there are other errors in this process that aren't obvious, then I don't know what they are. @dauTT could you provide some help perhaps?

dauTT:
@jseagrave21 ,
in regards to your note:

> Due to permission errors I was not able to run make test so I am not sure how the changes in the fixtures (mainly the small advancement in height) has affected other tests. I was thinking it might work after the fixtures are uploaded to AWS.

You can test everything locally before uploading the fixture to AWS. Both fixures files (notif_fixtures_v9.tar.gz, fixtures_v9.tar.gz) need to be localted here :
image

run: 
```
make test
```
to see if everything passes. If not, you will need to fix the test cases.

ixje:
I'm curious to learn why the wallet address of wallet-1 (in the `neo-boa` PR) needs updating. Any idea?  I mean as far as you explained your approach I can't see why that is needed as I expect only an additional multi-sig address and some new TX's related to it. So it worries me a bit as long as I don't understand the root cause for that.

Do these "wallet sync" warnings exist in the current build/other PRs?

jseagrave21:
Unless I am mistaken, the wallet address for wallet-1 in the neo-boa test was wrong. It doesn't match the address from the `WalletFixtureTestCase`

The only time I've seen the `wallet sync` warnings is with this new build.

ixje:
Ok, I now wonder if the address was wrong how there were no tests failing 🤔 

Regarding warnings; I think what probably needs to happen is opening all wallets, make sure they're all synced to the latest block in the chain, then close and save.

jseagrave21:
Yes, I checked that. They were all at 100% before I exited then stopped the docker container. Is there a step I missed?

jseagrave21:
@ixje It looks like you were right. I went and ran `np-prompt -u` without the container running and reopened all the wallets. Now I am not seeing any "wallet sync" warnings.

jseagrave21:
Should I add an update to the `tests.py` instructions mentioning this extra step? I am not sure how I would phrase it.

ixje:
yeah it would be great if it can be added as a step/reminder in these instructions: https://neo-python.readthedocs.io/en/latest/tests.html#fixtures-guidelines

jseagrave21:
This is ready for the new fixtures to be uploaded to AWS. Everything should work now.

ixje:
the fixtures contain `LOCK` files which will make LevelDB think the chain is already open

jseagrave21:
There are `LOCK` files in the current fixtures. And `make test` works. Will Travis still fail?

jseagrave21:
@localhuman @ixje This PR is ready for your review please note that the change here https://github.com/CityOfZion/neo-python/pull/712 was implemented to show neo-boa successfully pass `make test` with the current build. I will revert this change as soon as you have reviewed the PR and told me to.

jseagrave21:
I am planning on using these tests as the basis for the refactored tests for `parse_and_sign`; mentioned here: https://github.com/CityOfZion/neo-python/pull/726#issuecomment-441996896

jseagrave21:
I am planning on using these tests as the basis for the refactored tests for `parse_and_sign`; mentioned here: https://github.com/CityOfZion/neo-python/pull/726#issuecomment-441996896

jseagrave21:
@metachris @localhuman please review this PR. I just updated it and it is working as intended. Please note I have edited `.travis.yml` to look at my `patch-1` to show that the build is successful. Upon merging of my `neo-boa` PR, I will update `.travis.yml` to its original state.

This PR is necessary for establishing the new test fixtures so I can test the functionality I am working on for glaiming GAS from multisig addresses as mentioned by @ThomasLobker in # _neopython in Slack.

Issue #717, opened 2018-11-13
build_raw_transactions.py relies on an active blockchain node


jseagrave21:
### Current behavior
_NOTE: This issue is carried over from an unresolved issue from #dev-general on Discord_

User @amir reported receiving this error while trying to build a raw transaction per the example:
```
Traceback (most recent call last):
  File "./neophone.py", line 229, in <module>
    NeoPhone()
  File "./neophone.py", line 52, in init
    getattr(self, parsed_args.command)(leftover_args)
  File "./neophone.py", line 157, in transfer
    neolib.transfer()
  File "/Users/amirbaer/Home/code/curv/repos/research/neo/python/neolib.py", line 124, in transfer
    context = ContractParametersContext(contract_tx)
  File "/usr/local/lib/python3.6/site-packages/neo/SmartContract/ContractParameterContext.py", line 83, in init
    self.ScriptHashes = verifiable.GetScriptHashesForVerifying()
  File "/usr/local/lib/python3.6/site-packages/neo/Core/TX/Transaction.py", line 721, in GetScriptHashesForVerifying
    raise Exception("Invalid operation")
Exception: Invalid operation
```

After investigating, I found the problem here:
https://github.com/CityOfZion/neo-python/blob/f003f0a66cdd10cef91f30a6cda5549844ebd013/neo/Core/TX/Transaction.py#L719-L721

### Expected behavior

Building a raw transaction to relay via RPC method `sendrawtransaction` should not rely on an active blockchain instance.

### How to reproduce

Run `example1` or `example2` without a blockchain db activate in the environment.

### Your environment

- OS: Windows 10 Pro
- neo-python version: v0.8.3.dev0
- Python version: 3.7

jseagrave21:
This issue has been addressed! :tada:

PR #718, opened 2018-11-13
Allow a raw transaction to be built without an active blockchain db in the environment

jseagrave21:
- Addresses https://github.com/CityOfZion/neo-python/issues/717
- `GetScriptHashesForVerifying` now does not rely on an active blockchain if assets NEO or GAS are used

Trial and Error

Unittest

No

- [x] Did you add any tests?


jseagrave21:
@ixje @localhuman although `test_build_raw_transactions.py` is not included in `make test` it will pass the unittest if ran separately. I could not figure out how to include a test of `build_raw_transactions.py` without breaking other tests (as you can see from the commits 😄 ).

neo/Core/TX/Transaction.py
```diff
@@ -716,7 +716,9 @@ def GetScriptHashesForVerifying(self):
                     hashes.add(UInt160(data=attr.Data))
 
         for key, group in groupby(self.outputs, lambda p: p.AssetId):
-            asset = GetBlockchain().GetAssetState(key.ToBytes())
+            asset = Helper.StaticAssetState(key)
```
ixje: This will prevent any new assets registered to the blockchain to not function unless we update the `StaticAssetState()` helper function, which is not desirable.  What we can do is add an internal flag to the base TX class, which you set specifically and only when building raw transactions, and `GetScriptHashesForVerifying` can check the flag to determine if it will use the `StaticChecker` or the dynamic.

neo/Core/TX/test_transactions.py
```diff
@@ -203,3 +204,16 @@ def test_tx_big_remark(self):
             tx = Transaction.DeserializeFrom(reader)
 
             self.assertEqual(tx.Hash.ToString(), self.giant_tx_hash)
+
+    def test_GetScriptHashesForVerifying_invalid_operation(self):
+        with patch("neo.Core.Helper", return_value=None):
+            with patch("neo.Blockchain", return_value=None):
+                with self.assertRaises(Exception) as e:
```
ixje: the patching isn't needed, because in this test the `LevelDBBlockchain` isn't used, but the abstract `Blockchain` class, so `asset = GetBlockchain().GetAssetState(key.ToBytes())` points to 
https://github.com/CityOfZion/neo-python/blob/e002fa580388aedb6b83ef8fd8a4a2867b858813/neo/Core/Blockchain.py#L304-L306
which will always be `None` for
https://github.com/CityOfZion/neo-python/blob/e002fa580388aedb6b83ef8fd8a4a2867b858813/neo/Core/TX/Transaction.py#L719-L721

If the idea is to increase coverage, then why don't we provide an asset that really does not exist?

neo/Core/TX/Transaction.py
jseagrave21: Okay, that makes sense. However, wouldn't the StaticChecker also need to be updated if new assets were added? I am assuming the static checker would verify the input hash vs NEO/GAS.

neo/Core/TX/test_transactions.py
jseagrave21: Mostly because I thought mocking output was easier because I could start with a `tx` that was already built in the test. Also, originally, I was trying to include the test of `build_raw_transactions.py` so every part would have been covered except this. I will see what I can do.

neo/Core/TX/Transaction.py
ixje: Yes that's a good point. I would rather see it failing though in an isolated case (raw transactions) than in a generic case. I don't know the full extend of the code base of what else will break by doing this. To me and with the knowledge I currently have this sounds like the safest approach while maintaining backwards compatibility.

jseagrave21: I am still not sure I understand, but I am implementing a concept with a new flag. In my original method, if the input hash doesn't match NEO or GAS for the comparison, it will return `None` at which point the normal process will proceed.

ixje:
Nice! 🎉

PR #719, opened 2018-11-15
Initial Django inspired component loading

ixje:
A different approach for loading plugins as suggested in #699, Django inspired

add a class loader

manually run `api_server.py` with Rest and RPC servers

Yes; the `RPC_SERVER` and `REST_SERVER` variables are now stored at module level and not part of the `SettingsHolder` class. The implication is that all networks (main/test/priv) use the same settings. The alternative is moving it into the `SettingsHolder` which means we need to update all `protocol.xxx.json` files to include new keys to load from and otherwise have a sane default.

What approach would be preferred here (requesting extra opinions 🙏 ): @metachris @jseagrave21 @ThomasLobker @dethos 
choice 1: all networks have the same plugin loading settings
choice 2: all network can have individual plugin loading settings as defined in their `protocol.xxx.json`

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -13,7 +13,9 @@ All notable changes to this project are documented in this file.
 - Move some warnings and 'expected' errors to `DEBUG` level to avoid logging to console by default
 - Empty VerificationScripts for deployed contracts now work as intended
 - Fix RPC's ``getaccountstate`` response schema to match ``neo-cli`` `#714 <https://github.com/CityOfZion/neo-python/issues/714>`
+- Add fix to ensure tx is saved to wallet when sent using RPC
 - Add bad peers to the ``getpeers`` RPC method `#715 <https://github.com/CityOfZion/neo-python/pull/715>`
+
 
 [0.8.2] 2018-10-31
 -------------------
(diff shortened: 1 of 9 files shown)


ixje:
PS: Don't merge yet. If this was the intended idea then I'll also have to update the documentation to remove any `ExtendedJsonRpc` references and make some note that there will be/is a plugin for it

jseagrave21:
I may not fully understand the implications, but I think simpler is better. If we could make one change to update the mainnet, testnet, privnet settings between "default" and "extended", I think that is okay, since a user is usually only using one instance at time.

Either way, I think _good documentation_ will allow for a user to operate the servers in the new way regardless of the implementation.

localhuman:
Looks good!  I think the only problem is that a user who installed from pip could not specify which plugin to use without editing the installed settings file.

I think choice 2 would allow for that.  As long as we have sensible defaults for when plugins aren't specified in `protocol.xxx.json` we should be ok to allow for configuring plugins via that file.

ixje:
This should be it, I didn't see any documentation that referenced the `ExtendedJsonRpc` server.

jseagrave21:
> PS: Don't merge yet. If this was the intended idea then I'll also have to update the documentation to remove any `ExtendedJsonRpc` references and make some note that there will be/is a plugin for it

@ixje Are you adding instructions in this PR on how to use the new setup. Or could you implement those instructions with https://github.com/CityOfZion/neo-python/pull/664?

ixje:
For the time being I'm not adding the instructions with this PR. We should document this at a point though.

PR #724, opened 2018-11-20
Fix gzip failure in current implementation of ExtendedJsonRpcApi

jseagrave21:
Currently when a request is made with a header including support for gzip such as `Accept-Encoding: deflate, gzip;q=1.0, *;q=0.5`, this error occurs

```
======================================================================
ERROR: test_gzip_compression (neo.api.JSONRPC.test_extended_json_rpc_api.ExtendedJsonRpcApiTestCase)
----------------------------------------------------------------------
  File "/mnt/c/users/jseag/neo/neo-python-dev/neo/api/JSONRPC/test_extended_json_rpc_api.py", line 163, in test_gzip_compression
    res = self.app.home(mock_req)
  File "/mnt/c/users/jseag/neo/neo-python-dev/neo/api/utils.py", line 27, in wrapper
    response_data = gzip.compress(bytes(response_data, 'utf-8'), compresslevel=COMPRESS_FASTEST)
TypeError: encoding without a string argument

----------------------------------------------------------------------
```

This PR fixes this problem and maintains all other functionality.

Trial and Error

Unittest and personal testing 

No

- [x] Did you add any tests?


jseagrave21:
From my research, it seems as though functions with decorators are not meant to be modified or extended. To this end, I am proposing a new idea: adding an identifier input `server_type` which can be altered as desired. My intention is to be able to identify RPC server types with an `OPTIONS` request where it would otherwise need to be accomplish through trial and error.

jseagrave21:
This method allows for identification of server_type while maintaining all other functionality

neo/api/JSONRPC/JsonRpcApi.py
```diff
@@ -172,12 +172,17 @@ def home(self, request):
...
         elif "OPTIONS" == request.method.decode("utf-8"):
-            return {'supported HTTP methods': ("GET", "POST"),
-                    'JSON-RPC server type': "default"}
+            return self.options_response()
 
         error = JsonRpcError.invalidRequest("%s is not a supported HTTP method" % request.method.decode("utf-8"))
         return self.get_custom_error_payload(request_id, error.code, error.message)
 
+    @classmethod
+    def options_response(self):
```
LysanderGG: The argument should be `cls` rather than `self`

jseagrave21: Nice! Thanks

LysanderGG: Same in `neo/api/JSONRPC/ExtendedJsonRpcApi.py` 😅

jseagrave21: I wondered about that but does that mean test_json_rpc_api.py is wrong here https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/api/JSONRPC/test_json_rpc_api.py#L37?

LysanderGG: Technically the parameter can be named whatever you want, but it's a convention to name it `cls` for class methods as it is a convention to name the first parameter `self` for instance methods.
https://www.python.org/dev/peps/pep-0008/#function-and-method-arguments

jseagrave21:
@LysanderGG Okay, I think I got them all. Thank you for the review!

PR #725, opened 2018-11-27
Improve CommandBase.register_sub_command to use the subcommand's Comm…

LysanderGG:
Registering a `subcommand` and declaring the `subcommand`'s `CommandDesc` was repeating the `subcommand`'s `id`.

For example we had:
```
self.register_sub_command('create', CommandWalletCreate())
```
and 
```
def command_desc(self):
        p1 = ParameterDesc('path', 'path to store the wallet file')
        return CommandDesc('create', 'creates a new NEO wallet address', [p1])
```

The same `id` has to be specified twice which could lead to errors.

Use the `subcommand`'s `CommandDesc.command` directly when registering a subcommand.

I tested np-prompt.

No.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -69,15 +69,20 @@ class CommandBase(ABC):
     def execute_sub_command(self, id, arguments):
         self.__sub_commands[id].execute(arguments)
 
-    # ids: can be either a string or a list of strings.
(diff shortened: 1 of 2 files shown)


jseagrave21:
This looks great. I will update #726 as soon as it is merged.

ixje:
👍

PR #726, opened 2018-11-27
Add support for `wallet` `send`, `sendmany`, `sign`, `open`, and `close`

jseagrave21:
- supports https://github.com/CityOfZion/neo-python/issues/623
- adds support for `wallet` `send`, `sendmany`, `sign`, `open`, and `close`

Trial and Error

manual testing. I haven't added coverage yet.

**-update-**

I added coverage for `send` and `sendmany`. Let me know what you think.

No. But please check my implementation to see if I am on the right track.

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -67,7 +67,7 @@ class CommandBase(ABC):
 
     # Raise KeyError exception if the command does not exist
     def execute_sub_command(self, id, arguments):
-        self.__sub_commands[id].execute(arguments)
+        return self.__sub_commands[id].execute(arguments)
 
     def register_sub_command(self, sub_command, additional_ids=[]):
         """
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -66,18 +66,25 @@ class CommandBase(ABC):
         pass
 
     # Raise KeyError exception if the command does not exist
-    def execute_sub_command(self, id, arguments):
-        self.__sub_commands[id].execute(arguments)
-
-    # ids: can be either a string or a list of strings.
-    def register_sub_command(self, ids, sub_command):
-        if isinstance(ids, list):
-            for id in ids:
-                self.__register_sub_command(id, sub_command)
+    def execute_sub_command(self, id, arguments=None):
+        if arguments is not None:
+            self.__sub_commands[id].execute(arguments)
         else:
-            self.__register_sub_command(ids, sub_command)
+            self.__sub_commands[id].execute()
 
-    def __register_sub_command(self, id, sub_command):
+    def register_sub_command(self, sub_command, additional_ids=[]):
+        """
+        Register a command as a subcommand.
+        It will have it's CommandDesc.command string used as id. Additional ids can be provided.
+         Args:
+            sub_command (CommandBase): Subcommand to register.
+            additional_ids (List[str]): List of additional ids. Can be empty.
+        """
+        self.__register_sub_command(sub_command, sub_command.command_desc().command)
+        for id in additional_ids:
+            self.__register_sub_command(sub_command, id)
+
+    def __register_sub_command(self, sub_command, id):
         if id in self.__sub_commands:
             raise ValueError(f"{id} is already a subcommand of {self.command_desc().command}.")
         if sub_command.__parent_command and sub_command.__parent_command != self:
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -65,18 +65,15 @@ class CommandBase(ABC):
     def command_desc(self):
         pass
 
-    # Raise KeyError exception if the command does not exist
-    def execute_sub_command(self, id, arguments=None):
-        if arguments is not None:
-            self.__sub_commands[id].execute(arguments)
-        else:
-            self.__sub_commands[id].execute()
+    def execute_sub_command(self, id, arguments):
+        self.__sub_commands[id].execute(arguments)
 
     def register_sub_command(self, sub_command, additional_ids=[]):
         """
         Register a command as a subcommand.
         It will have it's CommandDesc.command string used as id. Additional ids can be provided.
-         Args:
+
+        Args:
             sub_command (CommandBase): Subcommand to register.
             additional_ids (List[str]): List of additional ids. Can be empty.
         """
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -65,8 +65,11 @@ class CommandBase(ABC):
     def command_desc(self):
         pass
 
-    def execute_sub_command(self, id, arguments):
-        self.__sub_commands[id].execute(arguments)
+    def execute_sub_command(self, id, arguments=None):
+        if arguments is not None:
+            self.__sub_commands[id].execute(arguments)
+        else:
+            self.__sub_commands[id].execute()
 
     def register_sub_command(self, sub_command, additional_ids=[]):
         """
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -66,11 +66,8 @@ class CommandBase(ABC):
         pass
 
     # Raise KeyError exception if the command does not exist
-    def execute_sub_command(self, id, arguments=None):
-        if arguments is not None:
-            self.__sub_commands[id].execute(arguments)
-        else:
-            self.__sub_commands[id].execute()
+    def execute_sub_command(self, id, arguments):
+        self.__sub_commands[id].execute(arguments)
 
     def register_sub_command(self, sub_command, additional_ids=[]):
         """
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -65,6 +65,7 @@ class CommandBase(ABC):
     def command_desc(self):
         pass
 
+    # Raise KeyError exception if the command does not exist
     def execute_sub_command(self, id, arguments=None):
         if arguments is not None:
             self.__sub_commands[id].execute(arguments)
--- a/neo/Prompt/Commands/Send.py
+++ b/neo/Prompt/Commands/Send.py
@@ -13,8 +13,67 @@ from neocore.Fixed8 import Fixed8
 import json
 from prompt_toolkit import prompt
 import traceback
+from neo.Prompt.PromptData import PromptData
+from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
 from logzero import logger
 
+
+class CommandWalletSend(CommandBase):
+
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        framework = construct_send_basic(PromptData.Wallet, arguments)
+        if type(framework) is list:
+            process_transaction(PromptData.Wallet, contract_tx=framework[0], scripthash_from=framework[1], fee=framework[2], owners=framework[3], user_tx_attributes=framework[4])
+
+    def command_desc(self):
+        p1 = ParameterDesc('assetId or name', 'the asset you wish to send')
+        p2 = ParameterDesc('address', 'the NEO address you will send to')
+        p3 = ParameterDesc('amount', 'the amount of the asset you wish to send')
+        p4 = ParameterDesc('--from-addr={addr}', 'specify the NEO address you wish to send from', optional=True)
+        p5 = ParameterDesc('--fee={priority_fee}', 'attach a fee to give your tx priority (> 0.001)', optional=True)
+        p6 = ParameterDesc('--owners=[{addr}, ...]', 'specify tx owners', optional=True)
+        p7 = ParameterDesc('--tx-attr=[{"usage": <value>,"data":"<remark>"}, ...]', 'specify unique tx attributes', optional=True)
+        params = [p1, p2, p3, p4, p5, p6, p7]
+        return CommandDesc('send', 'send an asset', params=params)
+
+class CommandWalletSendMany(CommandBase):
+
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        framework = construct_send_many(PromptData.Wallet, arguments)
+        if type(framework) is list:
+            process_transaction(PromptData.Wallet, contract_tx=framework[0], scripthash_from=framework[1], scripthash_change=framework[2], fee=framework[3], owners=framework[4], user_tx_attributes=framework[5])
+
+    def command_desc(self):
+        p1 = ParameterDesc('number of outgoing tx', 'the number of tx you wish to send')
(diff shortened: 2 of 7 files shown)


neo/Prompt/Commands/Wallet.py
```diff
@@ -41,11 +47,19 @@ def execute(self, arguments):
         wallet = PromptData.Wallet
         item = get_arg(arguments)
 
-        # Create and Open must be handled specially.
+        # Create, Open, and Close must be handled specially.
         if item == 'create':
             self.execute_sub_command(item, arguments[1:])
             return
 
+        elif item == 'open':
```
LysanderGG: We could regroup all the commands in a `set` and check if `item in xxx`. (if `close` can be moved out, see after)

neo/Prompt/Commands/Wallet.py
```diff
@@ -41,11 +47,19 @@ def execute(self, arguments):
...
+            self.execute_sub_command(item, arguments[1:])
+            return
+
+        elif item == 'close':
```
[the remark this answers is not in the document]
jseagrave21: I don't think so because this follows the original implementation (i.e. `wallet close` closes the open wallet)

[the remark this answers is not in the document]
LysanderGG: You added `close` as a "specially handled command" instead of a "normal command".

So for now, there is 
```
elif item == 'close':
    self.execute_sub_command(item)
    return
```
before
```
if not wallet:
    print("Please open a wallet")
    return
```

If you do not handle `close` a special way, it would be handled in 
```
try:
    self.execute_sub_command(item, arguments[1:])
except KeyError:
    print(f"Wallet: {item} is an invalid parameter")
```

Therefore we would have the 2 cases:
- a wallet is opened: works as now
- no wallet is opened: `print("Please open a wallet")` would be executed instead of nothing.

The complete implementation of `execute()` could be:
```
def execute(self, arguments):
    wallet = PromptData.Wallet
    item = get_arg(arguments)

    # Create, Open, and Close must be handled specially.
    if item in {'create', 'open'}:
        self.execute_sub_command(item, arguments[1:])
        return

    if not wallet:
        print("Please open a wallet")
        return

    if not item:
        print("Wallet %s " % json.dumps(wallet.ToJson(), indent=4))
        return

    try:
        self.execute_sub_command(item, arguments[1:])
    except KeyError:
        print(f"Wallet: {item} is an invalid parameter")
```

If it is still not clear (I don't find a good way to explain), let's talk on Discord later on.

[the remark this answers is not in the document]
jseagrave21: Okay, I think I understand. You are saying to just keep `close` as a normal subcommand. Okay. Lemme see what I can do.

jseagrave21:
What do you guys think of the method of testing I implemented? I could continue this method of testing for the other methods introduced to ensure coverage is maintained as much as possible while refactoring is completed.

neo/Prompt/Commands/Wallet.py
```diff
@@ -109,12 +120,66 @@ def command_desc(self):
...
+        return CommandDesc('open', 'opens a NEO wallet', [p1])
+
+
+class CommandWalletClose(CommandBase):
+
+    def __init__(self):
+        super().__init__()
+
+    def execute(self):
+        if PromptData.Wallet:
```
LysanderGG: Maybe the implementation could be moved to a static method or to PromptData.
It could avoid having to instanciate CommandWalletClose to call execute. (`CommandWalletClose().execute()` in prompt.py).

What do you think?

neo/Prompt/Commands/Wallet.py
```diff
@@ -55,7 +61,10 @@ def execute(self, arguments):
             return
 
         try:
-            self.execute_sub_command(item, arguments[1:])
```
LysanderGG: In my opinion, the original signature `execute(self, arguments)` should be respected. None could be given as arguments for commands like `CommandWalletClose`.

neo/Prompt/CommandBase.py
```diff
@@ -66,8 +66,11 @@ def command_desc(self):
         pass
 
     # Raise KeyError exception if the command does not exist
-    def execute_sub_command(self, id, arguments):
-        self.__sub_commands[id].execute(arguments)
+    def execute_sub_command(self, id, arguments=None):
+        if arguments is not None:
```
LysanderGG: In my opinion, the original signature `execute(self, arguments)` should be respected. None could be given as arguments for commands like `CommandWalletClose`.

neo/Prompt/Commands/tests/test_send_command.py
```diff
@@ -328,25 +326,15 @@ def test_could_not_send(self, mocked_tracback_module):
...
+                    command_wallet = Wallet.CommandWallet()
+                    command_wallet.execute(args)
 
-            # check for 2 transfers
-            transfers = 0
-            for info in json_res['vout']:
-                if info['address'] == self.watch_addr_str:
-                    transfers += 1
-            self.assertEqual(2, transfers)
+                    self.assertIn("Relayed Tx", mock_print.getvalue())
```
ixje: the `execute()` method does not have an enforced return type. Use this to your advantage to return the normal `res` value from `process_transaction` as in the original tests
[here](https://github.com/CityOfZion/neo-python/pull/726/files?utf8=%E2%9C%93&diff=split#diff-d795c43528b80a2b124aeb31f31808afR51)

this way you don't have to patch `stdout` and you have far more options for testing correct behaviour. We now threw away good logic in favour of trusting a printed message indicating all went right. 

Note that the actual goal isn't coverage. The goal is testing correct behaviour of the functions/modules under different circumstances. `coverage` is just an indicator that can tell you if you missed any obvious parts in your test scenario's

neo/Prompt/Commands/Wallet.py
jseagrave21: This is why I originally had it as a class method.  But changed it after your review. Couldn't we just keep it as a class method?

neo/Prompt/Commands/Wallet.py
jseagrave21: Okay, I didn't know of that was preferred. I had that originally but I thought it looked kinda silly.

neo/Prompt/Commands/tests/test_send_command.py
jseagrave21: I tried this originally and all it ever returned was `None`. Also `pylint` warned about it in VSC. Do you have any hints?

This is why I moved the test logic to the next test. So, this basic test shows everything works and then I test the original functions thoroughly because they return values other than `None`

neo/Prompt/Commands/Wallet.py
jseagrave21: I think it was just the way I originally implemented it.

neo/Prompt/Commands/Wallet.py
jseagrave21: A static method still required  me to instantiate the class

neo/Prompt/Commands/Wallet.py
```diff
@@ -105,12 +120,66 @@ def execute(cls, arguments):
...
+        p1 = ParameterDesc('path', 'path to open the wallet file')
+        return CommandDesc('open', 'opens a NEO wallet', [p1])
+
+
+class CommandWalletClose(CommandBase):
+
+    def __init__(self):
+        super().__init__()
+
+    @classmethod
```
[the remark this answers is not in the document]
jseagrave21: See https://github.com/CityOfZion/neo-python/pull/726/files#r236708544

neo/Prompt/Commands/tests/test_send_command.py
ixje: @LysanderGG if you read my comment above you see I assumed we could use the return value of `execute()`. Atm this doesn't work because for the `CommandBase.execute_sub_command` doesn't return the value. Do you see/have any objections for changing the behaviour of that? I don't have a full clear mind at this point to see possible limitations and you know this code the best :)

If we return the results of `execute` commands they become more obvious to test. The other approach is just directly calling a the command class. An example test of the former is:
```
                    args = ['send', 'neo', self.watch_addr_str, '50']                    
                    command_wallet = Wallet.CommandWallet()
                    res = command_wallet.execute(args)
``` 
an example of the latter is
```
                    args = ['neo', self.watch_addr_str, '50']                    
                    command_wallet_send = Wallet.CommandWalletSend()
                    res = command_wallet_send.execute(args)
```

jseagrave21:
@LysanderGG I have implemented the ability to return data which is very useful for testing. Additionally, I have refactored `test_send_command.py`. Also, I introduced what I think would be good return values for your wallet commands. I am open to your suggestions of course. Also, will you be creating the testing for the wallet commands you refactored?

jseagrave21:
@LysanderGG I have implemented the ability to return data which is very useful for testing. see https://github.com/CityOfZion/neo-python/pull/726#discussion_r236785324
Additionally, I have refactored `test_send_command.py` and added tests for `wallet open` and `wallet close` to `test_wallet_commands.py`. Also, I introduced what I think would be good return values for your wallet commands. I am open to your suggestions of course. Also, will you be creating the testing for the wallet commands you refactored?

neo/Prompt/Commands/Wallet.py
LysanderGG: I realize that my message was probably confusing, sorry for that.

I was suggesting to keep execute as is (to override `execute(self arguments)`) and have it call another static method.
Imo a `@classmethod` should only be used when you want to override it in a child class or when you need to have access to `cls`.

I think the following should work:

```python
class CommandWalletClose(CommandBase):
...
def execute(self, arguments=None):
    CommandWalletClose.close_wallet()


@staticmethod
def close_wallet():
    if not PromptData.Wallet:
        return False

    path = PromptData.Wallet._path
    PromptData.Prompt.stop_wallet_loop()
    PromptData.Wallet.Close()
    PromptData.Wallet = None
    print("Closed wallet %s" % path)
    return True
```

It should be possible to call it with `CommandWalletClose.close_wallet()`.
I think it would be even better to have this function in `PromptData` and call `PromptData.close_wallet()`.

LysanderGG:
@jseagrave21 Thank you!
I added one comment about `CommandWalletClose`, the rest looks gtm!

> Also, will you be creating the testing for the wallet commands you refactored?

Yes I will do it. 
Thank you for clearing out the path for tests!

neo/Prompt/Commands/Wallet.py
jseagrave21: @LysanderGG Thank you. I will give it a shot. I prefer this way because it makes sense to me where it is located. I wouldn't intuitively think to look for the function called by `wallet close` in `PromptData`. But, that is my opinion.

jseagrave21:
@LysanderGG @ixje Okay, this PR is ready for a [final] review 😉

neo/Prompt/Commands/Send.py
```diff
@@ -13,16 +13,77 @@
...
+
+
 def construct_send_basic(wallet, arguments):
-    if not wallet:
-        print("please open a wallet")
-        return False
     if len(arguments) < 3:
         print("Not enough arguments")
-        return False
+        return
```
LysanderGG: Because `construct_send_basic` can return `[contract_tx, scripthash_from, fee, owners, user_tx_attributes]` it should return `None` or `False` in early returns.

neo/Prompt/Commands/Send.py
```diff
@@ -35,60 +96,58 @@ def construct_send_basic(wallet, arguments):
...
 
 
 def construct_send_many(wallet, arguments):
-    if not wallet:
-        print("please open a wallet")
-        return False
     if len(arguments) is 0:
         print("Not enough arguments")
-        return False
+        return
```
LysanderGG: Should return something here too.

neo/Prompt/Commands/Wallet.py
```diff
@@ -100,22 +109,76 @@ def execute(self, arguments):
...
+        if PromptData.Wallet:
+            PromptData.close_wallet()
+
+        path = get_arg(arguments, 0)
+
+        if path:
+
+            if not os.path.exists(path):
+                print("Wallet file not found")
+                return
```
LysanderGG: Should return something here too.

neo/Prompt/Commands/Send.py
jseagrave21: @LysanderGG 
@ixje has explained to me that it is undesirable to have a function return both a boolean and say a function. Instead if you just leave return blank, it returns none, which will also be interpreted as False, so all my tests pass.

neo/Prompt/Commands/Send.py
jseagrave21: https://github.com/CityOfZion/neo-python/pull/726#discussion_r236978513

neo/Prompt/Commands/Wallet.py
jseagrave21: https://github.com/CityOfZion/neo-python/pull/726#discussion_r236978513

neo/Prompt/Commands/Send.py
LysanderGG: I agree for `boolean` and something else (I suggested `False` because it was this way before).
I think it should return `None` explicitly then.

When I see `return` (nothing) I assume that the function doesn't return anything. It can be misleading imo.

@ixje any thoughts?

ixje: I'm indifferent when it comes to `return` or `return None`. I guess I'm used to the idea that `return` always returns `None` under the hood, so to speak. Might as well make it explicit if that clarifies it for all future readers. I'll probably need to slapping on the fingers to remind me of doing that though ;)

ixje: The add a few more words regarding return values; at some point I expect to start applying gradual typing. We've had past cases where functions were returning `None`, `bool` or a `tuple`. That's just not workable.

LysanderGG: Thanks for the answer, I won't make this as a blocker for this PR, so @jseagrave21 as you want :smile:

jseagrave21: I would like to leave it this way so the format is consistent with #716. Consistency is what matters most to me so people will be able to follow our example throughout the code.

ixje:
@jseagrave21 don't worry, I'm going to merge in a minute and will then update these return parts as I want to update a couple of descriptions. We learn as we go, and while reviewing and testing your PR I noticed the following below

pre
The readability isn't optimal, so we need to present it differently. I tried out some formatting and came to this. 
post

Nothing you could necessarily do about upfront and no need for you to update this as I already have the code. Just take note of how the commands are labelled as well as how their descriptions look for future commits.

ixje:
@LysanderGG no worries. I appreciate a critical look. Keeps us all sharp and in the end improves overall quality.

PR #728, opened 2018-11-27
Removed some white space and hash

WyattMufson:
Super minor, was reading through the code and saw some added white space and what looked like part of a comment that had been forgotten to be erased.


ixje:
👍

PR #730, opened 2018-11-28
Prompt cleanup

ixje:
- clean up command descriptions
- apply alphabetic ordering of help (sub)command(s). Note: this does not apply to parameters. Just (sub)commands.
- remove colour from help. We had all help except the root level in black. Now also root level
- clean up `prompt.py` by removing unused imports and functions that have been or are about to be migrated to their own commands.

mostly see above

manual observation

@LysanderGG FYI; I had to removed your set comprehension for help printing because it would throw away my key sorting. 

- [ ] Did you add any tests?


jseagrave21:
This looks great. The only thing I can see is that during the refactoring process, it was nice to have the old help and command structure available as a ready-reference. However, we can use the `development` branch online just as easily.

neo/Prompt/Commands/Send.py
```diff
@@ -31,15 +31,19 @@ def execute(self, arguments):
...
-        p3 = ParameterDesc('amount', 'the amount of the asset you wish to send')
-        p4 = ParameterDesc('--from-addr={addr}', 'specify the NEO address you wish to send from', optional=True)
-        p5 = ParameterDesc('--fee={priority_fee}', 'attach a fee to give your tx priority (> 0.001)', optional=True)
-        p6 = ParameterDesc('--owners=[{addr}, ...]', 'specify tx owners', optional=True)
-        p7 = ParameterDesc('--tx-attr=[{"usage": <value>,"data":"<remark>"}, ...]', 'specify unique tx attributes', optional=True)
+        p1 = ParameterDesc('assetId or name', 'the asset (NEO/GAS) to send')
+        p2 = ParameterDesc('address', 'the destination address')
+        p3 = ParameterDesc('amount', 'the amount of the asset to send')
+        p4 = ParameterDesc('--from-addr', 'source address to take funds from (if not specified, take first address in wallet)', optional=True)
+        p5 = ParameterDesc('--fee', 'a fee to give your transaction priority (> 0.001) i.e. --fee=0.01', optional=True)
```
LysanderGG: It should be `e.g.` rather than `i.e.`

neo/Prompt/CommandBase.py
```diff
@@ -56,6 +56,7 @@ def __init__(self):
         super().__init__()
         self.__sub_commands = dict()
         self.__parent_command = None
+        self.__additional_ids = []
```
LysanderGG: A set would allow for faster lookup, and we don't need ordered collection here.

neo/Prompt/Commands/Send.py
jseagrave21: Agreed. 
e.g. == for example
i.e. == in other words (more like this is how it should look, not an example)

PR #731, opened 2018-11-29
[refactor-prompt] load cli commands dynamically (enables plugin support)

ixje:
Adds the feature of dynamically loading prompt commands. Also supports loading cmd plugins installed via `pip`.

Load plugin directory or via a `PROMPT_COMMANDS` list in `settings.py`. 

In this list you either specify a package to search supported modules in e.g. `neo.Prompt.Commands.*`
or you specify the module directly e.g. `neo.Prompt.Commands.Wallet`

It will search inside the module(s) for anything that inherits from `CommandBase` and has the `_isGroupBaseCommand` instance attribute set to `True`. The latter was needed because we need to be able to differentiate between Group Commands and subcommands, but they now all inherit from `BaseCommand`. The alternative to this would be to have group commands (e.g. `wallet`) inherit from a secondary class (e.g. `class GroupCommand: pass`) just for the sake of identification. 

write tests

Yes, root commands (e.g. 'wallet') need to set `_isGroupBaseCommand` to `True` in its constructor or it won't be picked up.

- [X] Did you add any tests?


ixje:
I'm looking for first feedback, specifically on the `_isGroupBaseCommand` approach. 

A limitation we currently have is that `pip install`'ed commands can only register themselves as a root level command (for a lack of a better word). The next step would be to allow commands to be registered as a subcommand (e.g. to get `wallet my_pip_plugin_command`)

We could for example do this by adding the following instance attributes:
```
_isSubCommandPlugin = True
_SubCommandParent = CommandWallet
```
Using this information we can then cycle through all commands until we find `CommandWallet` and then call `register_sub_command()` with the instance of the plugin command

ixje:
I'm considering to support loading plugins that are registered via `setuptools` under a specific entry point (e.g. `neopython.prompt.cmds`)

Would it be confusing if we on the one hand have the blow in settings; being a list of paths where it loads command modules from. 
```
PROMPT_COMMANDS = [
    'neo.Prompt.Commands.*'
]
```
while also under the hood loading anything registered via the entry point`neopython.prompt.cmds`? 

An example of this can be a plugin author that releases a package which you install like `pip install neopython-cli-mysupercmd`. Instead of heaving to `pip install` _and_ edit `PROMPT_COMMANDS` you would then only have to `pip install` and restart `neo-python`.

jseagrave21:
From my perspective, the fewer steps, the better. I am not sure who we expect our user base to be, but I tend to think I am the lowest on the totem pole, where I literally just blindly followed the Readme instructions to get me off the ground when I first started. Actually, some of the first PRs I made were to the Readme because it wasn't simple enough for me to complete without asking for help. So, I think whatever we do, make it consistent and consider that people may be using `neo-python` because it is supposed to be (and is, I think) easier than `neo-cli` (C#) but that may not mean that they have very much Python (or Pypi, or...) experience at all.

LysanderGG:
> I'm looking for first feedback, specifically on the _isGroupBaseCommand approach.

I don't have better ideas, other than the `RootCommand` class inheriting `BaseCommand` (and defining nothing special) you already suggested. Maybe it would lead to a litlle less code.


> _isSubCommandPlugin = True
> _SubCommandParent = CommandWallet

About `_isSubCommandPlugin`, I think this is unnecessary because it would prevent to use the same way to define SubCommands locally (in the repo code).

If there is only `_SubCommandParent`, all commands could use this approach and we could get rid of the  calls to `register_sub_command()` in root commands.

Do you think it would be harder to understand the parent-child relationship between commands?
Is there another use to `_isSubCommandPlugin` I haven't thought of?

ixje:
Good point `_isSubCommandPlugin = True` isn't necessary if we can just directly look for `_SubCommandParent` not being `None`. I just need some attribute I can `getattr` on while loading modules so I can identify what to do with it.

PR #732, opened 2018-12-02
[refactor-prompt] Prompt fixes

LysanderGG:
1. `help`, `quit` and `exit` being handled outside of `CommandBase` feature, the words were unknown from the autocompleter.

2. `CommandWalletCreateAddress.command_desc` was missing a aparameter.

1. Add these words to the autocompletion words list.

2. Added the parameter

Manual test.

No.

- [ ] Did you add any tests?


jseagrave21:
Is it worth creating tests for `help` commands to ensure mistakes like missing `params` doesn't happen?
The only problem I see with this idea is that the tests for higher level `help` commands would evolve as we completed more refactoring.

ixje:
👍

PR #733, opened 2018-12-02
Add missing tests for wallet create, verbose and create_addr

LysanderGG:
Add missing tests after PR #720 (also see https://github.com/CityOfZion/neo-python/pull/726#issuecomment-442285227)

The tests are for the following commands `wallet create`, `wallet verbose`, `wallet create_addr`.

Added some tests.

Run `make test`.

No.

- [x] Did you add any tests?

Note:

- I couldn't find any other test related to wallet migration so I was unable to test the command `wallet migrate`.


jseagrave21:
This looks good. Could you also add a test for `wallet`?

jseagrave21:
Also, it would be nice to see a test for an invalid `wallet` command

jseagrave21:
And, I was looking here: https://coveralls.io/builds/20405190/source?filename=neo/Prompt/Commands/Wallet.py#L109
Is the logic here wrong? It seems like we have nothing actually triggering `start_wallet_loop`

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -34,8 +36,73 @@ def GetWallet1(cls, recreate=False):
...
+                print("couldn't remove wallets %s " % e)
+
+        with patch('neo.Prompt.PromptData.PromptData.Prompt'):
+            with patch('neo.Prompt.Commands.Wallet.prompt', side_effect=["testpassword", "testpassword"]):
+                # test wallet create successful
+                path = UserWalletTestCase.new_wallet_dest()
+                args = ['create', path]
+                self.assertFalse(os.path.isfile(path))
+                res = CommandWallet().execute(args)
+                self.assertEqual(str(type(res)), "<class 'neo.Implementations.Wallets.peewee.UserWallet.UserWallet'>")
```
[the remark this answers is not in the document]
jseagrave21: @LysanderGG based on @ixje's review, you should probably change this test too https://github.com/CityOfZion/neo-python/pull/733/files#diff-f8bf8cf1fb6fe5737ee276e280ad2959R154
I suppose my test set a bad precedent 😳

LysanderGG:
@ixje @jseagrave21 Thank you for the first review!

1. test for wallet (coverage)
-> Done, but there's not much to test.
Shall I test what is printed to std::out?

2. test invalid wallet (coverage)
-> Done, but there's not much to test.

3. ensure wallet loop is actually started when creating a new wallet (functionality)
The bug had been introduced with #726.
I fixed it.

Please review a second time when you have time.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -34,8 +36,84 @@ def GetWallet1(cls, recreate=False):
...
+
     # Beginning with refactored tests
 
+    def test_wallet(self):
+        with patch('neo.Prompt.PromptData.PromptData.Prompt'):
+            res = CommandWallet().execute(None)
+            self.assertEqual(res, None)
+
+    def test_wallet_wrong_command(self):
+        with patch('neo.Prompt.PromptData.PromptData.Prompt'):
```
ixje: this does not test invalid command. It passes because it has no wallet open and therefore returns `None`.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -34,8 +36,84 @@ def GetWallet1(cls, recreate=False):
...
+        PromptData.Wallet = cls.GetWallet1(recreate=True)
+
+    @classmethod
+    def tearDown(cls):
+        PromptData.Wallet = None
+
     # Beginning with refactored tests
 
+    def test_wallet(self):
+        with patch('neo.Prompt.PromptData.PromptData.Prompt'):
```
ixje: what are we trying to patch? I don't see it used.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -93,6 +171,54 @@ def test_wallet_close(self):
 
                 self.assertTrue(res)
 
+    def test_wallet_verbose(self):
+        with patch('neo.Prompt.PromptData.PromptData.Prompt'):
+            # test wallet verbose with no wallet
+            args = ['verbose']
+            res = CommandWallet().execute(args)
```
ixje: same as previous. This does not test verbose, but fails on a not having an open wallet.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -93,6 +171,54 @@ def test_wallet_close(self):
...
+            # test wallet create address with no wallet open
+            args = ['create_addr', 1]
+            res = CommandWallet().execute(args)
+            self.assertFalse(res)
+
+            self.OpenWallet()
+
+            # test wallet create address with no argument
+            args = ['create_addr']
+            res = CommandWallet().execute(args)
```
ixje: not specifically at this line, but in the function it calls (not seeing that code here); 
Can you please give a more user friendly error message and/or capture the "not enough arguments" problem sooner? We now get

```
neo> wallet create_addr
int() argument must be a string, a bytes-like object or a number, not 'NoneType'
```
which is really confusing because you might try interpret this as it indicating that "create_addr" is read as an int() argument

neo/Prompt/Commands/tests/test_wallet_commands.py
ixje: if your IDE doesn't support coverage highlighting you could alternatively run `make coverage` or have a look at the `coveralls` output in your PR as a last resort.

neo/Prompt/Commands/tests/test_wallet_commands.py
jseagrave21: Also, one way to make this test less confusing would be `return Prompt data.Wallet` which you could after as true. You could also verify `stdout` if you wanted but not necessary.

jseagrave21: I also think this test is failing on no wallet

ixje:
> I couldn't find any other test related to wallet migration so I was unable to test the command wallet migrate

Skip it. I'm going to remove all `migrate` code. It's old an no longer used.

neo/Prompt/Commands/tests/test_wallet_commands.py
LysanderGG: This test is specifically testing `test wallet verbose with no wallet`

neo/Prompt/Commands/Wallet.py
```diff
@@ -108,11 +107,10 @@ def execute(self, arguments):
```
jseagrave21: This case is still not covered. I think at least the anticipated case [here](https://coveralls.io/builds/20433094/source?filename=neo%2FPrompt%2FCommands%2FWallet.py#L101) should be tested.
I am not sure what the expected error is, but @ixje has mentioned to me in previous reviews that catchall try/except methods are undesirable. If this is still desirable logic, then I am okay with leaving it.
Also, if we don't want to cover this case then OK, I am just point it out.
https://coveralls.io/builds/20433094/source?filename=neo%2FPrompt%2FCommands%2FWallet.py#L102

neo/Prompt/Commands/Wallet.py
```diff
@@ -108,11 +107,10 @@ def execute(self, arguments):
 
             if PromptData.Wallet:
                 PromptData.Prompt.start_wallet_loop()
-                return
+                return PromptData.Wallet
 
         else:
             print("Please specify a path")
-            return
```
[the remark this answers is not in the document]
LysanderGG: I saw this, but I'm mitigated about whether it should stay as is or not. 
If someone calls `CommandWalletCreate.execute()` directly it would skip the argument validation.

@ixje Any advice?

neo/Prompt/Commands/Wallet.py
```diff
@@ -147,11 +145,9 @@ def execute(self, arguments):
                 return PromptData.Wallet
             except Exception as e:
                 print("Could not open wallet: %s" % e)
-                return
 
         else:
             print("Please specify a path")
-            return
```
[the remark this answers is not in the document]
LysanderGG: Yeah I notified it here: https://github.com/CityOfZion/neo-python/pull/733#discussion_r238490876
I think `construct_send_basic` check should stay as is because it can be called from somewhere else.

neo/Prompt/Commands/Wallet.py
LysanderGG: As a note, this is the copy of the code in the `development` branch.
I can force an exception for the sake of coverage.

jseagrave21: Right that is not what I'm going for explicitly. What it looks like is this specified case where `os.path.isfile(path)` is `True` isn't tested.

LysanderGG: Got it :+1: , I can make sure to cover these cases.

neo/Prompt/Commands/Wallet.py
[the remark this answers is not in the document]
jseagrave21: Ooh. 😊  Sorry, I didn't see that earlier. Okay.

neo/Prompt/Commands/Wallet.py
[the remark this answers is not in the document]
jseagrave21: Similar to my previous realization, I understand now that this should be caught by your new logic concerning required input. I think what you have should be good but then we should remove redundant code.

LysanderGG:
@ixje @jseagrave21 
I pushed a new commit https://github.com/CityOfZion/neo-python/pull/733/commits/d8d0b04cb2a300b9b9051add1cc63889d3f12ebe
Please have a look.

I changed the tests to use the subcommand directly instead of testing through `CommandWallet`. I think it's better for the unit tests to use the smaller unit possible.

If we want integration tests, we should probably test from the prompt directly instead of `CommandWallet` that is kind of in between.

Please let me know what you think.

Note: testing with the subcommands allows to cover all the code. I added a some edge case detection in the subcommand's `execute()` method where necessary.
In practice, when called through `CommandWallet` it is detected earlier.

jseagrave21:
@LysanderGG I prefer integrated tests which simulate user input as much as possible. When I implemented the testing method here https://github.com/CityOfZion/neo-python/pull/726#issuecomment-441996896 I wasn't sure how to test directly from prompt.py. But if that is possible, I would prefer that.

ixje:
Back here https://github.com/CityOfZion/neo-python/pull/726#discussion_r236785324 I kind of reasoned that testing "GroupCommand" vs "SubCommand" doesn't really matter, although "GroupCommand" makes it more obvious what scenario we're working on and will need testing anyway. Which would be in line with @jseagrave21 's last comment.

After your discussion above I kind of realise that people might start calling commands directly when they're extending `neo-python`. e.g. when they implement a web interface, REST API, new RPC commands or want to control neo-python programmatically. They should not have to re-create the logic for let's say `sending assets` when there's a `CommandSend` already doing this. Now being aware of this I don't think we can avoid some code duplication between the "Group Command"  and "SubCommand" so to say. For the extensibility reasons mentioned before, I would now lean towards testing commands directly while not excluding testing group commands. (let me know if that last line doesn't make sense and I'll try to rephrase)

neo/Prompt/Commands/tests/test_wallet_commands.py
LysanderGG: Amend of ^ 
I removed the early detection and checked the argument in `CommandWalletCreateAddress`.

ixje:
> I think that the best to do is keep the functionnalities not related to the command line outside of Commands so that these functions can be used elsewhere.

I actually had that in mind (separate presentation from functionality) before writing my previous comment, but thought it might lead to more confusion at this point. 

> I'm a bit lost with this big discussion; can you
No worries, same here. This is a good showcase why we should keep PR's small, because discussions can still get big :)

I think for this one let's get it in and we'll think about moving out the body of `execute()` functions before we do the big merge. That way we can have it in a clean PR which is easier to discuss. @jseagrave21 any objections at this point or things that need solving before merging?

PS: I added the last paragraph as a TODO item to the original "CLI refactoring post"

PR #667, opened 2018-10-14
Update/Add Asset and Contract Search Methods

jseagrave21:
**What current issue(s) does this address, or what feature is it adding?**

**-update-**
- Update `getassetstate` RPC method
  - Allows using "neo" or "gas" to return their assetstate
- add `showallassets` LevelDBBlockchain function
  - `showallassets` returns a list of all asset hashes
- update `SearchAssetState` LevelDBBlockchain function per https://github.com/CityOfZion/neo-python/pull/739
- update prompt.py `help`
  - update `show_asset_state` and `show_contract_state` per https://github.com/CityOfZion/neo-python/pull/741

**How did you solve this problem?**
trial and error

The inspiration for this PR comes from my work on the neo-python API reference. In most cases, assets in neo-python can be searched by their name. The contract search methods update was inspired through their tangential relationship in `prompt.py` and LevelDBBlockckchain.

**How did you make sure your solution works?**
unittest

**Are there any special changes in the code that we should be aware of?**
No

**Please check the following, if applicable:**

- [x] Did you add any tests?
- [x] Did you run `make lint`?
- [x] Did you run `make test`?
- [x] Are you making a PR to a feature branch or development rather than master?
- [x] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

PR #737, opened 2018-12-04
[refactor prompt] Adds support and tests for the show method group and the search method group

jseagrave21:
- Referring to https://github.com/CityOfZion/neo-python/issues/623, this PR adds support for the `show` method group and creates the `search` method group as described here https://github.com/CityOfZion/neo-python/issues/623#issuecomment-443991459

- Some methods have been updated with minor changes. `ShowAllAssets` was added to incorporate approved changes from https://github.com/CityOfZion/neo-python/pull/667 and add continuity to `show asset` and `show contract` methods.

Trial and Error

Unittest

No

- [x] Did you add any tests?

### DIFF
--- a/neo/Core/Blockchain.py
+++ b/neo/Core/Blockchain.py
@@ -308,6 +308,9 @@ class Blockchain:
     def SearchAssetState(self, query):
         pass
 
+    def ShowAllAssets(self, query):
+        pass
+
     def GetHeaderHash(self, height):
         pass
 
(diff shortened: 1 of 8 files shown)


ixje:
Please split this PR into multiple. It's too big to review without losing track if changes are requested or requiring a huge effort to do so. Split it at least into the `show` and `search` group commands. Preferably split `show` into 2 PR's as that would already be 5 commands per PR (you might have to PR the second one after the first is in). Thanks!

ixje:
You can PR `search` + `show part 1` in parallel. They're separate groups so can be separate PR's. Then `show part 2` comes later.

PR #738, opened 2018-12-05
[refactor prompt] Add support for show method group (pt 1)

jseagrave21:
- Reference https://github.com/CityOfZion/neo-python/pull/737

- This PR supports the first five methods from the `show` method group (see https://github.com/CityOfZion/neo-python/issues/623#issue-362278030)

Trial and Error

Unittest

No

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Show.py
+++ b/neo/Prompt/Commands/Show.py
@@ -1,0 +1,167 @@
+import os
+import psutil
+import datetime
+from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
+from neo.Prompt.PromptData import PromptData
+from neo.Prompt.Utils import get_arg
+from neo.Core.Blockchain import Blockchain
+from prompt_toolkit.shortcuts import print_formatted_text
+from prompt_toolkit.formatted_text import FormattedText
+from neocore.UInt256 import UInt256
+from neo.IO.MemoryStream import StreamManager
+from neo.Network.NodeLeader import NodeLeader
+from neo.Implementations.Notifications.LevelDB.NotificationDB import NotificationDB
+from neo.logging import log_manager
+import json
+
+
+logger = log_manager.getLogger()
+
+
+class CommandShow(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+        self.register_sub_command(CommandShowBlock())
+        self.register_sub_command(CommandShowHeader())
+        self.register_sub_command(CommandShowTx())
+        self.register_sub_command(CommandShowMem())
+        self.register_sub_command(CommandShowNodes(), ['node'])
+
+    def command_desc(self):
+        return CommandDesc('show', 'show data from the blockchain')
+
+    def execute(self, arguments):
+        item = get_arg(arguments)
+
+        if not item:
+            print("run `show help` to see supported queries")
+            return
+
+        try:
+            return self.execute_sub_command(item, arguments[1:])
+        except KeyError:
+            print(f"show: {item} is an invalid parameter")
+            return
+
+
+class CommandShowBlock(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        item = get_arg(arguments)
+        txarg = get_arg(arguments, 1)
+
+        block = Blockchain.Default().GetBlock(item)
+
+        if block is not None:
+            block.LoadTransactions()
+            bjson = json.dumps(block.ToJson(), indent=4)
+            if txarg and 'tx' in txarg:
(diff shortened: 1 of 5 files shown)


neo/Prompt/Commands/Show.py
```diff
@@ -0,0 +1,159 @@
...
+        super().__init__()
+
+        self.register_sub_command(CommandShowBlock())
+        self.register_sub_command(CommandShowHeader())
+        self.register_sub_command(CommandShowTx())
+        self.register_sub_command(CommandShowMem())
+        self.register_sub_command(CommandShowNodes(), ['node'])
+
+    def command_desc(self):
+        return CommandDesc('show', 'show data from the blockchain')
```
ixje: I think we need to generalise the description a bit more because it also includes "node memory usage", "connected nodes", "overall client state" (and who knows what in the future).

I'm out of inspiration at this point, but I'm sure you can find a nice and short generalised line 👍

neo/Prompt/Commands/Show.py
```diff
@@ -0,0 +1,159 @@
...
+        except KeyError:
+            print(f"{item} is an invalid parameter")
+            return
+
+
+class CommandShowBlock(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
```
ixje: `show block` and `show headers` without an argument throws an exception. Example
```
neo> show block
Could not execute command: object of type 'NoneType' has no len()
and then some
```

`show tx` without arguments throws a different exception
```
neo> show tx
Could not find transaction from args: 'NoneType' object is not subscriptable ([])
```

neo/Prompt/Commands/Show.py
```diff
@@ -0,0 +1,159 @@
...
+    def command_desc(self):
+        p1 = ParameterDesc('index/hash', 'the index or scripthash of the block header')
+        return CommandDesc('header', 'show the header of a specified block', [p1])
+
+
+class CommandShowTx(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
```
ixje: show tx with an invalid argument gives a not user friendly error
```
neo> show tx 1
Could not find transaction from args: Invalid UInt256 Format: 1 chars != 64 chars (['1'])
```

neo/Prompt/Commands/Show.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                print(f"Could not find transaction for hash {txid}")
+                return
+        except Exception as e:
+            print("Could not find transaction from args: %s (%s)" % (e, arguments))
+            return
+
+    def command_desc(self):
+        p1 = ParameterDesc('hash', 'the scripthash of the transaction')
+        return CommandDesc('tx', 'show a specified transaction', [p1])
```
ixje: Should this maybe be? (Testing my English grammar here)
`show _the_ specified transaction` or `show a _specific_ transaction`

jseagrave21: I think those are all correct. What is your preference? (What sounds the best to you out loud?)

neo/Prompt/Commands/Show.py
```diff
@@ -35,13 +33,13 @@ def execute(self, arguments):
         item = get_arg(arguments)
 
         if not item:
-            print("run `show help` to see supported queries")
+            print("run `%s help` to see supported queries" % CommandShow().command_desc().command)
```
LysanderGG: Should use `self` inseatd of `CommandShow()` here

neo/Prompt/Commands/Show.py
jseagrave21: How about "show useful data"
?

neo/Prompt/Commands/Show.py
ixje: If you say they're all good, then leave it as is. I'm not the native speaker here. It just felt odd when reading, but it might just be me not being used to that form.

neo/Prompt/Commands/Show.py
jseagrave21: I removed all those user friendly messages because I thought @LysanderGG 's change to CommandBase here https://github.com/CityOfZion/neo-python/pull/733#pullrequestreview-181093593 was going to be merged. Oops 😊 
I will add all the original messages back in

neo/Prompt/Commands/Show.py
jseagrave21: How about now? (I just removed the exception)

ixje:
Looks good as far as I could test and see now 👍

PR #740, opened 2018-12-06
[refactor-prompt] Implement CommandWalletClaimGas

LysanderGG:
DO NOT MERGE AS IS.

Migrate the command `wallet claim` to the new Command system.

Implementation of `CommandWalletClaimGas`.

Added unit tests.

There are 2 tests failing for an unknown reason, I will add comments directly in the code.

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Wallet.py
+++ b/neo/Prompt/Commands/Wallet.py
@@ -38,6 +38,7 @@ class CommandWallet(CommandBase):
         self.register_sub_command(CommandWalletSend())
         self.register_sub_command(CommandWalletSendMany())
         self.register_sub_command(CommandWalletSign())
+        self.register_sub_command(CommandWalletClaimGas())
 
     def command_desc(self):
         return CommandDesc('wallet', 'manage wallets')
@@ -210,6 +211,33 @@ class CommandWalletCreateAddress(CommandBase):
         return CommandDesc('create_addr', 'create a new wallet address', params=[p1])
 
 
+class CommandWalletClaimGas(CommandBase):
+
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        from_addr_str = None
+        max_coins_per_claim = None
+
+        args = arguments
+        if args:
+            args, from_addr_str = get_from_addr(args)
+
+        if args:
+            max_coins_per_claim = get_arg(args, 0, convert_to_int=True)
+            if not max_coins_per_claim:
+                print("max_coins_to_claim must be an integer")
+                return None, False  # return same as failure in ClaimGas().
+
+        return ClaimGas(PromptData.Wallet, True, from_addr_str, max_coins_per_claim)
+
+    def command_desc(self):
+        p1 = ParameterDesc('max_coins_to_claim', 'maximum number of coins to claim', optional=True)
+        p2 = ParameterDesc('--from-addr', 'source address to claim gas from (if not specified, take first address in wallet)', optional=True)
+        return CommandDesc('claim', 'claim gas', params=[p1, p2])
+
+
 #########################################################################
 #########################################################################
 
@@ -321,22 +349,24 @@ def AddAlias(wallet, addr, title):
         print(e)
 
 
-def ClaimGas(wallet, require_password=True, args=None):
+def ClaimGas(wallet, require_password=True, from_addr_str=None, max_coins_to_claim=None):
     """
     Args:
         wallet:
         require_password:
-        args:
+        from_addr_str:
+        max_coins_to_claim:
     Returns:
         (claim transaction, relayed status)
             if successful: (tx, True)
             if unsuccessful: (None, False)
     """
-    if args:
-        params, from_addr_str = get_from_addr(args)
-    else:
-        params = None
-        from_addr_str = None
+    if not wallet:
+        print("Please open a wallet")
+        return None, False
+    if max_coins_to_claim and max_coins_to_claim <= 0:
+        print("max_coins_to_claim must be greater than zero")
+        return None, False
 
     unclaimed_coins = wallet.GetUnclaimedCoins()
 
@@ -345,17 +375,8 @@ def ClaimGas(wallet, require_password=True, args=None):
         print("no claims to process")
         return None, False
 
-    max_coins_per_claim = None
-    if params:
-        max_coins_per_claim = get_arg(params, 0, convert_to_int=True)
-        if not max_coins_per_claim:
-            print("max_coins_to_claim must be an integer")
-            return None, False
-        if max_coins_per_claim <= 0:
-            print("max_coins_to_claim must be greater than zero")
-            return None, False
-    if max_coins_per_claim and unclaimed_count > max_coins_per_claim:
-        unclaimed_coins = unclaimed_coins[:max_coins_per_claim]
+    if max_coins_to_claim and unclaimed_count > max_coins_to_claim:
+        unclaimed_coins = unclaimed_coins[:max_coins_to_claim]
 
     unclaimed_coin_refs = [coin.Reference for coin in unclaimed_coins]
 
@@ -388,9 +409,9 @@ def ClaimGas(wallet, require_password=True, args=None):
     wallet.Sign(context)
 
     print("\n---------------------------------------------------------------")
-    print("Will make claim for %s GAS" % available_bonus.ToString())
-    if max_coins_per_claim and unclaimed_count > max_coins_per_claim:
-        print("NOTE: You are claiming GAS on %s unclaimed coins. %s additional claim transactions will be required to claim all available GAS." % (max_coins_per_claim, math.floor(unclaimed_count / max_coins_per_claim)))
+    print(f"Will make claim for {available_bonus.ToString()} GAS")
+    if max_coins_to_claim and unclaimed_count > max_coins_to_claim:
+        print(f"NOTE: You are claiming GAS on {max_coins_to_claim} unclaimed coins. {math.floor(unclaimed_count / max_coins_per_claim)} additional claim transactions will be required to claim all available GAS.")
     print("------------------------------------------------------------------\n")
 
     if require_password:
@@ -414,12 +435,10 @@ def ClaimGas(wallet, require_password=True, args=None):
             print("Relayed Tx: %s " % claim_tx.Hash.ToString())
             wallet.SaveTransaction(claim_tx)
         else:
-
             print("Could not relay tx %s " % claim_tx.Hash.ToString())
         return claim_tx, relayed
 
     else:
-
         print("could not sign tx")
 
     return None, False
--- a/neo/Prompt/Commands/Wallet.py
+++ b/neo/Prompt/Commands/Wallet.py
@@ -218,19 +218,19 @@ class CommandWalletClaimGas(CommandBase):
 
     def execute(self, arguments):
         from_addr_str = None
-        max_coins_per_claim = None
+        max_coins_to_claim = None
 
         args = arguments
         if args:
             args, from_addr_str = get_from_addr(args)
 
         if args:
-            max_coins_per_claim = get_arg(args, 0, convert_to_int=True)
-            if not max_coins_per_claim:
+            max_coins_to_claim = get_arg(args, 0, convert_to_int=True)
+            if not max_coins_to_claim:
                 print("max_coins_to_claim must be an integer")
                 return None, False  # return same as failure in ClaimGas().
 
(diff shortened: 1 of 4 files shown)


neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -240,6 +257,61 @@ def test_wallet_create_address(self):
...
+            claim_tx, relayed = CommandWallet().execute(args)
+            self.assertEqual(claim_tx, None)
+            self.assertFalse(relayed)
+
+        # test successfull
+        with patch('neo.Prompt.Commands.Wallet.prompt', side_effect=[WalletFixtureTestCase.wallet_1_pass()]):
+            args = ['claim']
+            claim_tx, relayed = CommandWallet().execute(args)
+            self.assertIsInstance(claim_tx, ClaimTransaction)
+            # self.assertTrue(relayed)
```
LysanderGG: `relayed` is always false here, probably because of `neo/Prompt/Commands/tests/test_claim_command.py`.
How should I generate new wallets with GAS to claim for test?
Please tell me if you have another idea.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -240,6 +257,61 @@ def test_wallet_create_address(self):
...
+
+        # test nothing to claim anymore
+        # with patch('neo.Prompt.Commands.Wallet.prompt', side_effect=[WalletFixtureTestCase.wallet_1_pass()]):
```
LysanderGG: I think that this test is failing because of the previous `relayed` to be false.

neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -240,6 +257,61 @@ def test_wallet_create_address(self):
...
+        #     args = ['claim']
+        #     claim_tx, relayed = CommandWallet().execute(args)
+        #     self.assertEqual(claim_tx, None)
+        #     self.assertFalse(relayed)
+
+        # test with --from_addr
```
LysanderGG: How could I test this?

ixje: I don't know the wallet structure from the top of my head, but what you would do with a `send` (which supports the same argument) is;
- create the tx
- inspect the "vin" list of that tx and make sure that it has 1 input where the `address` field equals to whatever is specified in `--from_addr`

You would need a wallet with at least 2 addresses that hold NEO (assuming that created claimable GAS) or 2 addresses with unclaimed GAS. Then you can test if the claim is actually taken from the right address. I don't know if this is currently supported, never looked at that part.

neo/Prompt/Commands/tests/test_wallet_commands.py
ixje: I don't think that `test_claim_command.py` plays a role here. Reason being that your testcase `UserWalletTestCase` extends `WalletFixtureTestCase` which in turn extends `BlockchainFixtureTestCase`. 
- `BlockchainFixtureTestCase` makes it so that each run we reset the blockchain state. Meaning any transactions / changes we performed on the blockchain in a test get reset.
- `WalletFixtureTestCase` adds to this that each wallet gets copied when running the test to make sure we never modify the original wallet. As with the blockchain, this ensures that we always have the same starting state for wallets when running a test case.

ixje: My first check would be to validate that the wallets you're using actually have unclaimed GAS. Although it should looking the `test_claim_command.py`. All I can suggest is stepping through the code because that's what I would have to do as well to debug. I'm guessing this fails:
https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/Network/NodeLeader.py#L323

ixje: Oh that's important information. In that case I'd probably look at which tests use `WalletFixtureTestCase` and disable those tests completely to verify if any of them mess up the state (for example because they do not copy the wallet files). 

Note that the overall tests fail due to a pylinting error not because a test is failing (although that might make sense since you commented parts out). See: https://travis-ci.org/CityOfZion/neo-python/builds/464202907#L3226

LysanderGG: I saw this, I will fix it :sweat_smile: 
The parts I commented out fail in local when running `make test`.

I will investigate, thank you for your directions!

LysanderGG: I opened the PR #742

neo/Prompt/Commands/tests/test_wallet_commands.py
LysanderGG: Thank you for the detailed answer.
It seems pretty time consuming to add this test, I would suggest to skip it for now 😅

ixje: I don't think it's a good idea to skip, because it likely isn't working at this point (like the `max_coins` param isn't). Perhaps @jseagrave21 can help you on it. He's done a similar thing before with `send`/`sendmany`

ixje:
Regarding the max claims. Have a moment and see what effort it takes to fix it. Otherwise I'm ok with removing that parameter. I don't see the use of partial claims and I don't think anyone will miss it if it was never working to begin with

neo/Prompt/Commands/tests/test_wallet_commands.py
jseagrave21: @LysanderGG let me know if I can help.

neo/Prompt/Commands/Wallet.py
```diff
@@ -210,6 +211,33 @@ def command_desc(self):
...
+            if not max_coins_per_claim:
+                print("max_coins_to_claim must be an integer")
+                return None, False  # return same as failure in ClaimGas().
+
+        return ClaimGas(PromptData.Wallet, True, from_addr_str, max_coins_per_claim)
+
+    def command_desc(self):
+        p1 = ParameterDesc('max_coins_to_claim', 'maximum number of coins to claim', optional=True)
+        p2 = ParameterDesc('--from-addr', 'source address to claim gas from (if not specified, take first address in wallet)', optional=True)
+        return CommandDesc('claim', 'claim gas', params=[p1, p2])
```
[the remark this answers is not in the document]
LysanderGG: Handled with b45e275c80b02e6b40901d8d85ef35ce5bc402d3
It will now be aligned with the longest parameter.

Please tell me if you want 15 as the minimum length (there is no minimum for now).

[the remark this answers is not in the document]
ixje: nice change. A minimum would be a good idea to have some consistency if params are short.

ixje:
Can it be that you do a "git reset --soft", update code based on feedback and then force push? 

If so, can you please just do regular updates and pushes in the future? I can squash them before merging to get rid of the extra commits. The reason is that by resetting + force pushing I can't tell what changes you made since my review (because everything is new in the last commit). 

Having said that; let me know when I should review again. I think the "--from-addr" is still to be done, and please let me know what you've did with "max claims" (either remove or implement is fine with me, see previous comment)

LysanderGG:
I did a `git commit --amend` or a `git rebase`. I'm used to work with gitlab where it's possible to see the diff with previous commits even after a force push.

> can you please just do regular updates and pushes in the future?
:ok_hand: 

I didn't handle the `--from-addr` and `max claims` yet. 
I will let you know when done!

neo/Prompt/Commands/Wallet.py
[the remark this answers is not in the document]
LysanderGG: Minimum added with 8c7c82a640470633c9582158e617d0668d900d41

ixje:
**A little bit of background info on GAS claiming to start the story**
If you hold NEO, then over time you earn GAS when the network is used as network usage costs GAS and this is distributed to the token holders per ratio. The simple (not fully accurate) way it works is; take the block index from which you became the owner of NEO up to the block index in which you decide to make the GAS claim. Apply some formula and you are entitled for `n` GAS. 

This is the "Unavailable" value in the first output below. That's the GAS you can claim and it will keep rising if you do not sending NEO. Say what?! Before you can claim it you have to send NEO to yourself. Why is that? Because we need to know and persist the "end block" for your claim to avoid double claiming. By sending NEO you create a new UTXO pair and you close the old one (see unspent/spent at this [coz rewards TX example](https://neotracker.io/tx/72766874e3ebe9ebb9596ea6c5915cdcebb5a7dd3e2fe0b443487bbe77865ef2)). So by sending NEO, the GAS becomes available that belonged to that UTXO pair (btw see some bitcoin wiki if you don't know the term, they have easy pictures explaining UTXO). This will be shown in the "claims.available". 

```
    "claims": {
        "available": "0.0",
        "unavailable": "108264.0"
    }
}
neo> wallet claim help

Claim gas

Usage: wallet claim (max_coins_to_claim) (--from-addr)

max_coins_to_claim - (Optional) maximum number of coins to claim
--from-addr        - (Optional) source address to claim gas from (if not specified, take first address in wallet)
neo> wallet claim 10
no claims to process
neo> wallet claim 10 --from-addr AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y
no claims to process
neo> wallet claim 10 --from-addr=AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y
no claims to process
```
Note: if you received all your NEO's in 1 transaction, then you have one UTXO and sending will free up all "claims.unavailable".  If you received e.g. 25 in 1 transaction, and 75 NEO in another. Then sending 25 NEO will release 25%, but sending 26 NEO will release all (assuming the sorting algorithm takes the UTXO with 25 NEO first, then the second UTXO for the last missing NEO). Yes, it could be possible to take the second UTXO as that has 75 NEO which would cover the 26 we need.

**Back to the PR/issue**
Ok, so hoping I didn't lose you there. If we now look at the `wallet claim`, then `max_coins_to_claim` does not specify `108264.0` (as in the example output below) but refers to the 1 or 2 UTXO's in the explanation above. A user has no clue how many there are or that it refers to the UTXO's. In other words, a user experience nightmare. note; this is not your fault. (the example used unfortunately only has 1 UTXO, that's why it shows the `108352.0` unavailable value. I hope you get the idea)

**How to move forward**

1. I think we should remove the `max_coins_to_claim`. As it's too confusing. 

2. What we should also do is add a `--to-addr`. By default claims will go to the first address in the wallet, but that might not be the place where you want it to go. If you look at the wallet claim transaction below, you'll see that the `vout.address` is the destination address. So if the param is used it should alter that
```
neo> wallet claim 10

---------------------------------------------------------------
Will make claim for 108352.0 GAS
------------------------------------------------------------------

Enter your password to complete this claim
[Password]> **********
claim tx: {
    "txid": "0xb31f9658026f8ab5b1c61d97cd27412242fdfd723f4b603db72d9a81aa68fd59",
    "size": 203,
    "type": "ClaimTransaction",
    "version": 0,
    "attributes": [],
    "vout": [
        {
            "n": 0,
            "value": "108352",
            "address": "AY1PVgK92HCUNXXTfBDH3VfWuUXGSC6hyU"
        }
    ],
    "vin": [],
    "sys_fee": "0",
    "net_fee": "0",
    "scripts": [
        {
            "invocation": "408413b3e22640d1fe144c3f1bd78cd8bb2b8eba284e2f3ccff8305ed3bddfdbfb036d51c27a18b4185db55febf40a03255400d42df2e08e90bd5b3d9ca7e402d8",
        }
    ],
    "claims": [
        {
            "txid": "0x9c146236c4ce685090442630bbd91e07e9d34db2fa5fd7aca81846621e801ee2",
            "vout": 0
        }
    ]
}
```


This became a way longer post than I anticipated. Feel free to DM me on discord/slack if it's not fully clear or if you have questions

jseagrave21:
I saw this and I thought I would ask if we wanted to implement this while you are working on it @LysanderGG: https://github.com/neo-project/neo-cli/pull/268/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR10 cc @ixje

LysanderGG:
@ixje I removed `max_coins_to_claim`.
As said in the previous post, I'd be in favor of adding ` --to-addr` and `changeAddress` after the prompt refactor (I'd like to try and finish it this year).

ixje:
I'm OK with postponing. I created an issue for it here: https://github.com/CityOfZion/neo-python/issues/755

ixje:
After restarting the build everything looks good. Thanks! 💯

PR #741, opened 2018-12-06
[refactor prompt] Add support for show method group (pt 2)

jseagrave21:
- Reference https://github.com/CityOfZion/neo-python/pull/737

- This PR supports the last five methods from the `show` method group (see https://github.com/CityOfZion/neo-python/issues/623#issue-362278030)

Trial and Error

Unittest

- `ShowAllAssets` was added so `show assets` mirrors `show contracts`
  - originally implemented here https://github.com/CityOfZion/neo-python/pull/667

- Added the ability to use 'neo' or 'gas' to show the corresponding assets
  -  originally implemented here https://github.com/CityOfZion/neo-python/pull/667

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Show.py
+++ b/neo/Prompt/Commands/Show.py
@@ -25,6 +25,11 @@ class CommandShow(CommandBase):
         self.register_sub_command(CommandShowTx())
         self.register_sub_command(CommandShowMem())
         self.register_sub_command(CommandShowNodes(), ['node'])
+        self.register_sub_command(CommandShowState())
+        self.register_sub_command(CommandShowNotifications())
+        self.register_sub_command(CommandShowAccount())
+        self.register_sub_command(CommandShowAsset())
+        self.register_sub_command(CommandShowContract())
 
     def command_desc(self):
         return CommandDesc('show', 'show useful data')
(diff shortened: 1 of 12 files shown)


neo/Prompt/Commands/Show.py
```diff
@@ -167,3 +172,171 @@ def execute(self, arguments=None):
...
+
+    def command_desc(self):
+        return CommandDesc('state', 'show the status of the node')
+
+
+class CommandShowNotifications(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
```
ixje:
```
neo> show notifications -1
Could not parse block height can't convert negative int to unsigned

neo> show notifications bla
Could not parse block height invalid literal for int() with base 10: 'bla'

neo> show notifications
Could not execute command: object of type 'NoneType' has no len()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/
and then some
```
unfriendly error message for an invalid arguments

similar errors are also present for `show contract 234` and `show asset 234`. I stopped reviewing at this point. Please do some manual testing, as these can be captured before PR'ing. Thanks!

jseagrave21:
Not sure what is going on with coveralls but clicking  on the latest coveralls comment [here](https://coveralls.io/builds/20524550) shows accurate coverage of `Show.py`. However, I don't understand the "uncovered" lines, I don't think that is accurate.

@ixje @LysanderGG I think this PR is ready for another review.

neo/Prompt/Commands/Show.py
```diff
@@ -167,3 +173,189 @@ def execute(self, arguments=None):
...
+                return asset.ToJson()
+            else:
+                print("Asset %s not found" % item)
+                return
+        else:
+            print("Please specify an asset hash or name")
+            return
+
+    def command_desc(self):
+        p1 = ParameterDesc('name/assetId/all', 'the name or assetId of the asset, or "all" shows all assets\n\n')
```
ixje: the extra help is not shown because the `)` terminates the `description` string after the first line

neo/Prompt/Commands/Show.py
```diff
@@ -167,3 +173,189 @@ def execute(self, arguments=None):
...
+class CommandShowContract(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        item = get_arg(arguments)
+        if item is not None:
+            if item.lower() == "all":
+                contracts = Blockchain.Default().ShowAllContracts()
+                print("Contracts: %s" % contracts)
```
ixje: I think we can give a better user experience than this
```
neo> show asset all
Assets: dict_keys([b'602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7', b'c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b'])
```
I think we could at least give them a column like looking output showing `asset name` - `asset id`

neo/Prompt/Commands/Show.py
```diff
@@ -167,3 +173,189 @@ def execute(self, arguments=None):
...
+            asset = Blockchain.Default().GetAssetState(assetId.ToBytes())
+
+            if asset is not None:
+                print(json.dumps(asset.ToJson(), indent=4))
+                return asset.ToJson()
+            else:
+                print("Asset %s not found" % item)
+                return
+        else:
+            print("Please specify an asset hash or name")
```
ixje:
```
neo> show asset
Please specify an asset hash or name
```
Note that `all` is also an option but is not mentioned here. We have to update this text and use the "generic multiple option name" we come up with. See main feedback what I mean with that.

neo/Prompt/Commands/Show.py
jseagrave21: Would you also like this change applied to `ShowAllContracts`? (I formatted `ShowAllAssets` to match `ShowAllContracts`)

jseagrave21:
>I'm looking for a word that would neatly express a group of possible input options. I don't think args is a good fit. My current best idea is specifier or identifier but it doesn't feel smooth or accurate. I'm hoping for ideas from you :)
>
I am thinking "attribute" which is in line with Python

jseagrave21:
Note to self (cc @ixje): Prior to merging remember to
- [ ] Update `contractstate` to match parody with neo-cli (see https://github.com/CityOfZion/neo-python/pull/750)
- [ ] Update Show.py. Specifically,`CommandShowContract` to remove mention of 'code' key.

ixje:
#750 is merged, I approved the new changes. Let me know when you think it's ready.

jseagrave21:
@ixje I think this PR should be good to go 👍

ixje:
💯

PR #742, opened 2018-12-06
Fix NodeLeader was not reset between test cases.

LysanderGG:
Related to https://github.com/CityOfZion/neo-python/pull/740#discussion_r239346845

`NodeLeader` was not reset between test cases.
This would lead to some problems when the blockchain is reset but the `NodeLeader` instance members are not, especially `NodeLeader.KnownHashes` and `MemPool`.
Transactions were still considered as known.

1. Add missing members in `NodeLeader.Setup()`
2. Call `NodeLeader.Instance().Setup()` in `BlockchainFixtureTestCase.setUpClass()`
3. Also removed the unsued member `NodeLeader.MissionsGlobal`

Unit tests succeed.

Note: maybe this could be merged into develop if the problem is confirmed and the fix correct.

- [ ] Did you add any tests?

### DIFF
--- a/neo/Network/NodeLeader.py
+++ b/neo/Network/NodeLeader.py
@@ -63,7 +63,6 @@ class NodeLeader:
     BREQMAX = 10000
 
     KnownHashes = []
-    MissionsGlobal = []
     MemPool = {}
     RelayCache = {}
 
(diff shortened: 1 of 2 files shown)


ixje:
👍

PR #745, opened 2018-12-08
Fix unconfirmed coins when sending via RPC

jseagrave21:
In the #python channel on Discord, @serefbulbul found an issue where transactions were failing on "Insufficient Funds" without funds actually being insufficient.

Upon testing, I found that the first transaction was successful but every transaction after failed. So, I checked `getbalance` and found the following:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "Balance": "32045.99999999",
        "Confirmed": "0.0"
    }
}
```

When reading the description of `SaveTransaction`:
https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/Wallets/Wallet.py#L1104-L1118

I moved `self.wallet.SaveTransaction(tx)` so the coins can be confirmed.

Manual Testing

No

- [ ] Did you add any tests?


ixje:
It is not clear to me why moving this fixes the issue from the Discord user. Can you elaborate why you think that is the case. The description says, confirming doesn't happen until after the block has been processed on the network. So whether we save the is transaction before or after we relay it to the network doesn't look like it would matter for the described issue, or am I missing something?

What I do realise while writing this is that we'd probably want to save the transaction if relaying fails. So that's a fix we could do regardless if it fixes the discord issue.

jseagrave21:
When sending from CLI we have it set up to only save if the transaction relays successfully. https://github.com/CityOfZion/neo-python/blob/master/neo/Prompt/Commands/Send.py#L208-L209
But we decided not to do this because it doesn't match neo-cli implementation of the RPC method
https://github.com/neo-project/neo/blob/master/neo/Network/RPC/RpcServer.cs#L479-L484
See https://github.com/CityOfZion/neo-python/pull/680#issuecomment-437636680

To test, I built a RPC server using privnet and send a transaction using "sendmany" similar to the way the Discord user described. I sent 1 GAS to two separate addresses. The first transaction was successful as I documented in Discord; however, every transaction after failed on "Insufficient Funds" like the user described.
After moving the line of code, I was able to make transactions normally again.

jseagrave21:
@ixje how do I open a wallet and query one of the C# nodes on privnet?

ixje:
> @ixje how do I open a wallet and query one of the C# nodes on privnet?
the privnet consists of 4 consensus nodes. Each consensus node must have a wallet open to work, so you don't have to do anything there. The CoZ privnet also enables RPC by default, thus all that you'll have to do is change the ip/port in your experiment to point to one of the privnet hosts.

jseagrave21:
@ixje Well, I tried all of the consensus ports and got the error "Insufficient Funds" for all of them. But, I don't know if they actually have wallets open with a balance. Moreover, I tried changing my test a little and switched the order back to the original. Just like you thought, it didn't make a difference, so you were right.

So, I continued testing. I tried increasing the speed of transaction requests and this seemed to trigger the insufficient funds error. Here are the last few transactions please notice change for the last transaction prior to "Insufficient Funds".
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0x39241a3dd7b65f190a17d2bdd21c0439ef3feed2923be55e06324ddd29909789",
        "size": 343,
        "type": "ContractTransaction",
        "version": 0,
        "attributes": [
            {
                "usage": 32,
                "data": "23ba2703c53263e8d6e522dc32203339dcd8eee9"
            }
        ],
        "vout": [
            {
                "n": 0,
                "value": "1",
                "address": "AVSENo13gRPaVbXJKyPTqe1kREG9tAvu4E"
            },
            {
                "n": 1,
                "value": "1",
                "address": "AGzk6f5UrcpbQ3gZvaYvbx49GGscpMH1HP"
            },
            {
                "n": 2,
                "value": "30",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        ],
        "vin": [
            {
                "txid": "0xeeb8c5072dc580d166d78edc935a687df65003e9ae535d5f1479e769bf6a55e5",
                "vout": 2
            }
        ],
        "sys_fee": "0",
        "net_fee": "0",
        "scripts": [
            {
                "invocation": "40118688adce98559dd0bae4b292542de6596583f7f3427165a9e72e64b39c268b2e0e5869ebd453af1fb7545c78e96ecefe78e25bdcd7683509623a4b5f13bea3",
            }
        ]
    }
}
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0x16f25a2f0ca71ff739dae4d48fe4b1a39915adc81642fa06c4858122e8db8c7b",
        "size": 343,
        "type": "ContractTransaction",
        "version": 0,
        "attributes": [
            {
                "usage": 32,
                "data": "23ba2703c53263e8d6e522dc32203339dcd8eee9"
            }
        ],
        "vout": [
            {
                "n": 0,
                "value": "1",
                "address": "AVSENo13gRPaVbXJKyPTqe1kREG9tAvu4E"
            },
            {
                "n": 1,
                "value": "1",
                "address": "AGzk6f5UrcpbQ3gZvaYvbx49GGscpMH1HP"
            },
            {
                "n": 2,
                "value": "15982",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        ],
        "vin": [
            {
                "txid": "0xc2f7fac79531d94d406367c7feafe425f893a580fa703c7b4df9572f5944df5a",
                "vout": 0
            }
        ],
        "sys_fee": "0",
        "net_fee": "0",
        "scripts": [
            {
                "invocation": "4069a5266440e577fb97262e7272de68c3a9c6cdf5db6e4df12c43998b217b5d176fb310184372de3a4c0efdf0aa2c55f2eab50ba655ade754eb68cd1adff8fc3f",
            }
        ]
    }
}
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -300,
        "message": "Insufficient funds"
    }
}
```
I have no idea where the 15982 GAS comes from but I think this is what causes the issue. Thoughts?

ixje:
I think that's just a different coinreference in takes the funds from. Inspect tx `0xc2f7fac79531d94d406367c7feafe425f893a580fa703c7b4df9572f5944df5a` and look at the `vout` with `n` == `2`. The value should be `15984` (15982+1+1). If not then that might be the issue.

jseagrave21:
Nope, you are right:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xc2f7fac79531d94d406367c7feafe425f893a580fa703c7b4df9572f5944df5a",
        "size": 203,
        "type": "ClaimTransaction",
        "version": 0,
        "attributes": [],
        "vout": [
            {
                "n": 0,
                "value": "15984",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        ],
        "vin": [],
        "sys_fee": "0",
        "net_fee": "0",
        "scripts": [
            {
                "invocation": "4004d2fe570d5f752d3ddaec1b1330b1e2433c8af4eefaa07b664a3de833fc32892fc7a0acadca7b939a5077bcfc3ebf1379763bf5f7d5a57a2f4279580cf68ec8",
            }
        ],
        "claims": [
            {
                "txid": "0x83df8bd085fcb60b2789f7d0a9f876e5f3908567f7877fcba835e899b9dea0b5",
                "vout": 0
            }
        ],
        "blockhash": "0x83973124067c2814b5b979bf259f4a038cc6b5c42f10473902b9a0e262ca8d07",
        "confirmations": 349,
        "blocktime": 1539130301
    }
}
```

jseagrave21:
Okay, I just tested it some more. It seems like the only problem is waiting for the transactions to be confirmed. The first transaction calls a small coinreference (~30) and the second transaction calls the larger coinreference (~15980). The third shows Insufficient Funds because, I am assuming, there are no confirmed coins in the wallet. Once the previous transactions are confirmed, then transactions may resume. Thoughts?

ixje:
> The third shows "Insufficient Funds" because, I am assuming, there are no confirmed coins in the wallet
we should be able to confirm this by querying the DB for all coinreferences for the address/wallet.

> Should we update the user feedback to give better understanding why there is no tx? 👇
I wouldn't know what else to say that's not going to confuse them more. The coinreferences is just 1 explanation (which we haven't fully confirmed yet), another is just not being fully synced.

jseagrave21:
By analyzing the `verbose` wallet we can find the coinreferences:
```
Wallet {
    "path": "neo-privnet.wallet",
    "addresses": [
        {
            "version": 0,
            "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
            "script_hash": "e9eed8dc39332032dc22e5d6e86332c50327ba23",
            "frozen": false,
            "votes": [],
            "balances": {
                "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b": "100000000.0",
                "0x602c79718b16e442de58778e148d0b1084e3b2dffd5de6b7b16cee7969282de7": "16024.0"
            },
            "is_watch_only": false
        }
    ],
    "height": 2034,
    "percent_synced": 100,
    "synced_balances": [
        "[NEO]: 100000000.0 ",
        "[NEOGas]: 16024.0 "
    ],
    "public_keys": [
        {
            "Address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y",
            "Public Key": "031a6c6fbbdf02ca351745fa86b9ba5a9452d785ac4f7fc2b7548ca2a46c4fcf4a"
        }
    ],
    "tokens": [],
    "claims": {
        "available": "0.0",
        "unavailable": "216.0"
    },
    "coins": [
        {
            "Reference": {
                "txid": "0x803ec81b9ddb7dec5c914793a9e61bf556deafb561216473ad7a8ee7a91979cc",
                "vout": 0
            },
            "Output": {
                "n": 0,
                "value": "40",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        },
        {
            "Reference": {
                "txid": "0x4ba4d1f1acf7c6648ced8824aa2cd3e8f836f59e7071340e0c440d099a508cff",
                "vout": 0
            },
            "Output": {
                "n": 0,
                "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                "value": "100000000",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        },
        {
            "Reference": {
                "txid": "0xc2f7fac79531d94d406367c7feafe425f893a580fa703c7b4df9572f5944df5a",
                "vout": 0
            },
            "Output": {
                "n": 0,
                "value": "15984",
                "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
            }
        }
    ],
    "transactions": [
        {
            "txid": "0xdc44739e2f97743f2ed258988327560e2185ed13eec0097938eef4aea584bf04",
            "size": 436,
            "type": "ContractTransaction",
            "version": 0,
            "attributes": [],
            "vout": [
                {
                    "n": 0,
                    "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                    "value": "100000000",
                    "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
                }
            ],
            "vin": [
                {
                    "txid": "0x7aadf91ca8ac1e2c323c025a7e492bee2dd90c783b86ebfc3b18db66b530a76d",
                    "vout": 0
                }
            ],
            "sys_fee": "0",
            "net_fee": "0",
            "scripts": [
                {
                    "invocation": "406c0e529d43daba69d8daaea954f697e17b4c1e2ff596c82eeb0c8a8d1d0dea2268e37b1e653d8fb488e74396f478067c18d0741d956ab8935a67f40eb67c93d940a9684c5497bd06d59e88bbedb75666c2b7f92a8304c8e280c94a495fa4c2d90436caa0c651c3f816f37e0add5623594bc782af38a461d77b548f1ab5de3a2bbe402cfbabd5ac2bd5e6bd69c7a61ba79bf1b37c4e8c1b20f23a48d4896e67a6f25016d39ed4c417fd00d9af455638c9d3fbacc049b2e467d2622827da41a0508881",
                    "verification": "532102103a7f7dd016558597f7960d27c516a4394fd968b9e65155eb4b013e4040406e2102a7bc55fe8684e0119768d104ba30795bdcc86619e864add26156723ed185cd622102b3622bf4017bdfe317c58aed5f4c753f206b7db896046fa7d774bbc4bf7f8dc22103d90c07df63e690ce77912e10ab51acc944b66860237b608c4f8f8309e71ee69954ae"
                }
            ]
        },
        {
            "txid": "0x83df8bd085fcb60b2789f7d0a9f876e5f3908567f7877fcba835e899b9dea0b5",
            "size": 223,
            "type": "ContractTransaction",
            "version": 0,
            "attributes": [
                {
                    "usage": 32,
                    "data": "23ba2703c53263e8d6e522dc32203339dcd8eee9"
                }
            ],
            "vout": [
                {
                    "n": 0,
                    "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                    "value": "100000000",
                    "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
                }
            ],
            "vin": [
                {
                    "txid": "0xdc44739e2f97743f2ed258988327560e2185ed13eec0097938eef4aea584bf04",
                    "vout": 0
                }
            ],
            "sys_fee": "0",
            "net_fee": "0",
            "scripts": [
                {
                    "invocation": "401b51c85fb9f6eaabdba5bcc257b00de28ed717534dc0ae8401fac54c0f438c1a3068bb4027f3f06100a4899e3cd11e126260e67fad0048be78901cca1b7a22c4",
                }
            ]
        },
        {
            "txid": "0x803ec81b9ddb7dec5c914793a9e61bf556deafb561216473ad7a8ee7a91979cc",
            "size": 203,
            "type": "ClaimTransaction",
            "version": 0,
            "attributes": [],
            "vout": [
                {
                    "n": 0,
                    "value": "40",
                    "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
                }
            ],
            "vin": [],
            "sys_fee": "0",
            "net_fee": "0",
            "scripts": [
                {
                    "invocation": "40bcd1fe766f8b3702f26a7b1161bc6dccaf45afe7b46ee679fcd09a46becb25580fc2f2cb85a2d97052369b9f53d35038785be1cf21f7e421946c51bab06579b8",
                }
            ],
            "claims": [
                {
                    "txid": "0xdc44739e2f97743f2ed258988327560e2185ed13eec0097938eef4aea584bf04",
                    "vout": 0
                }
            ]
        },
        {
            "txid": "0x4ba4d1f1acf7c6648ced8824aa2cd3e8f836f59e7071340e0c440d099a508cff",
            "size": 223,
            "type": "ContractTransaction",
            "version": 0,
            "attributes": [
                {
                    "usage": 32,
                    "data": "23ba2703c53263e8d6e522dc32203339dcd8eee9"
                }
            ],
            "vout": [
                {
                    "n": 0,
                    "asset": "0xc56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b",
                    "value": "100000000",
                    "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
                }
            ],
            "vin": [
                {
                    "txid": "0x83df8bd085fcb60b2789f7d0a9f876e5f3908567f7877fcba835e899b9dea0b5",
                    "vout": 0
                }
            ],
            "sys_fee": "0",
            "net_fee": "0",
            "scripts": [
                {
                    "invocation": "40a3799c78dec17823fde75233793a7039bf2b1dbca4383a6eef1ac829460ba14c1e6a50ab1f2174e689bebfc0bb7accc965a6fe3e46d517b317bba1325b7fdaca",
                }
            ]
        },
        {
            "txid": "0xc2f7fac79531d94d406367c7feafe425f893a580fa703c7b4df9572f5944df5a",
            "size": 203,
            "type": "ClaimTransaction",
            "version": 0,
            "attributes": [],
            "vout": [
                {
                    "n": 0,
                    "value": "15984",
                    "address": "AK2nJJpJr6o664CWJKi1QRXjqeic2zRp8y"
                }
            ],
            "vin": [],
            "sys_fee": "0",
            "net_fee": "0",
            "scripts": [
                {
                    "invocation": "4004d2fe570d5f752d3ddaec1b1330b1e2433c8af4eefaa07b664a3de833fc32892fc7a0acadca7b939a5077bcfc3ebf1379763bf5f7d5a57a2f4279580cf68ec8",
                }
            ],
            "claims": [
                {
                    "txid": "0x83df8bd085fcb60b2789f7d0a9f876e5f3908567f7877fcba835e899b9dea0b5",
                    "vout": 0
                }
            ]
        }
    ]
}
```
The two references are "15984" and "40". Also, whenever the error "Insufficient Funds" comes in these tests a query of "getbalance" returns `"Confirmed": "0.0"`
as shown in https://github.com/CityOfZion/neo-python/pull/745#issue-237080939

@ixje based on the fact that there isn't a better user feedback, I recommend closing this issue with the lesson learned being: **Wait for transactions to be confirmed on the blockchain before sending more transactions from the same wallet.**

Thoughts?

ixje:
Yep I agree. I think a work around for users (which I didn't think of before) is to use the `wallet split` command.

PR #752, opened 2018-12-16
[refactor-prompt] Migrate command: wallet delete_addr

LysanderGG:
Migrate the command `wallet delete_addr` to the new Command system.

Implementation of `CommandWalletDeleteAddress`

Added unit tests & manual testing

Not a change but `UserWallet.DeleteAddress` always returns True so even when using `wallet delete_addr unknown_address`, the prompt output is `Deleted address unknown_address`.
If we want to change this behaviour, let's discuss the details.

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/tests/test_wallet_commands.py
+++ b/neo/Prompt/Commands/tests/test_wallet_commands.py
@@ -240,6 +240,32 @@ class UserWalletTestCase(WalletFixtureTestCase):
         self.assertEqual(type(res), UserWallet)
         self.assertEqual(len(res.Addresses), 9)
 
+    def test_wallet_delete_address(self):
+        # test wallet delete address with no wallet open
+        args = ['delete_addr']
+        res = CommandWallet().execute(args)
+        self.assertFalse(res)
+
+        self.OpenWallet()
+
+        # test wallet delete address with no argument
+        args = ['delete_addr']
+        res = CommandWallet().execute(args)
+        self.assertFalse(res)
+
+        # test wallet delete address with unknown address
+        args = ['delete_addr', self.watch_addr_str]
+        res = CommandWallet().execute(args)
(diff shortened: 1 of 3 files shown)


neo/Prompt/Commands/tests/test_wallet_commands.py
```diff
@@ -240,6 +240,32 @@ def test_wallet_create_address(self):
...
+
+        # test wallet delete address with no argument
+        args = ['delete_addr']
+        res = CommandWallet().execute(args)
+        self.assertFalse(res)
+
+        # test wallet delete address with unknown address
+        args = ['delete_addr', self.watch_addr_str]
+        res = CommandWallet().execute(args)
+        self.assertTrue(res)  # Current specs.
```
ixje: So as you noticed `DeleteAddress` _always_ returns `True`. Which can give confusing user feedback like the following for an address that's not actually in the wallet:
```
neo> wallet delete_addr AJN2SZJuF7j4mvKaMYAY9N8KsyD4j1fNdf
Deleted address AJN2SZJuF7j4mvKaMYAY9N8KsyD4j1fNdf
```

It's not really related to the migration but how the underlying `DeleteAdress` works. Would you like to fix it or shall I make it a separate issue?

LysanderGG: It's a small fix so I'll do it in this PR!

LysanderGG:
@ixje 
5078100 handles your comment here https://github.com/CityOfZion/neo-python/pull/752#discussion_r242148127

I decided to return `success` returned by `Wallet.DeleteAddress`, please tell me if this is not correct.

I also made sure that (invalid) too short or too long addresses do not result in an exception dumped to the prompt but a user friendly message.

ixje:
all good 👍

PR #754, opened 2018-12-17
Address MemoryStream leaks

ixje:
The usage of `ScriptBuilder` and `StreamManger` has a lot of potential to leak memory (and did in some places). 

Turned `ScriptBuilder` and `MemoryStream` (the class used by StreamManager internally) into context managers. Then used them as context managers everywhere (except in the tests, it was already a hell of a lot of typing).

I'm aware that `ApplicationEngine` and `ExecutionEngine` can also leak via the `ExecutionContext`. I need more time to test the impact/correct working when making those context managers.

make test

- [ ] Did you add any tests?


ixje:
Thanks for the review. I'll hold back merging for a while and monitor.

ixje:
will integrate this directly into the asyncio branch instead of up keeping this up to date.

PR #757, opened 2018-12-18
[refactor-prompt] Add wallet token base + delete

ixje:
Moves implementation of #623 forward by implementing the base command for `wallet token` and implementing the subcommand `wallet token delete`

Write code
write tests

- [X] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Tokens.py
+++ b/neo/Prompt/Commands/Tokens.py
@@ -2,10 +2,79 @@ from neo.Prompt.Commands.Invoke import InvokeContract, InvokeWithTokenVerificati
 from neo.Prompt.Utils import get_asset_id, get_from_addr, get_tx_attr_from_args
 from neo.Wallets.NEP5Token import NEP5Token
 from neocore.Fixed8 import Fixed8
+from neocore.UInt160 import UInt160
 from prompt_toolkit import prompt
 from decimal import Decimal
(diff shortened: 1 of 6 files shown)


neo/Prompt/Commands/Tokens.py
```diff
@@ -2,10 +2,79 @@
...
+
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        wallet = PromptData.Wallet
+
+        if len(arguments) != 1:
+            print("Please specify the required parameter")
+            return
```
LysanderGG: Wouldn't it be better to return `False` in early returns since `DeleteNEP5Token` returns a boolean?

neo/Prompt/Commands/tests/test_token_commands.py
```diff
@@ -522,6 +535,88 @@ def test_token_serialize(self):
...
+                self.assertIn("invalid parameter", mock_print.getvalue())
+
+            with patch('sys.stdout', new=StringIO()) as mock_print:
+                args = ['token', 'nope']
+                res = CommandWallet().execute(args)
+                self.assertFalse(res)
+                self.assertIn("invalid parameter", mock_print.getvalue())
+
+    def test_wallet_token_delete(self):
+        # test wallet alias with no wallet open
```
LysanderGG: `test wallet alias` 😅

ixje:
thanks for the quick review :)

PR #758, opened 2018-12-18
[refactor-prompt] Add wallet token send

ixje:
add `wallet token send` to move #623 forward

add command, refactor `token_send`

made tests, make test

yes, refactored `token_send` to parameterise it. Also it's now throwing exceptions instead of printing errors because this is a typical function that others would potentially like to re-use. Print statements do not allow others to easily handle invalid arguments.

- [X] Did you add any tests?

TODO:
 - [ ] remove `isValidPublicAddress()` helper function once a new `neo-python-core` release is out and update `requirements.txt`

### DIFF
--- a/neo/Prompt/Commands/Tokens.py
+++ b/neo/Prompt/Commands/Tokens.py
@@ -11,13 +11,32 @@ from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
 from neo.Prompt.PromptData import PromptData
 from neo.Prompt.Utils import get_arg
 from neo.Implementations.Wallets.peewee.Models import NEP5Token as ModelNEP5Token
+from neo.Core.TX.TransactionAttribute import TransactionAttributeUsage
 import peewee
+import base58
+
+
+# TODO: remove this once neo-python-core is merged
+def isValidPublicAddress(address: str) -> bool:
+    """Check if address is a valid NEO address"""
+    valid = False
+
+    if len(address) == 34 and address[0] == 'A':
+        try:
+            base58.b58decode_check(address.encode())
+            valid = True
+        except ValueError:
+            # checksum mismatch
+            valid = False
+
+    return valid
 
 
 class CommandWalletToken(CommandBase):
     def __init__(self):
         super().__init__()
         self.register_sub_command(CommandTokenDelete())
+        self.register_sub_command(CommandTokenSend())
 
     def command_desc(self):
         return CommandDesc('token', 'various token operations')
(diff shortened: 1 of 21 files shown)


jseagrave21:
I noticed that you are verifying the printouts. Should we have been doing that all along?

ixje:
I personally find it more reliable. I think past reviews have shown that merely looking at a `False` result doesn't always mean that the business logic is failing on what you expected.

jseagrave21:
That is true. Let me know in the reviews if you'd like to make these kind of tests standard. Also, if it is worth updating previous tests.

neo/Prompt/Commands/Tokens.py
```diff
@@ -77,23 +80,102 @@ def command_desc(self):
...
+        p2 = ParameterDesc('from_addr', 'address to send token from')
+        p3 = ParameterDesc('to_addr', 'address to send token to')
+        p4 = ParameterDesc('amount', 'number of tokens to send')
+        p5 = ParameterDesc('--tx-attr', f"a list of transaction attributes to attach to the transaction\n\n"
+        f"{' ':>17} See: http://docs.neo.org/en-us/network/network-protocol.html section 4 for a description of possible attributes\n\n"  # noqa: E128 ignore indentation
+        f"{' ':>17} Example:\n"
+        f"{' ':>20} --tx-attr=[{{\"usage\": <value>,\"data\":\"<remark>\"}}, ...]\n"
+        f"{' ':>20} --tx-attr=[{{\"usage\": 0x90,\"data\":\"my brief description\"}}]\n", optional=True)
+
+        return CommandDesc('send', 'remove a token from the wallet', [p1, p2, p3, p4, p5])
```
ixje: description to be fixed

neo/Prompt/Commands/Tokens.py
```diff
@@ -77,23 +80,102 @@ def command_desc(self):
...
+        _, user_tx_attributes = get_tx_attr_from_args(arguments)
+
+        token = arguments[0]
+        send_from = arguments[1]
+        send_to = arguments[2]
+        try:
+            amount = float(arguments[3])
+        except ValueError:
+            print(f"{arguments[3]} is not a valid amount")
+            return
```
LysanderGG: Should return `False` rather than `None` here imo

neo/Prompt/Commands/Tokens.py
```diff
@@ -77,23 +80,102 @@ def command_desc(self):
...
+    if not user_tx_attributes:
+        user_tx_attributes = []
+
+    token = None
+    for t in wallet.GetTokens().values():
+        if token_str == t.symbol:
+            token = t
+            break
+        elif token_str == t.ScriptHash.ToString():
+            token = t
```
LysanderGG: No break here?
Maybe `Utils.get_asset_id()` could be split to a `Utils.get_token()`.

ixje: I'll keep the split in mind. I've done some changes in my next PR that affects this one and therefore might not require it.

ixje:
> maybe we should do a second path on the already migrated commands sometime.
I don't understand what you mean with "second path". Perhaps it's a typo?

LysanderGG:
I meant "second pass" 😅 
Going back to the already implemented commands to improve the parameters checking, so this is not for this PR in particular.

ixje:
Are you referring to the repeated (pseudo):
```
if len(arguments) < count(mandatory args):
    error
if len(arguments) > count(mandatory + optional):
   error 
```

If so, I think we could put some helper function in the base class that iterates over the `ParamDesc` objects , counts the optional + mandatory args and boiler plates the above. I thought about it, but decided to skip for a later moment

LysanderGG:
I mean as general, your verification code here seems way more robust/user friendly than in most of the other commands.

> If so, I think we could put some helper function in the base class that iterates over the ParamDesc objects , counts the optional + mandatory args and boiler plates the above.

I did something similar some time ago but removed it because it was leading to some unreachable code
https://github.com/CityOfZion/neo-python/pull/733#discussion_r238490876

Let's add it to the todo and decide at the end.

Again this was not a problem about your PR but rather a good point that we should consider applying to the other commands 😄

neo/Prompt/Commands/Tokens.py
```diff
@@ -77,23 +80,103 @@ def command_desc(self):
...
+def token_send(wallet, token_str, send_from, send_to, amount, prompt_passwd=True, user_tx_attributes=None):
+    """
+
+    Args:
+        wallet (Wallet): a UserWallet instance
+        token_str (str): symbol name or script_hash
+        send_from (str): a wallet address
+        send_to (str): a wallet address
+        amount (float): the number of tokens to send
+        prompt_passwd (str): (optional) whether to prompt for a password before sending it to the network
```
LysanderGG: It's a boolean.

neo/Prompt/Commands/Tokens.py
```diff
@@ -77,23 +80,103 @@ def command_desc(self):
...
+        user_tx_attributes (list): a list of ``TransactionAttribute``s.
+
+    Returns:
+        a Transaction object if successful, False otherwise.
```
LysanderGG: I think it's None if it fails

neo/Prompt/Commands/Tokens.py
ixje: Good catch 👍

neo/Prompt/Commands/Tokens.py
LysanderGG: My bad!
I checked something and it was returning `None`, I don't know what I checked 🤔

PR #759, opened 2018-12-18
[refactor prompt] Add support for config group pt1

jseagrave21:
- Supports https://github.com/CityOfZion/neo-python/issues/623

- Adds support for the first five commands in the `config` group including, `output`, `sc-events`, `sc-debug-notify`, `vm-log`, and `node-requests`.

Trial and Error

Personal Testing and `make test`

No

- [x] Did you add any tests?


neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+                    print("Please specify a block request part and max size for each node, like 30 and 1000")
+                    return False
+            elif len(arguments) == 1:
+                return NodeLeader.Instance().setBlockReqSizeByName(arguments[0])
+        else:
+            print("Invalid number of arguments")
+            return False
+
+    def command_desc(self):
+        p1 = ParameterDesc('attribute', 'specify "slow"/"normal"/"fast" preset block request sizes, or specific block request size')
```
ixje: I understand why you used `attribute` but I think for this case `block-size` could work better. After reading the help output it just felt weird. 
```suggestion
         p1 = ParameterDesc('block-size', 'a preset of "slow"/"normal"/"fast", or a specific block request size (max. 500) e.g. 250 ')
```

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+        p2 = ParameterDesc('queue-size', 'specific max queue size. Required if custom block request size is specified', optional=True)
+        return CommandDesc('node-requests', 'configure the block request size', [p1, p2])
```
ixje: It is good that I never used this setting before because that made it clear to me that the descriptions where not self explanatory. I think `configure block request settings` would make more sense as a description because only one of the parameters is `block request size` the other is queue size related.
```suggestion
         return CommandDesc('node-requests', 'configure block request settings', [p1, p2])
```

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+                    return False
+            elif len(arguments) == 1:
+                return NodeLeader.Instance().setBlockReqSizeByName(arguments[0])
+        else:
+            print("Invalid number of arguments")
+            return False
+
+    def command_desc(self):
+        p1 = ParameterDesc('attribute', 'specify "slow"/"normal"/"fast" preset block request sizes, or specific block request size')
+        p2 = ParameterDesc('queue-size', 'specific max queue size. Required if custom block request size is specified', optional=True)
```
ixje: 2 things; 
1. the description isn't clear on what queue specifically, so a user will have a hard time understanding what changing this setting will have for an effect. I personally had to read the code to understand the purpose because I've not used this command before.
2. it's odd that the item is flagged as optional but the description says mandatory. 

I think we should 
1. change the description to something like `The maximum number of outstanding block requests`
2. we should change https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/Network/NodeLeader.py#L137-L143
to have default values equal to the normal settings (`100`/`10000`) and a maximum of 500 for `breqpart` (as pointed out in the NodeLeader comment above). 

In the most ideal situation we would probably have a custom help printer for this command indicating that only one of 2 params is mandatory. We could create an enhancement issue for this for a later point in time. For now I'd say just make them both mandatory. We could help to user by printing the current settings.

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+        p1 = ParameterDesc('attribute', 'either "on"|"off" or 1|0')
+        return CommandDesc('vm-log', 'toggle VM instruction execution logging on or off', [p1])
+
+
+class CommandConfigNodeRequests(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments):
+        if len(arguments) in [1, 2]:
```
ixje: Right now we can have the following output
```
neo> config node-requests -1 bla
Please specify a block request part and max size for each node, like 30 and 1000
neo> config node-requests -1 -1
[I 181219 09:30:07 NodeLeader:146] invalid values. Please specify a block request part and max size for each node, like 30 and 1000
```

I'd like to see uniform error messages

To solve this you can change
https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/Network/NodeLeader.py#L143
to raise a `ValueError` instead of printing a message, so you can act on it (and other interfaces maybe in the future).

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+
+class CommandConfigOutput(CommandBase):
+    def __init__(self):
+        super().__init__()
+
+    def execute(self, arguments=None):
+        return start_output_config()
+
+    def command_desc(self):
+        return CommandDesc('output', 'configure the output level settings')
```
ixje:
```suggestion
         return CommandDesc('output', 'configure the log output level settings')
```

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            else:
+                print("Cannot configure log. Please specify on|off")
+                return
+        else:
+            print("Cannot configure log. Please specify on|off")
+            return
+
+    def command_desc(self):
+        p1 = ParameterDesc('attribute', 'either "on"|"off" or 1|0')
+        return CommandDesc('sc-events', 'toggle smart contract events on or off', [p1])
```
ixje:
```suggestion
         return CommandDesc('sc-events', 'toggle printing smart contract events', [p1])
```
the possible options are given by the `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            if c1 == 'on' or c1 == '1':
+                settings.set_log_smart_contract_events(True)
+                print("Smart contract event logging is now enabled")
+                return c1
+            elif c1 == 'off' or c1 == '0':
+                settings.set_log_smart_contract_events(False)
+                print("Smart contract event logging is now disabled")
+                return c1
+            else:
+                print("Cannot configure log. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc` (1|0 missing)

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+                return
+        else:
+            print("Cannot configure log. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            if c1 == 'on' or c1 == '1':
+                settings.set_emit_notify_events_on_sc_execution_error(True)
+                print("Smart contract emit Notify events on execution failure is now enabled")
+                return c1
+            elif c1 == 'off' or c1 == '0':
+                settings.set_emit_notify_events_on_sc_execution_error(False)
+                print("Smart contract emit Notify events on execution failure is now disabled")
+                return c1
+            else:
+                print("Cannot configure log. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+                return
+        else:
+            print("Cannot configure log. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            return
+
+    def command_desc(self):
+        p1 = ParameterDesc('attribute', 'either "on"|"off" or 1|0')
+        return CommandDesc('sc-debug-notify', 'toggle Notify events on execution failure on or off', [p1])
```
ixje:
```suggestion
         return CommandDesc('sc-debug-notify', 'toggle printing Notify events on execution failure', [p1])
```

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            if c1 == 'on' or c1 == '1':
+                settings.set_log_vm_instruction(True)
+                print("VM instruction execution logging is now enabled")
+                return c1
+            elif c1 == 'off' or c1 == '0':
+                settings.set_log_vm_instruction(False)
+                print("VM instruction execution logging is now disabled")
+                return c1
+            else:
+                print("Cannot configure VM instruction logging. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+                return
+        else:
+            print("Cannot configure VM instruction logging. Please specify on|off")
```
ixje: error message options differs from `ParameterDesc`

neo/Prompt/Commands/Config.py
```diff
@@ -0,0 +1,205 @@
...
+            return
+
+    def command_desc(self):
+        p1 = ParameterDesc('attribute', 'either "on"|"off" or 1|0')
+        return CommandDesc('vm-log', 'toggle VM instruction execution logging on or off', [p1])
```
ixje:
```suggestion
         return CommandDesc('vm-log', 'toggle VM instruction execution logging to file', [p1])
```

neo/Prompt/Commands/Config.py
jseagrave21: I agree. This is a much better description. I couldn't think of any at the time. 👍

neo/Prompt/Commands/Config.py
jseagrave21: @ixje When would you print the current settings? Or would you add this as another option e.g. `config node-requests current`?

jseagrave21:
> See the comments for most things. One small remark regarding config.py. In the future please use git mv to rename a file instead of deleting and recreating. That way it will preserve the history. It's not a big deal now, but would be better in the future

Thank you for the feedback. I figured there was a better way but I didn't know 😊

neo/Prompt/Commands/Config.py
jseagrave21: I kept the original message, in case this function is used separately.

neo/Prompt/Commands/Config.py
ixje: Easiest is probably to add the following in `CommandConfigNodeRequests`
```
    def handle_help(self, arguments):
        super().handle_help(arguments)
        print(f"\nCurrent settings {self.command_desc().params[0].name}: {NodeLeader.Instance().BREQPART}  {self.command_desc().params[1].name}: {NodeLeader.Instance().BREQMAX}")
```

jseagrave21: That is awesome 😄

ixje:
Awesome! We're going fast like this :)

PR #760, opened 2018-12-19
[refactor-prompt] Migrate command: wallet unspent

LysanderGG:
Migrate the command `wallet unspent` to the new `Command` system.

Implementation of `CommandWalletUnspent`

Added unit tests & manual testing

The current behavior (branch developement) of `wallet unspent` is to have the asset as last parameter.
For instance:
- `wallet unspent neo --count` will ignore `neo` and count all unspent assets.
- `wallet unspent --count neo` will only count unspent `neo`

My implementation changes this to use `--` prefix for other arguments so the asset can be at any position.
I added `--from-addr` prefix instead of relying on the size of the argument for address detection.

Please let me know if any of these looks like a bad choice.

- [x] Did you add any tests?


LysanderGG:
@ixje Thank you for the review, I've implemented your comments.

> It would be nice if the user gets some feedback if there are no watch addresses. It's silent now which can make them wonder if something happened.

Note that it can also happen for known assets or addresses without unspent outputs.
For instance a new (empty) wallet will print out `No unspent assets matching the arguments.` for every valid argument combination.

ixje:
💯

PR #762, opened 2018-12-19
[refactor prompt] Add support for config maxpeers

jseagrave21:
- Supports https://github.com/CityOfZion/neo-python/issues/623
- Adds support for `config maxpeers`
- Improves functionality for maxpeers input which is useful if running an API server

Trial and Error

Manual Testing and `make test`

No

- [x] Did you add any tests?


jseagrave21:
@ixje Okay, this PR is ready for another review

PR #763, opened 2018-12-20
[refactor-prompt] add wallet token history

ixje:
move #623 forward by adding `token history`

port to new command

make test

Yes. When I added the `get_token` utility function in `neo.Prompt.Utils` I was facing circular import errors. To get around them I changed some import statements to not import functions directly, but import the module and call from there. 

- [X] Did you add any tests?

### DIFF
--- a/neo/Prompt/Utils.py
+++ b/neo/Prompt/Utils.py
@@ -10,6 +10,11 @@ from neocore.Cryptography.ECCurve import ECDSA
 from decimal import Decimal
 from prompt_toolkit.shortcuts import PromptSession
 from neo.logging import log_manager
+from neo.Wallets import NEP5Token
+from typing import TYPE_CHECKING
+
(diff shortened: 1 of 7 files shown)


neo/Prompt/Utils.py
```diff
@@ -334,3 +339,33 @@ def gather_param(index, param_type, do_continue=True):
...
+
+    Returns:
+        NEP5Token instance if found.
+    """
+    if token_str.startswith('0x'):
+        token_str = token_str[2:]
+
+    token = None
+    for t in wallet.GetTokens().values():
+        if token_str == t.symbol:
```
LysanderGG: Maybe `if token_str in [t.symbol, t.ScriptHash.ToString()]:` is a bit more elegant

ixje: Changed!

PR #764, opened 2018-12-20
[refactor-prompt] add wallet export commands

ixje:
move #623 forward by implementing `wallet export nep2` and `wallet export wif` commands.

write code

make test

Yes the old prompt required you to enter your wallet password multiple times before you could export. This was mostly a protection on the cli level but didn't really unlock anything as the wallet is already open with the keys useable. Now the prompt instantly exports without asking for input. 

- [X] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Wallet.py
+++ b/neo/Prompt/Commands/Wallet.py
@@ -316,7 +316,7 @@ class CommandWalletExportWIF(CommandBase):
             return False
 
         address = arguments[0]
-        keys = PromptData.Wallet.GetKeys()
+        keys = wallet.GetKeys()
         for key in keys:
             if key.GetAddress() == address:
                 print(f"WIF: {key.Export()}")
@@ -355,7 +355,7 @@ class CommandWalletExportNEP2(CommandBase):
             print("Please provide matching passwords")
             return False
 
-        keys = PromptData.Wallet.GetKeys()
+        keys = wallet.GetKeys()
         for key in keys:
             if key.GetAddress() == address:
                 print(f"NEP2: {key.ExportNEP2(passphrase)}")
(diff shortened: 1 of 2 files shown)


neo/Prompt/Commands/Wallet.py
```diff
@@ -279,6 +280,90 @@ def command_desc(self):
...
+            if key.GetAddress() == address:
+                print(f"NEP2: {key.ExportNEP2(passphrase)}")
+                return True
+        else:
+            print(f"Could not find address {address} in wallet")
+            return False
+
+    def command_desc(self):
+        p1 = ParameterDesc('address', 'public address in the wallet')
+        p2 = ParameterDesc('passphrase', 'secret to encrypt the private key with (min len: 10)')
```
ixje: I just realised that this parameter should be removed and that we should use a prompt. Otherwise the password will be stored in the prompt_log/history

neo/Prompt/Commands/Wallet.py
```diff
@@ -279,6 +280,95 @@ def command_desc(self):
...
+
+    def execute(self, arguments):
+        wallet = PromptData.Wallet
+
+        if len(arguments) != 1:
+            print("Please specify the required parameter")
+            return False
+
+        address = arguments[0]
+        keys = PromptData.Wallet.GetKeys()
```
LysanderGG: Could use `wallet` here or remove the definition of `wallet` local var which is unused.

neo/Prompt/Commands/Wallet.py
```diff
@@ -279,6 +280,95 @@ def command_desc(self):
...
+            print(f"Passphrase is too short, length: {len_pass}. Mininum length is 10")
+            return False
+
+        passphrase_confirm = prompt("[key password again] ", is_password=True)
+
+        if passphrase != passphrase_confirm:
+            print("Please provide matching passwords")
+            return False
+
+        keys = PromptData.Wallet.GetKeys()
```
LysanderGG: Could use `wallet` here or remove the definition of `wallet` local var which is unused.

ixje:
thanks, processed :)

ixje:
> A note for future, maybe subsub commands could be moved to another file, I feel like this file Wallet.py is becoming huge.

@LysanderGG agreed. I added a TODO to #623 for this

PR #767, opened 2018-12-21
[refactor-prompt] add import multisig_addr

ixje:
move #623 forward by adding `multisig_addr`

write code

write tests and make test

no

- [x] Did you add any tests?


ixje:
It did cross my mind to delete it, then I thought perhaps people might be using it. Then again, this whole PR is refactoring a lot anyway so people will just have to deal with breaking changes while we're not at v1.0. It's just needed to move forward.

PR #769, opened 2018-12-23
[refactor-prompt] add token approve and token allowance

ixje:
move #623 forward

migrate and refactor code

tests

no

- [X] Did you add any tests?


ixje:
Great timing. Now I can merge first and avoid conflicts when I add the last 2 commands in that group :)

PR #771, opened 2018-12-24
[refactor-prompt] add config nep8

ixje:
move #623 forward by

migrate

make test

no

- [X] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Config.py
+++ b/neo/Prompt/Commands/Config.py
@@ -4,6 +4,7 @@ from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
 from neo.Prompt.Utils import get_arg
 from neo.Settings import settings
 from neo.Network.NodeLeader import NodeLeader
+from distutils import util
 import logging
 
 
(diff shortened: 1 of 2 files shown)


jseagrave21:
Does it matter that this "toggle" function doesn't follow the same example as some others already approved like "vm-log" , "sc-debug-notify", etc?

ixje:
as in user feedback messages you mean? 
"Invalid option" vs "Cannot <feature>. Please specify on|off or 1|0"
or something else?

jseagrave21:
Yep that's what I meant

ixje:
Yeah I noticed and there's probably some more places where there are inconsistencies in naming/feedback used. I was planning to make a full sweep when all commands are in and align them. 

The cool thing about the `strtobool` function I recently found is that it parses a lot of different inputs.

https://docs.python.org/3.6/distutils/apiref.html#distutils.util.strtobool

I think it simplifies the code and gives more freedom to the user how they would like to express on/off.

jseagrave21:
> Yeah I noticed and there's probably some more places where there are inconsistencies in naming/feedback used. I was planning to make a full sweep when all commands are in and align them.
:+1: 
> The cool thing about the `strtobool` function I recently found is that it parses a lot of different inputs.
> 
> https://docs.python.org/3.6/distutils/apiref.html#distutils.util.strtobool
> 
> I think it simplifies the code and gives more freedom to the user how they would like to express on/off.

Very cool! Okay looks good to me.

ixje:
✅

PR #780, opened 2018-12-30
[refactor-prompt] adds --no-passwd-prompt flag for send and sendmany

jseagrave21:
- Supports @ixje's comment from https://github.com/CityOfZion/neo-python/pull/761#pullrequestreview-186642328. See https://github.com/CityOfZion/neo-python/pull/761#discussion_r243232432

- Adds `--no-passwd-prompt` flag for `send` and `sendmany`

- Added a new utility `get_passwd_prompt`
  - @ixje could you consider implementing this flag instead of specific booleans at certain indices? I think it will make the code more flexible. See https://github.com/ixje/neo-python/blob/da827333089bef9d87c46daa7ae47b2686c05aca/neo/Prompt/Commands/Tokens.py#L172-L178

Personal testing and unittest

No

- [x] Did you add any tests?  **updated redundant tests**


LysanderGG:
Not sure about the argument in `CommandTokenSendFrom` either now that I thought about it 😅

jseagrave21:
I don't think it is a bad idea. If the option exists in the code I think it should be available to the user (for the most part). Developer-only features will require additional extra documentation.

LysanderGG:
It means that if someone gets access to the prompt with an open wallet he can send funds without password confirmation at all

jseagrave21:
This is why I originally completely removed the feature.

jseagrave21:
Based on this comment https://github.com/CityOfZion/neo-python/pull/791#issuecomment-451635329, please let me know if there is any reason to keep this open, or I can just close it.

PR #782, opened 2018-12-31
[refactor-prompt] Cleanup 1

ixje:
There are a couple of clean up todo's in #623 
This addresses the following 2:
```
- Move wallet commands related to address in its own address group ( alias, create, delete, split)
- Wallet is becoming huge. Split file into multiple (e.g. based on subgroup)
```

moved business logic + tests of `wallet import`,  `wallet export` and `address` related commands into their own files and subgroup.

make test

no

- [ ] Did you add any tests?


ixje:
don't want to do too much in one pr when possible.

PR #783, opened 2018-12-31
[refactor-prompt] cleanup 2

ixje:
1) some group commands had the error message 
```
Please specify an action. See help for available actions
```
the other half
```
run `<cmd name> help`   to see supported queries
```

2) A good part of the commands used
```
Please specify the required parameters
```
others use slightly more specific

```
Please specify the <missing argument name>
```
and then some had completely different responses like

```
invalid number of   arguments
Not enough   arguments
etc
```

3) some commands that toggle settings accepted either `1` or `on`, others accepted more boolean values like `t`, `true` etc.

streamline responses amongst commands for certain actions. see commits

make test

no

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Config.py
+++ b/neo/Prompt/Commands/Config.py
@@ -27,7 +27,7 @@ class CommandConfig(CommandBase):
         item = get_arg(arguments)
 
         if not item:
-            print("run `%s help` to see supported queries" % self.command_desc().command)
+            print(f"run `{self.command_desc().command} help` to see supported queries")
             return
 
         try:
@@ -101,7 +101,7 @@ class CommandConfigDebugNotify(CommandBase):
 
     def command_desc(self):
         p1 = ParameterDesc('attribute', 'either "on"|"off" or 1|0')
-        return CommandDesc('sc-debug-notify', 'toggle printing Notify events on execution failure', [p1])
+        return CommandDesc('sc-debug-notify', 'toggle printing smart contract Notify events on execution failure', [p1])
 
 
 class CommandConfigVMLog(CommandBase):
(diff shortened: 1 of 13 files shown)


neo/Prompt/Commands/Config.py
```diff
@@ -53,23 +53,23 @@ def __init__(self):
...
+            return False
+
+        if flag:
+            print("Smart contract event logging is now enabled")
+            return True
         else:
-            print("Cannot configure log. Please specify on|off or 1|0")
-            return
+            print("Smart contract event logging is now disabled")
+            return False
```
jseagrave21: Referring to follow-on commands, I think this should be `return True` since the command is successfully disabling the SC event logging. See https://github.com/CityOfZion/neo-python/pull/783/files#diff-799c82d84fa184b364281a4e3f88141cR100 and https://github.com/CityOfZion/neo-python/pull/783/files#diff-799c82d84fa184b364281a4e3f88141cR128

neo/Prompt/Commands/tests/test_config_commands.py
```diff
@@ -49,7 +49,7 @@ def test_config_sc_events(self):
         # test turning them off
         args = ['sc-events', '0']
         res = CommandConfig().execute(args)
-        self.assertTrue(res)
+        self.assertFalse(settings.log_smart_contract_events)
```
jseagrave21: Similarly to other config commands, I think this should still include the original `self.assertTrue(res)`. The additional change could be included also if desired.

neo/Prompt/Commands/Config.py
ixje: good catch! 💪

neo/Prompt/Commands/tests/test_config_commands.py
ixje: I've added this to the other commands as well (with the exception of NEP8)

PR #786, opened 2019-01-03
[refactor-prompt] add sc invoke

ixje:
move #623 forward. Add the final command `sc invoke` (formerly known as `testinvoke`)

port code

make test

no

- [X] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/SC.py
+++ b/neo/Prompt/Commands/SC.py
@@ -1,12 +1,15 @@
 from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
 from neo.Prompt.PromptData import PromptData
-from neo.Prompt.Utils import get_arg
+from neo.Prompt import Utils as PromptUtils
 from neo.Prompt.Commands.BuildNRun import Build, BuildAndRun, LoadAndRun
+from neo.Prompt.Commands.Invoke import TestInvokeContract, InvokeContract
 from neo.Core.Blockchain import Blockchain
+from neocore.UInt160 import UInt160
+from neocore.Fixed8 import Fixed8
+from prompt_toolkit import prompt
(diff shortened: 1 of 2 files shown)


neo/Prompt/Commands/SC.py
```diff
@@ -127,11 +132,83 @@ def command_desc(self):
...
+        hash_string = arguments[0]
+        try:
+            script_hash = UInt160.ParseString(hash_string)
+        except Exception:
+            # because UInt160 throws a generic exception. Should be fixed in the future
+            print("Invalid script hash")
+            return False
+
+        tx, fee, results, num_ops = TestInvokeContract(wallet, arguments, from_addr=from_addr, invoke_attrs=invoke_attrs, owners=owners)
+        if tx is not None and results is not None:
```
LysanderGG: Usual `if tx and results:` 😄

neo/Prompt/Commands/SC.py
```diff
@@ -127,11 +132,83 @@ def command_desc(self):
...
+
+            parameterized_results = [ContractParameter.ToParameter(item) for item in results]
```
LysanderGG: `parameterized_results` is only used in the debug string, maybe it would be better to call `ToJson()` in this comprehension.

It would give:
`parameterized_results = [ContractParameter.ToParameter(item).ToJson() for item in results]`

What do you think?

neo/Prompt/Commands/SC.py
ixje: how could I have missed that 🤦‍♂️ , thanks!

neo/Prompt/Commands/SC.py
ixje: sounds good

Issue #624, opened 2018-09-20
Create a more interactive and clean CLI


ixje:
_Rationale:_
The current CLI can be described as a vertical scrolling log where the bottom line is interactive and allows for command inputs that get displayed in the log. While it gives a quick historical overview, it is not suited for interactive inspection of data such as Blocks and Transactions. 

Event logging is interlaced with command history and can clutter the view. Error messages can even disrupt the proper representation of the statusbar (see below). In short, there is enough room for improvement 
![console](https://i.imgur.com/YuFyNLE.png)

_Goal:_
Create a clean interactive CLI. The [mitmproxy](https://mitmproxy.org/ ) CLI is a great example how interactive and clean a console application can be. It uses the [urwid](http://urwid.org/) library for the heavy lifting (beware it has a steep learning curve).

CLI wishes:

- Allow for interactive block / transaction inspection. 
- A minimal example of how that could look can be see [here](https://cdn.discordapp.com/attachments/410364599133667328/469956379059159042/neo-python-console.mov)
- Work on all platforms (Windows/Linux/OSX). The Windows requirement can be the usage of Cygwin or a proper capable terminal.
- Separate log messages from command and block history (e.g. see how mitxproxy has a splitpane capability)
- Have plugin based command support (see the related issue: [Refactor prompt commands to be plugin based](https://github.com/CityOfZion/neo-python/issues/623))

ixje:
I gave a moment of thought about how we could group commands.

```
quit
help

show
- block {index/hash} (tx)
- header {index/hash}
- tx {hash}
- mem # returns memory in use and number of buffers
- nodes # returns connected peers
- state
- notifications {block_number or address}
- account {address} # returns account state
- asset
-- {assetId} # returns asset state
-- search {query} <- not ideal. open for ideas how to structure otherwise.
- contract 
-- {contract hash} # returns contract state
-- search {query}


config 
- debug {on/off}
- sc-events {on/off}
- maxpeers {num_peers}
- node-requests {reqsize} {queuesize}
- node-requests {slow/normal/fast}
- compiler-nep8 {on/off}

wallet
- create
- open 
- close
- {verbose}
- claim (max_coins_to_claim)
- migrate
- rebuild {start block}
- create_addr {number of addresses <= 3}
- delete_addr {addr}
- delete_token {token_contract_hash}
- alias {addr} {title}
- tkn_send {token symbol} {address_from} {address to} {amount}
[...]

PR #788, opened 2019-01-04
[refactor-prompt] restore theming support

ixje:
resolve the following todo of #623 
> after the previous item, make sure console themes still work. This requires "Format Tokens" from the prompt_toolkit package, but it should only be applied when used in the CLI. Meaning; the business logic should not be coupled to a PromptInterface() instance because that's the one who creates the theme tokens. solution idea: supply print function as param.

replace built-in print with `prompt_print` in the classes below `Commands` as follows:
```
from neo.Prompt.PromptPrinter import prompt_print as print
```
This was the most straight forward way to replace all `print` calls in command code. One advantage is that we can easily swap out the `print` call for something custom. One place we do this is for testing where we swap out the `prompt_toolkit.print_formatted_text` with the buildint `print` such that all our `patch('sys.stdout')` in test-cases still work.

In the old prompt some part of the command execution would be themed (e.g. red coloured) whereas others not because they used regular prints. Now everything related to a command response is printed themed.

make test and some manual testing

see the "how solved" section. 

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/tests/test_config_commands.py
+++ b/neo/Prompt/Commands/tests/test_config_commands.py
@@ -36,7 +36,7 @@ class CommandConfigTestCase(BlockchainFixtureTestCase):
         self.assertFalse(res)
 
     def test_config_output(self):
-        # importing because it setups `peewee` logging, which is checked at the test below
+        # importing because it sets up `peewee` logging, which is checked at the test below
         from neo.Implementations.Wallets.peewee.UserWallet import UserWallet
 
         args = ['output']
--- a/neo/Prompt/Commands/tests/test_config_commands.py
+++ b/neo/Prompt/Commands/tests/test_config_commands.py
@@ -6,9 +6,20 @@ from copy import deepcopy
 from neo.Network.NodeLeader import NodeLeader
 from mock import patch
 from io import StringIO
+from neo.Prompt.PromptPrinter import pp
 
 
 class CommandConfigTestCase(BlockchainFixtureTestCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        # replace the prompt_toolkit formatted print function with the default such that we can test easily
+        pp.printer = print
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        pp.reset_printer()
 
     @classmethod
     def leveldb_testpath(self):
@@ -25,6 +36,9 @@ class CommandConfigTestCase(BlockchainFixtureTestCase):
         self.assertFalse(res)
 
     def test_config_output(self):
+        # importing because it setups `peewee` logging, which is checked at the test below
+        from neo.Implementations.Wallets.peewee.UserWallet import UserWallet
+
         args = ['output']
         with patch('neo.Prompt.Commands.Config.prompt', side_effect=[1, 1, 1, "\n", "\n"]):  # tests changing the level and keeping the current level
             res = CommandConfig().execute(args)
--- a/neo/Prompt/CommandBase.py
+++ b/neo/Prompt/CommandBase.py
@@ -1,6 +1,7 @@
 from abc import ABC, abstractmethod
 from neo.Prompt.Utils import get_arg
 from typing import List
+from neo.Prompt.PromptPrinter import prompt_print as print
 
 
 class ParameterDesc():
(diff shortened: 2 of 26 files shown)


neo/Prompt/Commands/tests/test_config_commands.py
```diff
@@ -25,6 +36,9 @@ def test_config(self):
         self.assertFalse(res)
 
     def test_config_output(self):
+        # importing because it setups `peewee` logging, which is checked at the test below
```
jseagrave21: Should be "sets up" not "setups"

ixje:
> I don't think the output from show state should look how it currently does because of the default command text formatting:

What would you have in mind it should look like?

My thoughts on the current approach were; 
there are 2 types of information being printed on the console. 1) information triggered by us and information triggered in the background by networking code, VM execution etc. The former could be limited to `cli command responses` and are (currently) in red, the latter are in whatever colour the logger (debug,info,error) prints them. This gives a clear distinction for the user what they caused and what not.

I know it's a bit limited in the presentation, but I think that can be handled/improved with https://github.com/CityOfZion/neo-python/issues/624

ixje:
> A final separate issue I noticed is that the wallet command does not have an adequate help description. I recommend changing command_desc in CommandWallet to:

we can't do that because that will give
```
neo> help

Commands:
   config          - configure internal settings
   sc              - develop smart contracts
   search          - search the blockchain
   show            - show various node and blockchain data
   wallet          - show wallet details or manage wallets

                     Usage Examples:
                        wallet
                        wallet COMMAND
```
if we ever get a base group command with e.g. a `z` then you'd get something 
```
neo> help

Commands:
   config          - configure internal settings
   sc              - develop smart contracts
   search          - search the blockchain
   show            - show various node and blockchain data
   wallet          - show wallet details or manage wallets

                     Usage Examples:
                        wallet
                        wallet COMMAND
    z-group        - my custom command
```
Not too pretty imo. The best I can come up with is
```
neo> wallet help

Usage: wallet COMMAND (or "wallet" to show the wallet contents)
```

This requires changing all double underscore methods/attributes in `CommandBase` to single underscore such that we can override `handle_help()` for `CommandWallet`. I'll ping Lysander on slack to see if he has ideas (can't tag him here)

ixje:
addressed the `show state help` issue in a separate PR #792

LysanderGG:
Reply to https://github.com/CityOfZion/neo-python/pull/788#issuecomment-451639268

I can think of a few solutions:

1. have one (or several) subfunctions returning a string that `handle_help` would print, instead of directly printing.
For this particular case, we could have:

```python
def _usage_str(self):
    return f"Usage: {self.__command_with_parents()} COMMAND"

def handle_help(self, arguments):
        item = get_arg(arguments)
        if item == 'help':
            if len(self.__sub_commands) > 0:
                # show overview of subcommands and their purpose
                print(f"\n{self._usage_str()}\n")
                print(f"{self.command_desc().short_help.capitalize()}\n")
                ...
```

Then overriding `_usage_str` for `CommandWallet` should handle this problem.


2. We could restore the original idea of having a short help message and a longer help message.
If the long help message is not defined, it would be the same as the short one.

The longer help could be used outside of listings.
For this particular problem, we would only need to provide a long help for `CommandWallet`.


3. Almost same as 2. but instead of adding a member to `CommandDesc`, we could define the following command that could be overwritten:

```
def longer_help(self):
    return self.command_desc().short_help
```


I don't think one is better than another for now 🤔 
Maybe 1. is good enough for now and simple enough.

neo/Prompt/Commands/tests/test_config_commands.py
jseagrave21: Hey no worries! I only wish I could be of more help. It seems like you made a big push the other day!

jseagrave21:
> What would you have in mind it should look like?

I was just referring back to the original presentation, where most outputs were in "number" format. In this case, I think the thinking was there were three types instead of two: input, ouput, and other. So all commands, seen when using `help`, would be red since they are commmands, but all outputs would be in "number" format, except for unrequested outputs like you mentioned (e.g. errors, etc.) .

> I know it's a bit limited in the presentation, but I think that can be handled/improved with #624

Maybe it is better to table this discussion til then?

PR #791, opened 2019-01-04
[refactor-prompt] enforce pw prompt send

ixje:
In #780 (and another PR) we discussed if we should be prompted for the wallet password before we can send funds. The question has been posted on slack and after some thought I think that at least from the cli it should be asked for. 

Previously we could bypass it using a `--ask_for_pw` type of flag. While the logic remains in some parts of the code (so people can programmatically choose), from the CLI it is no longer possible to bypass for security reasons.

remove optional arg to bypass

make test

no

- [X] Did you add any tests?


ixje:
I think not, sorry. It took some input and more dabbling on the options before I got to this conclusion.

PR #793, opened 2019-01-05
[refactor-prompt] polish and clarify insufficient funds reason

ixje:
Currently whenever you try to send assets with insufficient funds you get the following output
```
neo> wallet send NEO AG4Z3kNtdQxDvdCPYyht1Mq8ntdi7uHuvy 25
[E 190105 12:16:44 Wallet:1071] insufficient funds for asset id: c56f33fc6ecfcd0c225c4ab356fee59390af8560be0e930faebe74a6daff7c9b
[D 190105 12:16:44 Send:233] insufficient funds
neo>
```
A double error message, and also one that doesn't always clearly identify the real reason. The above account had more than enough funds, but not enough `vouts`
 
Changed the error handling of `MakeTransaction` such that we can differentiate the root cause. The above scenario now shows
```
neo> wallet send NEO AG4Z3kNtdQxDvdCPYyht1Mq8ntdi7uHuvy 25
[Password]> **********
[I 190105 13:32:00 Transaction:619] Verifying transaction: b'73d8a0106e8c2b9f2a4d6a7733fe57a534ec5968ba3d6ba67dffb253cb21a66e'
Relayed Tx: 73d8a0106e8c2b9f2a4d6a7733fe57a534ec5968ba3d6ba67dffb253cb21a66e
neo> wallet send NEO AG4Z3kNtdQxDvdCPYyht1Mq8ntdi7uHuvy 25
Insufficient funds. No unspent outputs available for building the transaction.
If you are trying to sent multiple transactions in 1 block, then make sure you have enough 'vouts'.
Use `wallet unspent` and `wallet address split`, or wait until the first transaction is processed before sending another.
neo>
```

manual testing

no

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Send.py
+++ b/neo/Prompt/Commands/Send.py
@@ -228,11 +228,17 @@ def process_transaction(wallet, contract_tx, scripthash_from=None, scripthash_ch
                                     change_address=scripthash_change,
                                     fee=fee,
                                     from_addr=scripthash_from)
+    except ValueError:
+        print("Insufficient funds. No unspent outputs available for building the transaction.\n"
+              "If you are trying to sent multiple transactions in 1 block, then make sure you have enough 'vouts'\n."
+              "Use `wallet unspent` and `wallet address split`, or wait until the first transaction is processed before sending another.")
+        return None
 
-        if tx is None:
-            logger.debug("insufficient funds")
-            return None
+    if tx is None:
+        logger.debug("insufficient funds")
+        return None
 
+    try:
         # password prompt
         passwd = prompt("[Password]> ", is_password=True)
         if not wallet.ValidatePassword(passwd):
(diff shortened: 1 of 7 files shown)


neo/Prompt/Commands/Send.py
```diff
@@ -228,11 +228,17 @@ def process_transaction(wallet, contract_tx, scripthash_from=None, scripthash_ch
...
+        print("Insufficient funds. No unspent outputs available for building the transaction.\n"
+              "If you are trying to sent multiple transactions in 1 block, then make sure you have enough 'vouts'\n."
+              "Use `wallet unspent` and `wallet address split`, or wait until the first transaction is processed before sending another.")
+        return None
 
-        if tx is None:
-            logger.debug("insufficient funds")
-            return None
+    if tx is None:
+        logger.debug("insufficient funds")
```
jseagrave21: At this point, if there really is insufficient funds (i.e. the user tries to send more than the wallet balance), won't it not display, since the users doesn't usually see DEBUG info?

ixje: nah, because real insufficient funds will also give the long message. Only if the wallet is not synced it will hit that, but then it has a `logger.warning` before that to indicate the wallet isn't synced.

PR #799, opened 2019-01-07
[refactor-prompt] Improve send & sendmany (third try)

jseagrave21:
- Allows for KeyboardInterrupt during prompting in `sendmany`
- Prints the fee associated with the transaction when using `send`
- Cleans up a couple outputs and verifies outputs during testing
- Replaces `return None` with `return` which is in line with other refactoring
- retitles `test_send_command.py`  as `test_send_commands.py` to match others

**update**
I re-created this PR to try to change the title of `test_send_command.py` in a way that would still allow for a review based on feedback in https://github.com/CityOfZion/neo-python/pull/797.
Also, you can refer to https://github.com/CityOfZion/neo-python/pull/797#pullrequestreview-189654277 for @LysanderGG 's original review.

Trial and Error

Manual Testing and unittest

No

- [x] Did you add any tests?


ixje:
thanks!

PR #804, opened 2019-01-08
Refactor prompt

ixje:
**What current issue(s) does this address, or what feature is it adding?**
Address #623 
Huge thanks @jseagrave21 and @LysanderGG

**How did you solve this problem?**
see #623

**How did you make sure your solution works?**
make test

**Are there any special changes in the code that we should be aware of?**
I've chosen to not include #731 at this time. Mostly because I want this hard work in the development branch such that people can test it for a couple of days before we do a new release. The module loading part can be addresses at a later point. For that I want to spent a little bit more time to see how we can also support targeted loading (meaning; load the custom cmd into a specific existing sub group)

**Please check the following, if applicable:**

- [X] Did you add any tests?
- [X] Did you run `make lint`?
- [X] Did you run `make test`?
- [X] Are you making a PR to a feature branch or development rather than master?
- [X] Did you add an entry to `CHANGELOG.rst`? (if not, please do)

PR #805, opened 2019-01-09
refactor-prompt

ixje:
This is a repost of #804 via my personal github repo. #804 seems to have build issues while this PR continuously builds fine. 

Address #623 
Huge thanks @jseagrave21 and @LysanderGG

see #623

make test

I've chosen to not include #731 at this time. Mostly because I want this hard work in the development branch such that people can test it for a couple of days before we do a new release. The module loading part can be addresses at a later point. For that I want to spent a little bit more time to see how we can also support targeted loading (meaning; load the custom cmd into a specific existing sub group)

- [X] Did you add any tests?


jseagrave21:
:tada:

PR #723, opened 2018-11-20
Network stability improvements

ixje:
**What current issue(s) does this address, or what feature is it adding?**
Addresses #325 and #492 
There's a couple of issues with the current networking code that I'll only describe at a high level:
- nodes get stuck/stop syncing at some point
- debugging output is insufficient to investigate the reason of poor stability

**How did you solve this problem?**
- Added health checking of peers based on block/header requests vs actual responses within a given time threshold. Bad peers are disconnected from and moved into the `DEAD_ADDRS` list.
- Added monitoring of connected peers and if we reach 0 we reset and queue up peers. Address selection is sorted from "never connected to" to "recently disconnected from" so we always try to pick a fresh one.
- Added data throttling when our request speed exceeds a certain response speed threshold
- Move some threading calls into their proper threads (mainly into the reactor) to avoid unexpected behaviour.
- Improved the nodes output to give a bit more information on the clients and addresses. i.e.
```
neo> nodes
Total Connected: 2
Peer 0  /NEO:2.8.0/ -  18.222.168.189:20333 - IO   0.323453 MB in /   0.009913 MB out
Peer 1  /Neo:2.9.0/ -    47.52.168.73:20333 - IO   0.131664 MB in /   0.006687 MB out

KNOWN_ADDRS: 174
DEAD_ADDRS: 6
QUEUED_ADDRS: 3
```

- Added a lot of debug output and nicely format them to make analysis easier in the future.
- Added 2 "hidden" commands `disconnect_node <id>` and `handles` for debugging purposes. They do not show up in the `help` output of the prompt, but can be used.
- Fixed an issue where block persisting would pass, but the notificationDB fails emitting the event which caused the BlockPersist task to cancel and the node gets 'stuck'. 

**How did you make sure your solution works?**
For the past 2 weeks I've been monitoring and adjusting this daily. 

**Are there any special changes in the code that we should be aware of?**
1) @localhuman I'd say this one is specifically for you because it's notificationDB related which I think you work a lot with? Block `1231519` has one 'malformed' transaction that throws an exception in the NotificationDB when serialising a `Neo.Contract.Create` event.   

2) test-cases have not been finalised. I've skipped 2 because I wanted to get this PR out so people can start testing if they want.

**Known issues**
1) minor - `DEAD_ADDRS` can contain duplicates. This has little effect on usage or stability.
[...]

PR #809, opened 2019-01-13
properly reset NodeLeader for test cases

ixje:
A common root cause for tests failing to relay is because NodeLeader isn't reset properly. Other tests could have created a similar TX which is now stored in the `KnownHashes` and therefore `Relay` fails.

add reset logic

make test and faced this issue again when merging `development` into #723 

no

- [ ] Did you add any tests?


ixje:
at this point it builds fine, so I'd say no.

I don't know what is going on with Travis but it unexpectedly fails left and right where it never failed before and where it does not fail locally. It's getting really frustrating, so I want to touch the least possible at this point.

Issue #84, opened 2017-11-12
Idea: add pyup.io


metachris:
pyup.io is a service that helps you to keep your requirements files up to date. It sends you automated pull requests whenever there’s a new release for one of your dependencies.

https://pyup.io

ixje:
@metachris ping :)

LysanderGG:
I checked out pyup.io (see #819)
@ixje do you have the necessary credentials to set-it up now?

LysanderGG:
Note: we got a lot of merge requests to `master` because it is the current default branch.
Imo the default branch should be changed to `development` (merge requests will also default to merges to `development`).

ixje:
Closing due to focus on NEO3

PR #819, opened 2019-01-16
Add .pyup.yml

LysanderGG:
#84
I tried out pyup.io and thought it would indeed be good.
Here is a suggestion of config file to have one PR a week to update dependencies.

Here is what the PR looks like https://github.com/LysanderGG/neo-python/pull/50

Test on my local fork

No

- [ ] Did you add any tests?


ixje:
👍

PR #858, opened 2019-01-19
Update .pyup.yml

LysanderGG:
Explicitly set the branch to merge against

.

.

No.
However, I think that this should be merged to master in order for pyup to read it.

- [ ] Did you add any tests?


ixje:
looks good. On a side note; any idea why the "weekly" settings might not have worked. I closed ~30 individual PR's

LysanderGG:
The file is only on develop, as long as it is not on master it won't be taken into account I think.

ixje:
d'oh, totally overlooked that. I thought it was on master. Can you change the destination branch?

ixje:
Thanks. I'll re-enable pyup shortly 🤞

PR #860, opened 2019-01-19
Fix ImportMultiSigAddr

jseagrave21:
Currently in v0.8.3 if you import a multi-sig addr and you want to require 2 signatures including your own public key, it is not possible.

Example:
```
neo> wallet import multisig_addr <own pub key> 2 <key 2>
Missing remaining signing keys. Minimum required: 2 given: 1
neo> wallet import multisig_addr <own pub key> 2 <own pub key> <key 2>
Provided signing keys are not unique
```
forcing me to do
```
neo> wallet import multisig_addr <own pub key> 1 <key 2>
Added multi-sig contract address <addr> to wallet
```
but now this multsig address only requires one signature to send

The new function assumes own pub will be a signing key so I added 1 to the `len` of signing keys

Personal testing and unittest

No

- [x] Did you add any tests? (updated)


CHANGELOG.rst
```diff
@@ -3,6 +3,10 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+[0.8.4] in progress
+----------------------
+- Fix ``ImportMultiSigAddr``
```
ixje:
```suggestion
- Fix incorrect error handling when importing a 1 out of 2 multi-signature address
```

neo/Prompt/Commands/WalletImport.py
```diff
@@ -181,7 +181,7 @@ def execute(self, arguments):
 
         # validate minimum required signing key count
         signing_keys = arguments[2:]
-        len_signing_keys = len(signing_keys)
+        len_signing_keys = len(signing_keys) + 1  # since own pub key will be a signing key
```
ixje: :thumbsup:
Alternatively we could have moved this line up
https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Prompt/Commands/WalletImport.py#L197
and in between 
```
    signing_keys = arguments[2:]
    len_signing_keys = len(signing_keys)
```

jseagrave21: Oh, I think this makes more sense! Thank you for the suggestion! Then I don't have to comment on something that will happen in the future.

ixje:
50 point jseagrave

lllwvlvwlll:
https://neoscan.io/transaction/b6b1fe325916d225a08df83e9070650d43d57b515ad31fffa39c0c1879c6f32b

PR #869, opened 2019-01-25
Add ability to attach a fee when deploying a smart contract

jseagrave21:
- See title

This PR is a result of a discussion with @belane who referenced limits for tx size and fee requirements which can be found below:
- https://github.com/neo-project/neo-plugins/blob/d35297618006bc0dfdae3bacab22e98df682d84b/SimplePolicy/SimplePolicyPlugin.cs#L99
- https://github.com/neo-project/neo-plugins/blob/master/SimplePolicy/SimplePolicy/config.json
- https://github.com/neo-project/neo/blob/a8fd76b715e675bc7982f26481ecfb9b6ea68811/neo/Settings.cs

A user may add a fee when deploying a smart contract but more importantly,
now if `tx.Size()` is larger than the `MaxFreeTransactionSize` (1024), the required fee is calculated with a minimum set at the `LowPriorityThreshold` (0.001 GAS).

Unittest

I added the referenced values above to Settings. I am not sure if this is the best place for them.

- [x] Did you add any tests?


jseagrave21:
@belane Could you please review this PR and confirm the functionality is as desired? You should be able to use the new tests.

belane:
The logic seems fine, but neo-python command line changed a bit since last time I saw it.
Maybe is better to move the logic to InvokeContract or even to Transacction because this has to be fulfilled for all transactions.

@ixje what do you think?

jseagrave21:
@belane @ixje if this check needs to be applied to every transaction, one option may be to incorporate it into `Wallet.MakeTransaction`.

xxedocxx:
The current master and dev branch couldn't deploy my smart contract because it was more than 1024 bytes. I could deploy on this branch. Best @jseagrave21

neo/Prompt/Commands/SC.py
```diff
@@ -299,8 +318,8 @@ def command_desc(self):
                            f"the return type of the smart contract output\n\n"
                            f"{' ':>17} For more information about parameter types see\n"
                            f"{' ':>17} https://neo-python.readthedocs.io/en/latest/data-types.html#contractparametertypes\n", optional=True)
-
-        params = [p1, p2, p3, p4, p5, p6]
+        p7 = ParameterDesc('--fee', 'fee to give your transaction priority (> 0.001) e.g. --fee=0.01', optional=True)
```
ixje:
```suggestion
        p7 = ParameterDesc('--fee', 'Attach GAS amount to give your transaction priority (> 0.001) e.g. --fee=0.01', optional=True)
```
just saying "fee" doesn't tell the user where the fee is coming from. They have to derive it's GAS from the fact that NEO is indivisible. This suggestion clarifies it.

neo/Prompt/Commands/SC.py
```diff
@@ -283,7 +284,25 @@ def execute(self, arguments):
                 print("Incorrect password")
                 return False
 
-            return InvokeContract(wallet, tx, Fixed8.Zero(), from_addr=from_addr)
+            n_fee = Fixed8.Zero()
```
ixje: General remark for the code below. The information is printed _after_ we informed the user what their fees would be (from the testinvoke). This means when they enter their password they can get unexpected costs which we should avoid.

neo/Prompt/Commands/SC.py
```diff
@@ -283,7 +284,25 @@ def execute(self, arguments):
                 print("Incorrect password")
                 return False
 
-            return InvokeContract(wallet, tx, Fixed8.Zero(), from_addr=from_addr)
+            n_fee = Fixed8.Zero()
+            if priority_fee is not None:
+                n_fee = priority_fee
+                if n_fee is False:
+                    logger.debug("invalid fee")
+                    return False
```
ixje: 2 things
1) if the user does not have logging set to debugging then their deploy will fail silently after passw entry
2) we should capture invalid arguments before test running.

neo/Prompt/Commands/SC.py
```diff
@@ -283,7 +284,25 @@ def execute(self, arguments):
...
+
+            # calculate the required network fee for the tx and compare to n_fee
+            if tx.Size() > settings.MAX_FREE_TX_SIZE:
+                req_fee = Fixed8.FromDecimal(settings.FEE_PER_EXTRA_BYTE * (tx.Size() - settings.MAX_FREE_TX_SIZE))
+                if req_fee < settings.LOW_PRIORITY_THRESHOLD:
+                    req_fee = settings.LOW_PRIORITY_THRESHOLD
+                if n_fee < req_fee:
+                    print(f'Transaction cancelled. The tx size ({tx.Size()}) exceeds the max free tx size ({settings.MAX_FREE_TX_SIZE}).')
+                    print(f'A network fee of {req_fee.ToString()} GAS is required.')
+                    return False
```
ixje: The overall checks here look good to me. We have to think about a more generic place to put this stuff though. Your suggestion of `wallet.MakeTransaction` makes sense as that appears (note: has to be validated) to be called before all type of transactions (Invoke, Claim etc). 

If we put it there we need to 
- excluded the check for `ClaimTransactions` according to https://github.com/neo-project/neo-plugins/blob/d35297618006bc0dfdae3bacab22e98df682d84b/SimplePolicy/SimplePolicyPlugin.cs#L101
It would probably be best to add this to the config file so it can be easily updated in the future when NEO decides to add more types of TX's to it.
- make sure we updated the docstrings
- have a new exception type for more control. e.g. `TransactionFeeException`. Otherwise we can't provide proper feedback to the user. Without it something like the following will just print insufficient funds, instead of telling the user that they need to increase their TX fee.
https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Prompt/Commands/Invoke.py#L52-L56

neo/Prompt/Commands/SC.py
jseagrave21: I copied this verbage from the other locations we use a fee such as https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Prompt/Commands/Send.py#L44
Should I change it everywhere?

neo/Prompt/Commands/SC.py
jseagrave21: Answer 1) This logic follows the same logic as in `Send.py` and you can see here https://github.com/CityOfZion/neo-python/blob/6c1d68f45af2fc0909cb64cfd8c51dab2dcfd195/neo/Prompt/Commands/tests/test_sc_commands.py#L286-L302 that a bad fee will present a normal printout.

Answer 2) I can move the code up in the sequence 👍

neo/Prompt/Commands/SC.py
jseagrave21: Responses:
1) If you do a search for 'MakeTransaction' on GH, it looks like the only time `MakeTransaction` is used is associated with `Invoke` and `Contract` transactions. `Claim` transactions do not use `MakeTransaction` (see https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Prompt/Commands/Wallet.py#L278), so I think that issue is solved.
2) _blushing_ what are docstrings?
3) And I will try to see how create a new Exception type. I think I will model it after the special exceptions in JsonRpcApi.py

neo/Prompt/Commands/SC.py
ixje: Yes please :)

neo/Prompt/Commands/SC.py
ixje: 1. sounds good then!
2. https://www.pythonforbeginners.com/basics/python-docstrings
3. it's pretty straightforward e.g.
```
class UnsufficientTXFee(Exception):
   pass
```
and then you can do i.e. `raise UnsufficientTXFee("require x, got y")`

jseagrave21:
@ixje please see changes in https://github.com/CityOfZion/neo-python/pull/869/commits/641ec3a954f5694147061ea58e45d4756f134690
I am not sure what docstrings I should update but I think everything else is good to go.

neo/Prompt/Commands/SC.py
```diff
@@ -233,12 +233,11 @@ def execute(self, arguments):
             print("Please open a wallet")
             return False
 
-        if len(arguments) < 6:
+        if len(arguments) < 5:
```
ixje: Is there a specific reason why you completely removed the ability to add fees? I think that feature is still viable to prioritise deployment over other transactions in the mempool.

jseagrave21:
[ ] TODO: Update fee description everywhere

neo/Core/TX/Transaction.py
```diff
@@ -774,3 +774,19 @@ def __init__(self, *args, **kwargs):
...
+    """
+
+    message = None
+
+    def __init__(self, message):
+        super(TransactionError, self).__init__(message)
+        self.message = message
+
+    @staticmethod
+    def FeeError(message=None):
```
ixje: I understand that you copied the structure used for the JSON server. For this case though I don't think it's optimal. The JSON server is very specific in which it needs an error code and message. More important is that it is also converted to the JSON/text format. Meaning something like so:
```
"error": {"code": -32700, "message": "Parse error"}}
```
The only way the receiving party can differentiate is on `error.message` or `message` alone. They _have_ to do string comparisons to decide how to act. 

In our case the `wallet.MakeTransaction` function is always called from other python code and so can benefit from handling python specific exceptions without having to parse `exception.message`. For example, say we extend your `TransactionError` to also include a `InsufficientFunds` static method. We'd have to do something like:
```
try:
    tx = wallet.MakeTransaction
except TransactionError as te:
   if "insufficient funds" in te.message:
       #do something
   elif "fee" in te.message:
      #do something
   else:
      # default do something
```

if we would make exceptions specific like e.g.
```
class TXFundsError(Exception):
   pass

class TXFeeError(Exception):
   pass
``` 
We can then do
```
try:
   tx = wallet.Maketransaction
except TXFundsError as e:
   #do something if insufficient funds
except TXFeeError as e:
  # do something if insufficient fees
```
This is more pythonic, easier to control (little bit of background [info](https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions)) and increases re-usability of `wallet.MakeTransaction`. Note that big frameworks like Django do this [as well]( 
https://docs.djangoproject.com/en/2.1/_modules/django/core/exceptions/)

neo/Prompt/Commands/Invoke.py
```diff
@@ -309,7 +315,7 @@ def test_invoke(script, wallet, outputs, withdrawal_tx=None,
         wallet_tx = tx
     else:
         try:
-            wallet_tx = wallet.MakeTransaction(tx=tx, from_addr=from_addr)
+            wallet_tx = wallet.MakeTransaction(tx=tx, from_addr=from_addr, skip_fee_calc=True)
```
ixje: if we skip fee calculations, will we then not get wrong information in places that print information like this?
https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Prompt/Commands/SC.py#L188-L189

neo/Prompt/Commands/SC.py
jseagrave21: I will add it back in. I wasn't thinking about the case where a fee wasn't required but was desired for prioritization.

neo/Core/TX/Transaction.py
jseagrave21: @ixje Thank you for the in depth explanation!

neo/Prompt/Commands/Invoke.py
jseagrave21: In this case, no because I include the logic after the transaction is completely constructed here https://github.com/CityOfZion/neo-python/pull/869/files/641ec3a954f5694147061ea58e45d4756f134690#diff-5ee74e6bf4c83897694f97ae0557af8eR388
Also, if the exception is raised here then Python will fail here https://github.com/CityOfZion/neo-python/pull/869/files/641ec3a954f5694147061ea58e45d4756f134690#diff-5ee74e6bf4c83897694f97ae0557af8eR322
because `wallet_tx` will have been called before being defined.

jseagrave21:
@ixje while updating the fee descriptions, I noticed that all of the "token" transactions lack the ability to add a priority fee. Would you like me to add this ability now, in a separate PR, or not at all?

neo/Prompt/Commands/tests/test_send_commands.py
```diff
@@ -223,6 +223,28 @@ def test_insufficient_funds(self):
...
+    def test_transaction_size_1(self):
+        with patch('sys.stdout', new=StringIO()) as mock_print:
+            with patch('neo.Core.TX.Transaction.Transaction.Size', return_value=1026):  # returns a size of 1026
+                PromptData.Wallet = self.GetWallet1(recreate=True)
+                args = ['send', 'gas', self.watch_addr_str, '5']
+
+                res = Wallet.CommandWallet().execute(args)
+
+                self.assertFalse(res)
+                self.assertIn('Transaction cancelled. The tx size (1026) exceeds the max free tx size (1024).\nA network fee of 0.001 GAS is required.', mock_print.getvalue())
```
ixje: it would be nice if we could add the same comment as you did for deploy
```
# notice the required fee is equal to the low priority threshold
```
I found that those were really useful in quickly understanding the reasoning behind the required value.

neo/Prompt/Commands/tests/test_send_commands.py
```diff
@@ -223,6 +223,28 @@ def test_insufficient_funds(self):
...
+    def test_transaction_size_2(self):
+        with patch('sys.stdout', new=StringIO()) as mock_print:
+            with patch('neo.Core.TX.Transaction.Transaction.Size', return_value=1411):  # returns a size of 1411
+                PromptData.Wallet = self.GetWallet1(recreate=True)
+                args = ['send', 'gas', self.watch_addr_str, '5', '--fee=0.001']
+
+                res = Wallet.CommandWallet().execute(args)
+
+                self.assertFalse(res)
+                self.assertIn('Transaction cancelled. The tx size (1411) exceeds the max free tx size (1024).\nA network fee of 0.00387 GAS is required.', mock_print.getvalue())
```
ixje: same here as above. A quick comment indicating `(1411-1024) * 0.0001 (FEE_PER_EXTRA_BYTE)=0.00387` would be nice if we ever have to revisit this test

neo/Prompt/Commands/tests/test_sc_commands.py
```diff
@@ -283,15 +283,56 @@ def test_sc_deploy(self):
...
+                    res = CommandSC().execute(args)
+                    self.assertFalse(res)
+                    self.assertIn("Priority Fee (0.001) + Deploy Invoke TX Fee (0.0) = 0.001", mock_print.getvalue())
                     self.assertTrue(mock_print.getvalue().endswith('Insufficient funds\n'))
 
+        # test with ok contract parameter gathering, and tx is too large for low priority (e.g. >1024), just insufficient funds to deploy
+        with patch('sys.stdout', new=StringIO()) as mock_print:
+            with patch('neo.Prompt.Commands.LoadSmartContract.prompt', side_effect=prompt_entries):
+                with patch('neo.Prompt.Commands.SC.prompt', side_effect=[self.wallet_1_pass()]):
+                    with patch('neo.Core.TX.Transaction.Transaction.Size', return_value=615):  # returns a size of 1026
```
ixje: any idea why `return_value` needs to be `615` instead of `1026`? In your `test_send_commands` you do seem to just use `1026`, so this confuses me.

neo/Prompt/Commands/tests/test_send_commands.py
jseagrave21: Okay will do!

neo/Prompt/Commands/tests/test_sc_commands.py
jseagrave21: So I have tried to think through this and I can't figure it out. Could you help?

ixje: I just had a look and the reason is that we're patching in the wrong place. The test uses an `InvocationTransaction`, specifically it calls this 
https://github.com/CityOfZion/neo-python/blob/e561e5b57d4093a9ee0c87a8312d4a4d16f6fd5e/neo/Core/TX/InvocationTransaction.py#L32-L39

What your patch mocks is `super(InvocationTransaction, self).Size()`, which leaves `GetVarSize(self.Script)` still to be added to the total. 

Change your patch to the following and it will work as intended
```
with patch('neo.Core.TX.InvocationTransaction.InvocationTransaction.Size', return_value=1026):  # returns a size of 1026
```

jseagrave21: Thank you so much!

ixje:
💯

ixje:
300 points for jseagrave

lllwvlvwlll:
https://neoscan.io/transaction/19bfe57ab16dd88457db167b863e300dd525fb633f491795cbb20b98b76d6f18

PR #875, opened 2019-01-30
Add a "Safemode" to restrict peer connections

jseagrave21:
One issue I had during the most recent bad block hiccup was after I resynced my node using the most recent bootstrap files, it ended up getting stuck again because it was connected to a node which was also still stuck (the bootstrap files were behind the bad blockheight). My idea is to implement a `safemode` which restricts peer connections to addresses within the `SEED_LIST` which is populated from the config files.

Normally, whenever there is an issue with nodes, I check http://monitor.cityofzion.io/ to see which nodes are still good. In this case, there were still about 4 nodes which were still good. But updating my `SEED_LIST` with only these nodes did not restrict my node from connecting to nodes which were still up but also stuck.

I implemented a `safemode` which can be manually set in `Settings.py` or during startup of `np-api-server`, or `np-prompt`. Additionally, I added a `safemode` command to the `config` group so `safemode` could be disabled after headers have exceeded the bad block, or toggled any other time at the user's discretion.
**NOTE:** I also include a restart of `NodeLeader` anytime `safemode` is toggled on because the expectation is that the user is trying to protect his/her node from getting stuck and having to resync.

Personal Testing and unittest

No

- [x] Did you add any tests?


jseagrave21:
@ixje I understand you are reworking the servers from `Twisted` to `asyncio`; however, I think this PR or the idea behind it is still relevant and should be incorporated. For example, I am using it again after my testnet node got stuck at 2379318 to make sure that my node doesn't get stuck again.

ixje:
While at first sight the implementation looks sound, I currently don't have the bandwidth to run this code and observe it does what it's supposed to do without breaking something else. If we'd add IP filtering I'd rather see a blacklist/whitelist policy type implementation rather than being restricted to a "good seedlist" only option. 

I'll leave the PR open such that others can take notice of it or be referred to when asked about such an option.

jseagrave21:
> If we'd add IP filtering I'd rather see a blacklist/whitelist policy type implementation rather than being restricted to a "good seedlist" only option.

My idea was for a dynamic seedlist which constantly updated from a curated list based on blockheight. This should avoid connecting to "stuck" nodes.

Also, I have heard that neo-cli v2.10.2 should eliminate the "forking" issue entirely, so perhaps this idea is obsolete. What do you think @ixje? If you still think it is useful, I would like to try to port it into the asyncio branch.

ixje:
I'm not convinced the potential of forking is resolved until at least 6 months (ideally a year) of it not happening. Even then it's just a matter of "what new changes do they apply that might cause regression?". So that could justify it.

Note that the `asyncio` branch already has blacklist/whitelist [support](https://github.com/ixje/neo-python/blob/asyncio_alpha2/neo/Network/neonetwork/network/ipfilter.py), so it would need the service interface. 

Regarding that part; I'm concerned about who's going to maintain the service backend. The `IPFilter` updating part could be as simple as periodically requesting a JSON file with ip's from the server, parse and apply. I'd like to see it work and used before making it a permanent part of `neo-python`.

jseagrave21:
Okay, sounds good. I am gonna try to integrate my idea.  The whitelist I am using is essentially the json file which feeds monitor.cityofzion.io and then this list would be periodically updated to reflect height and the connected nodes would be edited to always connect to the best nodes.

jseagrave21:
@ixje closing this and tabling the idea until I can think of a better implementation.

jseagrave21:
Also reference https://github.com/jseagrave21/neo-python/tree/safemode-asyncio

PR #878, opened 2019-02-02
Fix config maxpeers

jseagrave21:
Currently, if a user uses `config maxpeers` to set `settings.CONNECTED_PEER_MAX` < `len(NodeLeader.Instance().Peers)` and then uses `show nodes` to display the connected nodes, there won't be any change. In fact, the number of connected peers will not change unless connection with one is lost.

I updated the logic in `config maxpeers` to reduce the number of connected peers according to the new `CONNECTED_PEER_MAX` by disconnecting the last added peers first, but only if the new `CONNECTED_PEER_MAX` < `len(NodeLeader.Instance().Peers)` (see tests for proof).

Personal testing and unittest

No

- [x] Did you add any tests?


neo/Prompt/Commands/Config.py
```diff
@@ -198,8 +198,19 @@ def execute(self, arguments):
...
+                current_max = settings.CONNECTED_PEER_MAX
                 settings.set_max_peers(c1)
-                print("Maxpeers set to ", c1)
+                c1 = int(c1)
+                p_len = len(NodeLeader.Instance().Peers)
+                if c1 < current_max and c1 < p_len:
+                    to_remove = p_len - c1
+                    peers = NodeLeader.Instance().Peers
+                    for i in range(to_remove):
+                        peer = peers[-1]  # disconnect last peer added first
```
ixje: 👍

CHANGELOG.rst
```diff
@@ -3,6 +3,10 @@ Changelog
 
 All notable changes to this project are documented in this file.
 
+[0.8.4] in progress
+----------------------
+- Fix ``config maxpeers``
```
ixje:
```suggestion
- Fix ``config maxpeers`` to disconnect peers if limited is lower than before
```
I'd like to suggest a more descriptive entry to people know straight away what was fixed without having to go through the actual PR

lllwvlvwlll:
https://neoscan.io/transaction/159616887a764517e1df4a728de01a69ef0ff9ab2c5891ed7b8065a0e38b9e4e

PR #905, opened 2019-02-28
Fix sys_fee calculation in block persist

### LINKED ISSUE #749: `getrawtransaction` returns the wrong values
### Current behavior
`getrawtransaction` returns the wrong information

RPC query:
```
"jsonrpc": "2.0",
        "id": 1,
        "method": "getrawtransaction", 
        "params": ["ea39a3bbd678714dc7e3342af8cb02444346a95b318b5d50f17fab1d408116d9", 1]
```
neo-cli v2.9.2 return:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xea39a3bbd678714dc7e3342af8cb02444346a95b318b5d50f17fab1d408116d9",
        "size": 1665,
        "type": "InvocationTransaction",
        "version": 1,
        "attributes": [],
        "vin": [
            {
                "txid": "0x1cc42c3ff05736b2f520c7c2beb95b92b77a8bd050b0161b16789fa84168efe2",
                "vout": 0
            },
            {
                "txid": "0x79bea5006d833f55b9bc4d142db45206e7baa4f95aa64fda764281581d52d5fa",
                "vout": 0
            },
            {
                "txid": "0x584558fbf9ba0a48ad07c1038b99023a52f685b7f83de33c18c72b8d552756ba",
                "vout": 0
            },
            {
                "txid": "0x1c0e4e45c2c57e6d0002ac84cc766d9b951d3caf1f369307201553c3540da7a1",
                "vout": 0
            },
            {
                "txid": "0x5f7f3747583ea08abc231ecac86982bd54c419a7e59a7379e3b020a6382fc878",
                "vout": 0
            }
        ],
        "vout": [],
        "sys_fee": "490",
        "net_fee": "0",
        "scripts": [
            {
                "invocation": "40e97a343af95bd84000931f4d5441fe78d7f2397779048b58958857ac7b235ce99668d343416363b19cf3e47cbe76cafa7cdf8fdf96b936d14f1b913a634501f4",
                "verification": "2102c0a8014e4dd0edf71087ff9f4dfffd31dfc26e4ca9e480966b112727df9147d4ac"
            }
        ],
        "script": "OMITTED",
        "gas": "490",
        "blockhash": "0x848054686192e4bad372ed0cdcab888a6f9b3839a8b436f50f776a956c85b6d4",
        "confirmations": 50595,
        "blocktime": 1543489647
    }
}
```
neo-python v0.8.2:
```
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "txid": "0xea39a3bbd678714dc7e3342af8cb02444346a95b318b5d50f17fab1d408116d9",
        "size": 1665,
        "type": "InvocationTransaction",
        "version": 1,
        "attributes": [],
        "vout": [],
        "vin": [
            {
                "txid": "0x1cc42c3ff05736b2f520c7c2beb95b92b77a8bd050b0161b16789fa84168efe2",
                "vout": 0
            },
            {
                "txid": "0x79bea5006d833f55b9bc4d142db45206e7baa4f95aa64fda764281581d52d5fa",
                "vout": 0
            },
            {
                "txid": "0x584558fbf9ba0a48ad07c1038b99023a52f685b7f83de33c18c72b8d552756ba",
                "vout": 0
            },
            {
                "txid": "0x1c0e4e45c2c57e6d0002ac84cc766d9b951d3caf1f369307201553c3540da7a1",
                "vout": 0
            },
            {
                "txid": "0x5f7f3747583ea08abc231ecac86982bd54c419a7e59a7379e3b020a6382fc878",
                "vout": 0
            }
        ],
        "sys_fee": "4.9e-06",
        "net_fee": "489.9999951",
        "scripts": [
            {
                "invocation": "40e97a343af95bd84000931f4d5441fe78d7f2397779048b58958857ac7b235ce99668d343416363b19cf3e47cbe76cafa7cdf8fdf96b936d14f1b913a634501f4",
                "verification": "2102c0a8014e4dd0edf71087ff9f4dfffd31dfc26e4ca9e480966b112727df9147d4ac"
            }
        ],
        "script": "OMITTED",
        "gas": 49000000000,
        "blockhash": "0x848054686192e4bad372ed0cdcab888a6f9b3839a8b436f50f776a956c85b6d4",
        "confirmations": 50615,
        "blocktime": 1543489647
    }
}
```
### Expected behavior
for neo-python:
- "sys_fee" should be `490`
- "net_fee" should be `0`
- "gas" should be `490`

### How to reproduce

Compare the query between a node running neo-cli v2.9.2 and a node running neo-python v0.8.2. (see https://happynodes.f27.ventures/table to find appropriate nodes).

### Your environment

N/A

### LINKED ISSUE #902: [BUG] Decimal issue with claimable GAS from sys_fee
### Current behavior
All systems fees in transactions should be distributed to the NEO holders. In neo-python this becomes visible as unavailable GAS and after spending your NEO as available GAS to claim. If you have a wallet with 100M NEO inside, you should be getting the GAS of all systems fees. For example, if someone deploys a contract and spends 490 GAS then this GAS should become claimable in your wallet.

There seems to be a bug (in `neo-python` `0.8.4`) where the calculation of system fee distribution is ignoring the decimal. So for example if you have 32920 GAS as unavailable for claiming, and a 490 GAS contract is being deployed, the counter increases to 49000032920 GAS (instead of the expected 33410). Any claiming of GAS will now also fail because the amount in the claim transaction is wrong.

### Expected behavior
Correct incrementing of unavailable GAS for claiming in transactions with a `sys_fee`.

### How to reproduce
- Set up a private net
- Make sure you have a wallet with 100M NEO
- Wait until you can claim enough GAS to deploy a contract (490 GAS)
- Claim GAS and deploy a contract (spending 490 GAS on sys_fee)
- Type `wallet` to check your unavailable GAS balance in claims

### Your environment
- OS: Linux in Docker
- neo-python version: `0.8.4`
- Python version: `3.7`

ixje:
Fix #902

#902 is the unforeseen result of fixing (among others) a wrong `net_fee` calculation reported in #749. 
It all started when I there changed the `SystemFee` implementation of the `InvocationTransaction`from
```python
   return self.Gas // Fixed8.FD()
```
to
```python
   return self.Gas
```
to be inline with neocore's  ([ref](https://github.com/neo-project/neo/blob/b5bd56d9a789419fbb4bf861e3f373a2d6685b3f/neo/Network/P2P/Payloads/InvocationTransaction.cs#L17))
```C#
public override Fixed8 SystemFee => Gas;
```

This together with another change fixed #749. What I unfortunately did not foresee is that it would cause the issue described in #902, which is caused by this line:
https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py#L666

According to the neocore implementation the equivalent of our `block.TotalFees().value` would be ([ref](https://github.com/neo-project/neo/blob/bbbc0cf5e401e425e7d749f26b23b824953da10a/neo/Ledger/Blockchain.cs#L445))
```C#
(long)block.Transactions.Sum(p => p.SystemFee)
```
The crux of the problem is the cast from `Fixed8` to `long`, which is implemented as ([ref](https://github.com/neo-project/neo/blob/bbbc0cf5e401e425e7d749f26b23b824953da10a/neo/Fixed8.cs#L171))
```C#
        public static explicit operator long(Fixed8 value)
        {
            return value.value / D;
        }
```
Our version did not apply this cast (essentially divide by `Fixed8.FD()`) on the `TotalFees`. I believe this is just an error that originated in the faulty `InvocationTransaction.SystemFee` implementation and that propagated throughout the code base.

First off, I applied the same division to the `TotalFees` end result. 

Second, since we now get a float instead of an int, I changed the logic for creating a 8 bytes bytearray from an int, to use `struct.unpack("<d", float)` (which creates an 8 bytes double in little endian format). This appears to be inline with storing a float (8 bytes in C#).

Third, I took the time to cross referenced the implementation of all other transactions on `system fee` and `network fee` with the C# code and made 2 changes there.

Finally, I updated all tests + fixtures (because the db storage is different)

`make test` + compare with neo-cli

yes, we change the storage format of the system fee in the db. Previously
```
DBPrefix.DATA_Block + block hash + INT systemfee bytes (8) + block
```
now
```
DBPrefix.DATA_Block + block hash + FLOAT systemfee bytes (8) + block
```

This will require a resync of the chain 😞 , but should at least fix all known `net_fee` and `sys_fee` discrepancies. 

- [ ] Did you add any tests?


jseagrave21:
Hey there, I noticed this also requires an update of the fixtures. Any chance you were able to implement `v9` from #716 and then update to `v10`? Maybe merging this fix will also allow #716 to be merged?

ixje:
going to merge even though mainnet/testnet fixtures still can't be made (VM just got updated again)

PR #906, opened 2019-03-01
Prove "getbalance" RPC method

jseagrave21:
This PR is a result of a discussion with @skinderis on the #python channel in Discord where the functionality of the "getbalance" RPC method was questioned.

The original tests did not verify an expected balance was returned

Open wallets from `WalletTestCase` and verified that the RPC method returned the expected balances.

`make test`

No

- [x] Did you add any tests?


skinderis:
@jseagrave21 Good job 👍 I am Igoris by the way ;)

ixje:
thanks!

Issue #891, opened 2019-02-13
Update to throw exceptions instead of logging errors [np-main]


ixje:
This is a TODO for a project (see side bar)

Try to remove as much logging errors inside the business logic and raise Exceptions instead. Try to bubble it up to the point the that presentation layer (at this point only a CLI) can present useful and detailed information (if necessary).
- Do not use the generic Exception class but use the more specific exception classes like `ValueError`, `TypeError` etc ([standard lib exceptions](https://docs.python.org/3/library/exceptions.html)).
- Create custom exceptions if needed to give sufficient control to the end-user.

jseagrave21:
@ixje are you sure this one is done and not #888?

Issue #890, opened 2019-02-13
Improve neo-python reusability


ixje:
This is a roadmap milestone item.
See the `Roadmap` project on the side bar for more information where it fits in the bigger picture.
See the `neo-python reusability` project on the side bar for a breakdown of items and progress.

**Rationale**
Like with neo-python-core (issue [ref](https://github.com/CityOfZion/neo-python/issues/887)), the main project has a good amount of code that immediately reports errors to the CLI by logging instead of throwing exceptions. To make these code parts easier reusable they should be converted to throw exceptions. Other areas of improvement include docstrings, “smart code” and variable argument lists.

ixje:
current state is sufficient for LTS. NEO3.0 we'll focus on the larger improvements in the NEO3.0 build

PR #907, opened 2019-03-03
Improve usage of neo-python CLI

jseagrave21:
- Many actions receive user confirmation via prompt input (e.g. <password>, "confirm", etc.) but most do not handle a `KeyboardInterrupt` without crashing due to the unhandled error.
- Also, an option for password prompting for `ClaimGas` and `SplitUnspentCoin` still exists which is not consistent with the rest of neo-python
- Additionally, the description for "path" is wrong for `sc deploy`.
- Finally, a "fakepath" file remains whenever `make test` is run.

Supports https://github.com/CityOfZion/neo-python/issues/890 and I do not think it conflicts with https://github.com/CityOfZion/neo-python/issues/891 because the `KeyboardInterrupt` handling is integral to the user experience.

- add support for cancelling actions with `KeyboardInterrupt`
- remove the ability to circumvent a password prompt for `ClaimGas` and `SplitUnspentCoin`
- correct help documentation for `sc deploy` "path"
- cleanup "fakepath" file created in test_wallet.py

added tests and ran `make test`

No

- [x] Did you add any tests?


neo/Prompt/Commands/SC.py
```diff
@@ -313,7 +325,7 @@ def execute(self, arguments):
             return False
 
     def command_desc(self):
-        p1 = ParameterDesc('path', 'path to the desired Python (.py) file')
+        p1 = ParameterDesc('path', 'path to the desired smart contract (.avm) file')
```
ixje: 👍

ixje:
Nice! 💪  
It would not be the first time that I accidentally cause some errors because I try to abort via a  keyboard by instinct.

ixje:
200 points jseagrave

lllwvlvwlll:
https://neoscan.io/transaction/d5ae4a22ffe49912a2a3dc698d39ac63a8fcf2cef63ae136850d039c312591d8

PR #912, opened 2019-03-14
Add wallet height to bottom toolbar

hal0x2328:
Monitoring wallet height requires getting all the wallet info which can be slow depending on how many address/tokens/contracts exist in the wallet

Added the current wallet height to the bottom toolbar in the same format neo-cli uses (wallet height/db height/blockchain height)

Manual testing

No

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.5-dev] in progress
 -----------------------
+- Add current wallet height to bottom toolbar
 - Change class attributes of ``Transaction`` into instance attributes and fix constructor
 - Improve testing of "getbalance" RPC method
 - Support cancelling actions with ``KeyboardInterrupt``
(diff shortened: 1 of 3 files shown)


hal0x2328:
Wallet._current_height seems to always be one higher than the db/blockchain heights. Does this track? Should I modify the height to subtract 1 for display purposes to avoid confusion?

ixje:
I actually encountered this issue yesterday as well in my `asyncio` branch.
If you could change this https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/Wallets/Wallet.py#L653-L662
to this
```python
            while self._current_height < Blockchain.Default().Height and (block_limit == 0 or blockcount < block_limit):

                block = Blockchain.Default().GetBlockByHeight(self._current_height)

                if block is not None:
                    self.ProcessNewBlock(block)
                else:
                    break

                blockcount += 1
```
and a single wallet rebuild will be enough

at least that way we avoid confusion when looking at the toolbar code

hal0x2328:
Ok, made the requested change

ixje:
👍

Issue #781, opened 2018-12-31
Support named parameter type entry for sc build commands


ixje:
### Current behavior
Do after #623 is merged.

Currently the `sc build_run` and `sc load_run` commands require numeric entry of the contract parameter types defined in https://github.com/CityOfZion/neo-python/blob/e3780126f9cf8175f9f3c7ad27f36f2baae20508/neo/SmartContract/ContractParameterType.py#L29-L39

### Expected behavior
We can improve the user experience by allowing for named inputs, such that these numbers don't have to be remembered.

LysanderGG:
I'd like to implement this.

I would suggest both complete names and some abbreviations, for instance "sig", "bool", "int", "str", "arr"

ixje:
👌 
ping @jseagrave21 to avoid double work; as you also indicated you might want to pick this up.

jseagrave21:
@LysanderGG go for it! I started investigating yesterday but don't have anything written yet. cc @ixje 

> I would suggest both complete names and some abbreviations, for instance "sig", "bool", "int", "str", "arr"

I like that it would support both. The original issue was that we didn't want the user to have to remember anything proprietary. 

My original thought was requiring a list entry (like params for RPC)
So, replace
```
070202
```
With
```
[string, integer, integer]
```

Also, I was toying with the idea of still supporting the original format. But that is definitely not a must imo.

PR #913, opened 2019-03-15
add --return-type to override test invoke return parameter type

hal0x2328:
Contract invocation results are displayed using the default return type of the contract, even though some operations will return different types than the default, requiring the user to manually convert results, e.g. from bytearray to integer

Added a --return-type argument to the `sc invoke` command to allow the user to override the contract's default return type during a test invoke

Manual testing

No

- [ ] Did you add any tests?

### DIFF
--- a/neo/Prompt/Utils.py
+++ b/neo/Prompt/Utils.py
@@ -148,6 +148,15 @@ def get_to_addr(params):
     return params, to_addr
 
 
+def get_return_type_from_args(params):
+    return_type = None
+    for item in params:
+        if '--return-type' in item:
+            params.remove(item)
+            return_type = item.replace('--return-type=', '')
+    return params, return_type
+
+
 def get_fee(params):
     fee = None
     for item in params:
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,7 @@ All notable changes to this project are documented in this file.
 
 [0.8.5-dev] in progress
 -----------------------
+- Allow override of test invocation return type
 - Add current wallet height to bottom toolbar
 - Change class attributes of ``Transaction`` into instance attributes and fix constructor
 - Improve testing of "getbalance" RPC method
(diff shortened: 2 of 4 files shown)


neo/Prompt/Utils.py
```diff
@@ -148,6 +148,15 @@ def get_to_addr(params):
     return params, to_addr
 
 
+def get_return_type_from_args(params):
+    return_type = None
+    for item in params:
+        if '--return-type' in item:
+            params.remove(item)
+            return_type = item.replace('--return-type=', '')
```
jseagrave21: I recommend adding some logic here to ensure only acceptable ParameterTypes are allowed.

hal0x2328:
> This is a great addition. Will the code support the string equivalents of the ParameterTypes? I ask because I know it is a goal for users to be able to specify the input and output types with their string equivalents (i.e. "Integer" is acceptable, not just "02")
> See #781

It does support using the string equivalents but I don't use those in the example because the convention has always been hex notation, both in neo-cli and neo-python, and all the documented examples use the hex notation. Unless it changes everywhere users are still going to need to know the mapping between hex and string parameter notation.

neo/Prompt/Utils.py
hal0x2328: I asked myself when adding this option, does it add more value to add some sort of input sanity check rather than just educating the user through hitting an exception when they enter a bad parameter type? This is for advanced users after all. Personally I like the feedback loop approach, it doesn't really have any negative impact and it cuts down on the amount of code that needs to be maintained.

ixje: While true that it reduces code maintenance, we put in a lot of effort to get rid of exception throwing due to user input errors when we refactored the cli. I personally don't think that an advanced option or tool justifies not handling exceptions. In this case it's also as simple as doing

```python
try:
   [ContractParameter.AsParameterType(ContractParameterType.FromString(return_type), item).ToJson() for item in results]
except ValueError:
    # inform user
```

hal0x2328:
Added exception handler and unit tests.

ixje:
Thanks!

PR #923, opened 2019-03-28
Add PICKITEM for ByteArray into VM

ixje:
https://github.com/neo-project/neo-vm/pull/98

write code

run VM test vectors

no

- [X] Did you add any tests?


ixje:
`StackItem.New` did not create a `ByteArray` when receiving a `str` input value therefore failing some local tests. The VM test-vectors did not cover this.
C# ref: https://github.com/neo-project/neo-vm/blob/bfb68c806f52ed773ae7badff3b652e96e257c61/src/neo-vm/StackItem.cs#L95-L98

PR #924, opened 2019-03-29
Catch Ctrl+C on SC invoke password entry

metachris:

Ctrl+C on password entry when invoking a SC breaks neo-python

try-catch

manual test

no

- [ ] Did you add any tests?


ixje:
It was setup for master branch instead of dev. If you can resolve the conflicts we should be good.

metachris:
didn't check dev :P

https://github.com/CityOfZion/neo-python/blob/development/neo/Prompt/Commands/SC.py#L216

ixje:
ah I see. I thought this could have been a missed try/catch

PR #925, opened 2019-03-30
integrate neocore back into neo-python

ixje:
About a year ago we split some stuff into a separate repository thinking it would be useful, we’d expand on that and have easier maintenance due to separation. I believe it has not met its objective and the only things used without `neo-python` are perhaps the `Crypto` module and some CLI helper scripts. I don’t believe anyone is using `UInt160/256`, `Fixed8` or `BinaryReader/Writer` classes without making use of the other building blocks that are all still in `neo-python`.

Whenever there are changes that need to happen on both sides it becomes a hairy thing to manage with build systems failing left and right, requiring forced releases of the core just to get them working again etc. History has proven this enough times now and not just while under my maintenance. I thought `git submodules` would be a solution, but that seems to be too easy to do it wrong from the many reviews I’ve read. Which in turn will make contributions from outsiders harder.

The `neo-python` project is < 2MB so I don’t think there’s any valid argument against a bigger project/import size. This PR moves the few items we've put into `neo-python-core` back into the main project. This will also ease the bigger refactoring we want to do.

make test

yes, temporarily pointing travis to my local neo-boa repo to get the build to pass as this is a chicken/egg problem. After it's merged in `development` I can PR for `neo-boa` (which looks at `neo-python/development`). Once that's in I can change travis back to `neo-boa`. 

- [ ] Did you add any tests?


jseagrave21:
💪

PR #928, opened 2019-04-03
Fix 'gas' output type for InvocationTransaction.ToJson()

jseagrave21:
While trying to pinpoint the discrepancy in Size calculations for transactions as reported by @hal0x2328, I discovered that the 'gas' output when calling `InvocationTransaction.ToJson()` is a `str` and should be an `int`. See tests.

Translated existing neo-project unittests to neo-python.

`make test`

No

- [x] Did you add any tests?


ixje:
This is not correct. Gas is still a string
https://github.com/neo-project/neo/blob/5ce0e4c3192cdcce700105030ed03197961e0466/neo/Network/P2P/Payloads/InvocationTransaction.cs#L60

You can also validate this against the RPC output of a `2.10.0.0` node e.g. for tx 0x1bbda585edf52c28ef209602c185d21b12a0272b4bd0e79c7231e7a23393cd5b

jseagrave21:
Oops. Okay then their test is wrong. I should have checked in more places. Thank you, @ixje!

ixje:
Their test is fine. The test translates the value to a number for testing purposes
```C#
jObj["gas"].AsNumber().Should().Be(42);
```
https://github.com/neo-project/neo/blob/5ce0e4c3192cdcce700105030ed03197961e0466/neo/IO/Json/JString.cs#L23

For more clarity they could have probably used
```C#
jObj["gas"].AsString().Should().Be("42");
```
I don't know why their tests work like this, but it's testing the right thing.

PR #929, opened 2019-04-05
Improve transaction test coverage and fix PublishTransaction.Size()

jseagrave21:
This PR is a result of the ongoing size discussion on Slack.

I compared all the transaction types to the code found in https://github.com/neo-project/neo/tree/master/neo/Network/P2P/Payloads after @dauTT indicated that `PublishTransaction.Size()` was wrong. I assess this is not a big deal since `PublishTransaction` has been deprecated (see https://docs.neo.org/en-us/sc/quickstart/overview.html), but I fixed it. I did not find any other discrepancies. I updated the tests to be more descriptive and added tests to ensure every transaction type has basic coverage.

`make test`

No

- [x] Did you add any tests?


ixje:
thanks! looks good 👍

ixje:
200 points seagrave

lllwvlvwlll:
https://neoscan.io/transaction/c382c87ce7282664dc849fef5cb8c8de22cf3b596dd1b4e85c87efb194bb2003

Issue #896, opened 2019-02-13
Add alternative DB support (e.g. Rocksdb/reddis)


ixje:
Add support for Rocksdb or another database like Redis. This issue depends on https://github.com/CityOfZion/neo-python/issues/895

Support for Rocksdb was once started by a user and the code is still available [here](https://github.com/CityOfZion/neo-python/issues/157)

PR #939, opened 2019-05-10
Preperation to support different database backends

### LINKED ISSUE #894: Improve data(base) access
This is a roadmap milestone item.
See the `Roadmap` project on the side bar for more information where it fits in the bigger picture.
See the `improve data(base) access` project on the side bar for a breakdown of items and progress.

**Rationale**
The current used database (`leveldb`) only allows single process access. The idea behind this objective is to improve accessibility to processed data from other system/processes. This requires supporting a backend which at least allows reads from multiple processes simultaneously.

### LINKED ISSUE #895: Allow alternative backends
This is a TODO for a project (see side bar)

- Explore how to better decouple the current database layer. 
- Create a database interface to allow swapping back-ends.
   - Clearly document it for others
    - Allow for optional rollback support. This means we’d like to be able to rollback the state to a certain block height.

merl111:

preperation for #896 
closes #895 
closes #894 

The whole database access has been reworked and decoupled. Business logic which has been part of `LevelDBBlockchain` has been moved to the `Blockchain` class. A `factory` to dynamically generate the database class has been introduced. The class returned by the `factory` handles `write` `read` `delete` access to the database, therefore it's completely decoupled from the blockchain business logic. Same thing was done for the `NotificationDB`.

Synced full node.
Bootstrapped node.
All unit tests pass.

The whole business logic from `LevelDBBlockchain` has been moved to the `Blockchain` class.

- [x] Did you add any tests?


merl111:
I would also like to reference the [np-rocksdb](https://github.com/merl111/neo-python/tree/np-rocksdb) which is based on [alt-db-support](https://github.com/merl111/neo-python/tree/alt-db-support) and implements RocksDB as an alternative database backend. It's not fully tested yet though, all the unit tests pass ( fixtures are migrated to RocksDB ) but I have not yet synced a full node.

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
```diff
@@ -0,0 +1,235 @@
...
+
+logger = log_manager.getLogger()
+
+"""
+Description:
+    All the methods are used within the dynamically generated class from the
+    database factory.
+Usage:
+    For a new database implementation all methods defined in
+    AbstractDBImplementation have to be implemented.
```
ixje: I think the same comment applies here as I did on the `AbstractDBImplementation`. I'm aware that we refer to this module as an example how to add a new backend. Therefore I'm fine with adding some more words that are backend generic/howto. E.g.
```
Description:
    A backend implementation for the LevelDB database.

   It overrides all methods from the `AbstractDBImplementation` class. The database factory (`DBFactory`) uses these methods to dynamically generate a conforming database instance for internal usage.
```

neo/Storage/Implementation/LevelDB/PrefixedDBFactory.py
```diff
@@ -0,0 +1,35 @@
...
+
+def internalDBFactory(classPrefix):
+    """ Internal database factory method used for prefixed dbs and snapshots.
+
+    The returned class is very similar the the class returned in
+    neo.Storage.Implementation.DBFactory._dbFactory but has a different
+    __init__ method.
+
+    Args:
+        classPrefix (str): Prefix to name the class appropiately.
```
ixje:
```suggestion
        classPrefix (str): Prefix to name the class appropriately.
```

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
```diff
@@ -0,0 +1,235 @@
...
+"""
+
+# name of the init method for a generic database class
+_init_method = '_db_init'
+
```
ixje: I had to look in the code to understand what the follow section does, which signals that we can improve the documentation to make it more self-explanatory. While saying that I now realise that `AbstractDBImplementation` doesn't cover all the necessary requirements to support an alternative backend. We actually also need to specify a specific set of variables like `_init_method` such that it can be picked up by 
```
        # add __init__ method
        attributes['__init__'] = attributes.pop(functions._init_method)
```
from the factory. 

I should have noticed this sooner but I think the current factory is limited in the way that it needs modification to support new backends. This can be seen by looking at your `Rocksdb` update
https://github.com/merl111/neo-python/blob/54970a06316f54ebbca86d4bb573404ae197ddbe/neo/Storage/Implementation/DBFactory.py#L115-L116

So why does this matter? It means that if somebody creates an alternative backend then there are 2 options for using it
1) it remains user maintained and for every `neo-python` update they need to patch it back in for usage
2) it becomes part of `neo-python` and maintenance (partially) becomes our responsibility. 

I'm thinking 2 things:
1) I'd like maintenance of alternative DBs to not be `neo-python`'s responsibility. 
2) I'd like to see if we can actually inherit from the `AbstractDBImplementation` and add _all_ required items/methods that need to be overridden in one place.

The maintenance thing can be resolved by making the DB class to be used to be loaded from the settings file. Similar to Django's settings/config. We actually do this for the RPC and REST server. 

https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/data/protocol.privnet.json#L44

https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/bin/api_server.py#L282

If you have any questions regarding anything above, hit me up on Discord.

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
```diff
@@ -0,0 +1,235 @@
...
+_iter = None
+
+# batch instance to perform batch operations on the database
+_batch = None
+
+_lock = threading.RLock()
+
+
+@property
+def Path(self):
```
ixje: is this a left over? Couldn't find any references to it

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
```diff
@@ -0,0 +1,235 @@
...
+    As it's used within a context, getBatch has to called within a with block.
+
+    Example:
+    with self._db.getBatch() as batch:
+        batch.put(b'key1', b'value')
+        batch.put(b'key2', b'value')
+        batch.delete(b'key2')
+
+    If a database backend is implemented that does not support batches you have
+    to implement an object that mimics a batches the behaviour.
```
ixje:
```suggestion
    to implement an object that mimics batch behaviour.
```

neo/Storage/Interface/DBInterface.py
```diff
@@ -2,11 +2,37 @@
...
+logger = log_manager.getLogger()
 
 
-class DBCollection:
-    DB = None
+class DBProperties:
+
+    """ Store the different DB properties
+    Used to pass the relevant information to
+    no.Storage.Implementation.[BACKEND].[BACKEND]DBClassMethods.openIter
```
ixje:
```suggestion
    neo.Storage.Implementation.[BACKEND].[BACKEND]DBClassMethods.openIter
```

neo/Storage/Interface/DBInterface.py
```diff
@@ -2,11 +2,37 @@
...
+
+    Args:
+        prefix (str, optional): Prefix to search for.
+        include_value (bool, optional): include vale used for creating an
```
ixje:
```suggestion
        include_value (bool, optional): include value used for creating an
```

ixje: I don't understand the description of the argument. Unless my IDE is failing me I only see it used in creation of `DBProperties` but I don't see any logic that actually does anything with it later on?

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,56 @@
+from abc import ABC, abstractmethod
+"""
+Description:
+    Abstract class used to ensure the mandatory methods are overwritten
+    in everxy new database implementation.
```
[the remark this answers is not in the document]
merl111: Applied your description, I like it better too.

neo/Storage/Common/DebugStorage.py
```diff
@@ -15,18 +15,17 @@ def db(self):
...
+                DBProperties(prefix=DBPrefix.ST_Storage,
+                             include_value=False)) as iterator:
+            for key in iterator:
+                self._db.delete(key)
 
     def __init__(self):
 
         try:
-            self._db = plyvel.DB(settings.debug_storage_leveldb_path, create_if_missing=True)
+            self._db = GetBlockchain().Default().GetDB().cloneDatabase(
```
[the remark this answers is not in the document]
merl111: The old implementation was basically doing the same, but it was done in the static `instance` method and it calls `clone_from_live`, so I just removed the `clone_from_live`. If it's not cloned no data prefixed `ST_Storage` would be available, so I am not sure if that matters, depends on what one uses debug storage for I guess.

neo/Storage/Implementation/DBFactory.py
```diff
@@ -0,0 +1,122 @@
...
+
+DATABASE_PROPS = settings.database_properties()
+
+
+def getBlockchainDB(path=None):
+    """
+    Returns a database instance used with the blockchain class.
+
+    Args:
+        path (str, optional): the full path to the blockchain database directory.
```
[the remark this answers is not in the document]
merl111: Yes, I have received an email with you comment. :smiley:

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
merl111: Added your description :+1:

neo/Storage/Implementation/LevelDB/PrefixedDBFactory.py
merl111: Typo fixed.

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
merl111: Yes I thought it's needed at first, removed now.

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
merl111: Fixed.

neo/Storage/Interface/DBInterface.py
merl111: Fixed typo.

neo/Storage/Interface/DBInterface.py
merl111: I added an improved description, the old one was not really clear. Actually, I added the `DBProperties` class for convenience and to have a single object you can pass to multiple database methods that might need the same or similar information, for now it's just used for `openIter`. `include_value` is currently only set to `False` in `_BuildCollectionKeys` in the `DBInterface`, other than that it's always `True`.

neo/Storage/Implementation/LevelDB/LevelDBClassMethods.py
merl111: > I'd like maintenance of alternative DBs to not be neo-python's responsibility.

Ok, my first impression was that those different backend implementation should be part of neo-python, but if that's not the case my approach might not be perfect. I have changed a few things, `DBFactory` does not need any changes anymore if one implements a different backend and I implemented your suggestion regarding to load the class from the settings file. Please check last commit.

> I'd like to see if we can actually inherit from the AbstractDBImplementation and add all required items/methods that need to be overridden in one place.

I also changed the LevelDB implementation a bit, so, if I am not missing anything, all required methods that need to be overridden should be in the `AbstractDBImplementation`.

neo/Storage/Common/DebugStorage.py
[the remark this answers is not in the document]
ixje: Makes sense. Thanks

neo/Core/Blockchain.py
```diff
@@ -55,6 +97,108 @@ class Blockchain:
...
+        else:
+            logger.error("\n\n")
+            logger.warning("Database schema has changed from %s to %s.\n" % (version, self._sysversion))
+            logger.warning("You must either resync from scratch, or use the np-bootstrap command to bootstrap the chain.")
+
+            res = prompt("Type 'continue' to erase your current database and sync from new. Otherwise this program will exit:\n> ")
+            if res == 'continue':
+
+                with self._db.getBatch() as wb:
+                    with self._db.openIter() as iterator:
```
ixje: will throw an exception
```
 File "/Users/erik/Documents/code/test-issues/merl/neo/Core/Blockchain.py", line 192, in __init__
    with self._db.openIter() as iterator:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 239, in helper
    return _GeneratorContextManager(func, args, kwds)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 82, in __init__
    self.gen = func(*args, **kwds)
TypeError: openIter() missing 1 required positional argument: 'properties'
```

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
```diff
@@ -0,0 +1,190 @@
...
+            include_value=properties.include_value,
+            include_key=properties.include_key)
+
+        yield _iter
+        _iter.close()
+
+    @contextmanager
+    def getBatch(self):
+        """
+        Yields a batch instance which can be used to perform atomic updates on the
```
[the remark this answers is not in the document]
ixje: To add to this. With us now inheriting and implementing `AbstractDBImplementation` we should move all docstrings to `AbstractDBImplementation` and expand the descriptions. I found it's necessary to know that e.g. `def write(self, key,value)` gets a `key` argument that's a `bytearray`. Specifically an escaped bytearray e.g. `b'\xf0'`. The same for its value. And we need to know in what format to return data (e.g. for `def get`).

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,56 @@
...
+    implementations.
+
+    For a more detailed information on the methods and how to implement a new
+    database backend check out:
+    neo.Storage.Implementation.LevelDB.LevelDBClassMethods
+"""
+
+
+class AbstractDBImplementation(ABC):
+
```
ixje: we need to add a mandatory `def __init__(self, path):` that needs to be overridden.

ixje:
One more things I encountered when running your LevelDB implementation
```
File "./neo/bin/prompt.py", line 304, in main
    blockchain = Blockchain(DBFactory.getBlockchainDB())
  File "/Users/erik/Documents/code/test-issues/merl/neo/Core/Blockchain.py", line 192, in __init__
    with self._db.openIter(DBProperties(prefix='')) as iterator:
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/erik/Documents/code/test-issues/merl/neo/Storage/Implementation/LevelDB/LevelDBImpl.py", line 132, in openIter
    include_key=properties.include_key)
  File "plyvel/_plyvel.pyx", line 362, in plyvel._plyvel.DB.iterator
TypeError: Argument 'prefix' has incorrect type (expected bytes, got str)
```

ixje:
Another thing that came to mind; we should provide a set of test vectors that new implementations can run against. 

For example my `SQLite` test implementation doesn't fail until it hits some block in which it requests a TX. It can actually get data from the DB, but the format is wrong. This particular code (of `GetTransaction`) was once `LevelDB` specific and is now generic in `Blockchain.py`. That is fine but it expects a certain format

https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/Implementations/Blockchains/LevelDB/LevelDBBlockchain.py#L387-L392

Meaning it expects a `bytearray` where the first 4 bytes are escaped, and the remainder are not 🙈 
e.g.
```
b'\x01\x02\x03\x04aabbccdd`
```

neo/Storage/Interface/DBInterface.py
```diff
@@ -2,11 +2,37 @@
...
+
+    """ Store the different DB properties
+    Used to pass the relevant information to
+    neo.Storage.Implementation.[BACKEND].[BACKEND]DBClassMethods.openIter
+
+    Args:
+        prefix (str, optional): Prefix to search for.
+        include_value (bool, optional): used to define if the value should be
+                                        included when opening an iterator.
+        include_key (bool, optional): used to define if the key value shoud be
```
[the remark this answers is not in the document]
merl111: fixed.

neo/Storage/Interface/DBInterface.py
```diff
@@ -2,11 +2,37 @@
...
+logger = log_manager.getLogger()
 
 
-class DBCollection:
-    DB = None
+class DBProperties:
+
+    """ Store the different DB properties
+    Used to pass the relevant information to
+    neo.Storage.Implementation.[BACKEND].[BACKEND]DBClassMethods.openIter
```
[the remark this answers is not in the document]
merl111: > Note that I used the include_value/key description directly from the Plyvel documentation for Iterators because I think that's the generic idea.

Good idea, that's exactly what it's used for. 

> The suggested descriptions describes the effect it has and requires no reasoning of the reader.

Yes, for me it was clear what it does, but thinking about it as a person from the other side of the fence, it makes totally sense.

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,56 @@
...
+    of this abstract class.
+Usage:
+    The dynamically generated class coming from the database factory inherits
+    from the abstract class, means the generated class cannot be used if not
+    all methods defined in this class are overwritten.
+
+    If this class is extended, make sure you extend also all other database
+    implementations.
+
+    For a more detailed information on the methods and how to implement a new
```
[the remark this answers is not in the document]
merl111: Fixed, took the second suggestion, as description fits a bit better in my opinion.

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,56 @@
...
+    database backend check out:
+    neo.Storage.Implementation.LevelDB.LevelDBClassMethods
```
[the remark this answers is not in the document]
merl111: Fixed.

neo/Settings.py
```diff
@@ -143,6 +143,18 @@ def debug_storage_leveldb_path(self):
         self.check_chain_dir_exists()
         return os.path.abspath(os.path.join(self.DATA_DIR_PATH, self.DEBUG_STORAGE_PATH))
 
+    def database_properties(self):
+        return {'blockchain': {'path': self.chain_leveldb_path,
+                               'skip_version_check': False,
+                               'backend': 'neo.Storage.Implementation.LevelDB.LevelDBImpl.LevelDBImpl'},
```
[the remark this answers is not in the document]
merl111: Yep, it's clearly configuration. moved it to the `protocol.xxx.json` files.

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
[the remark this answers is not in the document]
merl111: > We need to describe the interface of batch

I have added description for all available methods for the interface of `batch` 

> To add to this. With us now inheriting and implementing AbstractDBImplementation we should move all docstrings to AbstractDBImplementation and expand the descriptions.

Moved all docstrings to `AbstractDBImplementation` and added docstrings to `write`, `get`, delete`.

neo/Core/Blockchain.py
merl111: fixed, missed that one.

neo/Storage/Implementation/AbstractDBImplementation.py
merl111: Done.

merl111:
> One more things I encountered when running your LevelDB implementation

Fixed.

merl111:
> Another thing that came to mind; we should provide a set of test vectors that new implementations can run against.

I have written a few generic tests during the implementation of `RocksDB`, I added those to the `alt-db-support` branch, but I am not sure if that's exactly what you are asking for.

>  For example my SQLite test implementation doesn't fail until it hits some block in which it requests a TX. It can actually get data from the DB, but the format is wrong. This particular code (of GetTransaction) was once LevelDB specific and is now generic in Blockchain.py. That is fine but it expects a certain format.

I think I am missing something here. 

A transaction is stored in `Persist`:

https://github.com/merl111/neo-python/blob/c912b2be80fc80887e73f403cb9584aa090e993e/neo/Core/Blockchain.py#L974

the value is created from `block.IndexBytes() + tx.ToArray()` means the first 4 bytes are the `IndexBytes`, so in my opinion the code in `GetTransaction` is not LevelDB specific, but specific to what is inserted when the tx is written to the database. 

I'll hit you up on discord, to discuss this.

ixje:
> the value is created from `block.IndexBytes() + tx.ToArray()` means the first 4 bytes are the `IndexBytes`, so in my opinion the code in `GetTransaction` is not LevelDB specific, but specific to what is inserted when the tx is written to the database.
> 
> I'll hit you up on discord, to discuss this.

True, it's not specific anymore. It's just that argument type and format is derived from what was once necessary for leveldb. For example for `sqlite3` the values need to be strings and queries return strings. However all methods in `neo-python` currently operate on `b''` bytearrays. Because that's what was once returned by LevelDB/Plyvel.

merl111:
Ok, I understand, that should be covered in the docstrings for `get` now. Bytestrings mit have some advantages over strings in some cases, so for now I would just check if get really returns a `bytearray`, would that sufficient?

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,204 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def createSnapshot(self):
+        """
+        Creates a snapshot of the current database, used for DebugStorage and
+        NotificationDB. To keep the snapshot compatible to the current design it's
+        created through a factory which returns basically the same class we use
+        for the real database and all the methods that can be used on the real db
+        can also be used on the snapshot.
```
ixje: I think we can remove this part
```
       To keep the snapshot compatible to the current design it's
        created through a factory which returns basically the same class we use
        for the real database and all the methods that can be used on the real db
        can also be used on the snapshot.
```
as this looks like implementation details and an explanation towards me why we 'used' a factory (which no longer applies).

Also **snapshot** to me means a copy of state at a particular point in time. Now the discussion point would be; should it be read-only or not? 
[Leveldb's](https://github.com/google/leveldb/blob/master/doc/index.md#snapshots) snapshot feature is readonly,  SQLite also seems to be [readonly](https://www.sqlite.org/c3ref/snapshot.html) although they don't describe it as implicit as leveldb. I'm currently leaning towards having read-only as well, which currently would only mean updating the documentation to ensure that it reads 
```
Creates a read-only snapshot of the current database, used for DebugStorage and NotificationDB
```

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
```diff
@@ -0,0 +1,90 @@
...
+    def cloneDatabase(self, clone_db):
+        db_snapshot = self.createSnapshot()
+        with db_snapshot.openIter(DBProperties(prefix=DBPrefix.ST_Storage,
+                                               include_value=True)) as iterator:
+            for key, value in iterator:
+                clone_db.write(key, value)
+        return clone_db
+
+    def createSnapshot(self):
+        SnapshotDB = load_class_from_path('neo.Storage.Implementation.LevelDB.LevelDBSnapshot.LevelDBSnapshot')
```
ixje: The `load_class_from_path` is intended for dynamic look ups. In this `LevelDBImpl` we know it's always going to be static so we should just use a normal `from neo.Storage.xxx import LevelDBSnapshot` and create an instance. That will also allow IDE's to pick it up if ever refactored/renamed.

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
```diff
@@ -0,0 +1,90 @@
...
+    @contextmanager
+    def getBatch(self):
+
+        with self._lock:
+            _batch = self._db.write_batch()
+            yield _batch
+            _batch.write()
+
+    def getPrefixedDB(self, prefix):
+        PrefixedDB = load_class_from_path('neo.Storage.Implementation.LevelDB.LevelDBSnapshot.LevelDBSnapshot')
```
ixje: same as above. Use a normal class instead of dynamic lookup.

neo/Storage/Implementation/test/test_db_factory.py
```diff
@@ -0,0 +1,124 @@
...
+
+        self._db.write(b'00002.w', b'w')
+        self._db.write(b'00002.x', b'x')
+        self._db.write(b'00002.y', b'y')
+        self._db.write(b'00002.z', b'z')
+
+        from neo.Storage.Interface.DBInterface import DBProperties
+
+        '''
+            Hhas to be converted as leveldb returns a custom iterator object,
```
ixje:
```suggestion
            Has to be converted as leveldb returns a custom iterator object, 
```

neo/Storage/Implementation/test/test_db_factory.py
```diff
@@ -0,0 +1,124 @@
...
+                if isinstance(new_obj, dict):
+                    return new_obj.items()
+                return new_obj
+            return obj
+
+        with self._db.openIter(DBProperties(prefix=b'00001', include_value=True, include_key=False)) as iterator:
+
+            iterator = make_compatible(iterator, list)
+            self.assertEqual(len(iterator), 3)
+            self.assertIsInstance(iterator, list)
```
ixje: I think we should add an explicit assert that the content of the iterator only includes `b'x'`, `b'y'` and `b'z'` which ensures that `include_value=True` + `include_key=False` is respected

neo/Storage/Implementation/test/test_db_factory.py
```diff
@@ -0,0 +1,124 @@
...
+
+        with self._db.openIter(DBProperties(prefix=b'00002', include_value=False, include_key=True)) as iterator:
+            iterator = make_compatible(iterator, list)
+            self.assertEqual(len(iterator), 4)
+            self.assertIsInstance(iterator, list)
```
ixje: same as above. have explicit assert(s) to validate `include_value=False` and `include_key=True` is respected

neo/Storage/Implementation/test/test_db_factory.py
```diff
@@ -0,0 +1,124 @@
...
+
+        with self._db.openIter(DBProperties(prefix=b'00002', include_value=True, include_key=True)) as iterator:
+            iterator = make_compatible(iterator, dict)
+            self.assertEqual(len(iterator), 4)
+            self.assertIsInstance(iterator, abc.ItemsView)
```
ixje: same as above

neo/Storage/Implementation/test/test_db_factory.py
```diff
@@ -0,0 +1,124 @@
...
+        with self._db.openIter(DBProperties(prefix=None, include_value=False, include_key=True)) as iterator:
+            iterator = make_compatible(iterator, list)
+            self.assertEqual(len(iterator), 7)
+            self.assertIsInstance(iterator, list)
+
+        with self._db.openIter(DBProperties(prefix=None, include_value=True, include_key=False)) as iterator:
+            iterator = make_compatible(iterator, list)
+            self.assertEqual(len(iterator), 7)
+            self.assertIsInstance(iterator, list)
+
```
ixje: what happens if we do  `DBProperties` with `include_value=False` and `include_key=False`? I know it doesn't make sense, but we should probably raise an exception as an invalid configuration. This can then be tested here as well

neo/Storage/Interface/DBInterface.py
```diff
@@ -2,11 +2,35 @@
...
+
+    Args:
+        prefix (str, optional): Prefix to search for.
+        include_value (bool, optional): whether to include keys in the returned data
+        include_key (bool, optional): whether to include values in the returned data
+    """
+
+    prefix = None
+    include_value = None
+    include_key = None
```
ixje: Please remove the class variables. I think they should just be instance variables

neo/Storage/Interface/DBInterface.py
```diff
@@ -16,13 +40,15 @@ class DBCollection:
     Changed = []
     Deleted = []
```
ixje: I realise this is old code that already had these class variables. I know this will sound contradicting to what I just said above, but leave them here for now as I have a commit in another branch that should remove all of these and make them instance variables

neo/bin/api_server.py
```diff
@@ -156,6 +156,8 @@ def main():
     # host
     parser.add_argument("--host", action="store", type=str, help="Hostname ( for example 127.0.0.1)", default="0.0.0.0")
 
+    # rollback
+    parser.add_argument("--rollback", action="store", type=int, help="Block id to rollback the chain to")
```
ixje: please remove as feature currently doesn't exist

neo/SmartContract/tests/test_smart_contract.py
```diff
@@ -45,6 +45,8 @@ def test_a_run_sc(self):
 
         block = Helper.AsSerializableWithType(hexdata, 'neo.Core.Block.Block')
 
-        result = self._blockchain.Persist(block)
+        result = False
+        with BlockchainFixtureTestCase.MPPersist():
```
ixje: can you explain the need for this monkey patched persist vs just persisting directly?
if it is about capturing the execution success status we could mock `ExecutionCompleted()`
https://github.com/CityOfZion/neo-python/blob/f08dd6b90a9acc7f2f4b73f2977a316c8899e7af/neo/SmartContract/ApplicationEngine.py#L317 and read the `success` argument passed to the function. Or if the SC emits any events subscribe to the event handler and assert that we get the expected output emitted here
https://github.com/CityOfZion/neo-python/blob/f08dd6b90a9acc7f2f4b73f2977a316c8899e7af/neo/SmartContract/ApplicationEngine.py#L322-L323

My main concern is the code duplication by the `MPersist()` knowing that these are critical areas (and subject to optimisation).

ixje:
This branch doesn't seem to sync. Probably because it can't retrieve data
```
neo> show block 0
[I 190521 10:50:51 Blockchain:674] Could not get block 'NoneType' object is not iterable
Could not locate block 0
neo> show header 0
Could not locate header 0
```

neo/Storage/Implementation/AbstractDBImplementation.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Returns:
+            None
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def cloneDatabase(self, clone_db):
+        """
+        Clones the current database into "clone_db"
```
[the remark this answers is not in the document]
merl111: fixed, renamed method and changed description.

neo/Storage/Implementation/AbstractDBImplementation.py
merl111: Done.
> Now the discussion point would be; should it be read-only or not?
I think it has to be read-only, at least for now, with `LevelDB` as the default database there is no other option. Aside from that, if you have multiple snapshots changing the same data, you can run into serious problems if there is no transaction handling. I think this is something to discuss for the future, maybe keeping it read-only is sufficient?

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
merl111: Done.

neo/Storage/Implementation/LevelDB/LevelDBImpl.py
merl111: Done.

neo/Storage/Implementation/test/test_db_factory.py
merl111: Done.

neo/Storage/Implementation/test/test_db_factory.py
merl111: Done.

neo/Storage/Implementation/test/test_db_factory.py
merl111: Done.

neo/Storage/Implementation/test/test_db_factory.py
merl111: Done.

neo/Storage/Implementation/test/test_db_factory.py
merl111: Catching it now in the init and there is a test for it.

neo/Storage/Interface/DBInterface.py
merl111: Done.

neo/Storage/Interface/DBInterface.py
merl111: Not touching those.

neo/bin/api_server.py
merl111: Sorry, I thought I have removed everything, seems like that slipped through.

neo/SmartContract/tests/test_smart_contract.py
merl111: This is nothing I invented, I just moved it to `BlockchainFixtureTestCase`. In the old code it was located within the `TestLevelDBBlockchain`. I just made it work with the current implementation. I can still change it if you really want to get rid of it.

ixje:
I did a quick test run just now and so far looks good 👍 . 

I'm gonna use your branch to do the audit with as the audit also includes storage comparison. Otherwise I'd have to do that part twice, which doesn't make sense. This does mean it can take a couple of days before it gets merged (fyi)

merl111:
Ok perfect, thank you for the response.

ixje:
500 points

lllwvlvwlll:
https://neoscan.io/transaction/26ef4b7c07219eadd7d74ce0eff4c3a43271921d263db6a400cb075d67c39cb0

PR #934, opened 2019-04-25
[alpha2] move from twisted to asyncio + new network

ixje:
Continuation of #930 

This PR ripped the 3rd party event-driven framework `twisted` out of `neo-python` and replaced it with the standard library `asyncio`. It also introduces new network code.

`neo-python` doesn't have 100% test coverage and this PR touches pretty much every component, therefore I'm looking for help with general testing of this branch. This can be anything from running the RPC server, to general interactions like opening/closing/rebuilding wallets, sending assets, searching contracts. Whatever feature is in there should have no regression and function as expected. If it shows unexpected behaviour (read the note below first) then please drop a comment here, also for any errors you think shouldn't be there/printed .

**Important notes**
1) requires python 3.7
2) This branch incorporates #905 and will give a warning that the database format has changed. Using `np-bootstrap` as suggested will **not** work as the bootstrap files have not yet been updated. 
3) The network code in this branch relies on a new network feature introduced in last weeks release of `neo-cli` -> `v2.10.1`. Therefore when using this in a private-net environment make sure you run it against nodes with `neo-cli` `v2.10.1`. Upgrade `neo-privatenet` using:
```
docker pull cityofzion/neo-privatenet:2.10.1
```
`neo-local` is not yet updated afaik. 

To elaborate why this is important: the `>= 2.9.x <= 2.10.0` series of `neo-cli` started tracking data sent to a peer and limit requesting the same data to only once per session. In order to keep up to date with the best height of a connected node in those series we have 2 options
1) reconnect and extract the height information during the handshake
2) request the last block, read the height from there and cache this block until you can finally process it. When syncing e.g. mainnet this means you can end up caching thousands of blocks in memory because you can't process them until in sync.

I found neither approach acceptable and convinced NEO to add a new network command that allows requesting the best height of a peer without disconnecting. Without going into too much detail, the new network code uses this information to avoid requesting data from a node that doesn't have the data in the first place. The impact for now is that neo-cli nodes < `2.10.1` basically cannot stay in sync. It will be a transition period that we'll have to live with.

PR #948, opened 2019-05-30
revert nodemanager.start.port

jseagrave21:
- addresses https://github.com/CityOfZion/neo-python/pull/934#pullrequestreview-231428364 and https://github.com/CityOfZion/neo-python/pull/934#pullrequestreview-243522758

- now points to `settings.NODE_PORT` instead of `8888`

manual testing

no

- [ ] Did you add any tests?


ixje:
I removed the changelog entry. This is so small (and actually a regression issue that's not out there in any release) that I consider it not big enough to report in the changelog.  Still good catch!

PR #949, opened 2019-05-30
include alias in pretty print

jseagrave21:
- resubmitting https://github.com/ixje/neo-python/pull/18
- addresses https://github.com/CityOfZion/neo-python/pull/934#issuecomment-496112560

- if an alias exists for the address, print the alias
- updated tests

**NOTE:** Right now there is no ability within neo-python to update/delete the alias once it is set. I think this should be a future feature update.

manual testing initially and `make test` after tests were updated

no

- [x] Did you add any tests? (updated)

### DIFF
--- a/neo/Implementations/Wallets/peewee/UserWallet.py
+++ b/neo/Implementations/Wallets/peewee/UserWallet.py
@@ -525,8 +525,17 @@ class UserWallet(Wallet):
                 'tokens': self._get_token_balances(addr_str)
             }})
 
+        aliases = dict()
+        alia = NamedAddress.select()
+        if len(alia):
+            for n in alia:
+                aliases[n.Title] = n.ToString()
+
         # pretty print
         for address, data in addresses.items():
+            for title, addr in aliases.items():
+                if address == addr:
+                    print(f"Alias      : {title}")
             addr_str = address + " (watch only)" if data['watchonly'] else address
             print(f"Address    : {addr_str}")
             if verbose:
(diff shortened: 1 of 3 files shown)


ixje:
I could not resolve the conflict via Github web. For some reason it won't load completely so I can't update the CHANGELOG. You'll have to do it 🤷‍♂

jseagrave21:
@ixje okay, it should be good now

neo/Implementations/Wallets/peewee/UserWallet.py
```diff
@@ -525,8 +525,17 @@ def pretty_print(self, verbose=False):
...
 
+        aliases = dict()
+        alia = NamedAddress.select()
+        if len(alia):
+            for n in alia:
+                aliases[n.Title] = n.ToString()
+
         # pretty print
         for address, data in addresses.items():
+            for title, addr in aliases.items():
```
ixje: Just as a general comment; 

I think we could have used `NamedAddress.get()` to query if an alias exists for the given address/scripthash instead of iterating over all available data. Now it doesn't really matter as we won't have a huge list of aliases. Just something to try and keep in mind when you encounter some performance critical part with larger data sets.

jseagrave21: Thank you for the feedback!

ixje:
I changed the log entry to
```
Include address aliases in ``wallet`` command output
```
because your entry described the internal change that requires "inside" knowledge. This describes the impact for the user

jseagrave21:
> I changed the log entry to
> 
> ```
> Include address aliases in ``wallet`` command output
> ```
> 
> because your entry described the internal change that requires "inside" knowledge. This describes the impact for the user

Thank you! I will try to keep that in mind for future Changelog entries.

ixje:
50 points

lllwvlvwlll:

PR #950, opened 2019-05-30
fix undesired debug statement printing

jseagrave21:
- resubmitting https://github.com/ixje/neo-python/pull/17
- addresses https://github.com/CityOfZion/neo-python/pull/934#issuecomment-496112217

- searched for logzero imports
- replaced with `log_manager` where necessary: Transaction.py & Send.py
- completed manual testing to verify expected result

manual testing

**NOTE:** I don't think I can add individual unittests for this issue because running `make test` displays every level of `logger`.

no

- [ ] Did you add any tests?


ixje:
thanks!

PR #953, opened 2019-05-31
Fix `config maxpeers`

jseagrave21:
**NOTE**: updated re-submission of https://github.com/ixje/neo-python/pull/12

- update for `minpeers`
- update for errors when the new `maxpeers` < connected peers
- update tests

- update to set `minpeers` equal to `maxpeers` if `maxpeers` < `minpeers`
- specifically, when new `maxpeers` < connected peers
  - fixes a RuntimeError caused by calling a coro in another loop
    - see https://github.com/ixje/neo-python/pull/12#pullrequestreview-231447541
  - fixes an error where the same node was disconnected multiple times
    - see https://github.com/ixje/neo-python/pull/12/files/7292761e76ac3a983dbbfc85b41b0b9ef406cbb7..be1307860d274b0db82b3eaf06ff6cb37037350b
- updates tests

manual testing and `make test`

no

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Config.py
+++ b/neo/Prompt/Commands/Config.py
@@ -1,4 +1,3 @@
-import asyncio
 from neo.Network.common import blocking_prompt as prompt
 from neo.logging import log_manager
 from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
@@ -8,6 +7,7 @@ from neo.Prompt.PromptPrinter import prompt_print as print
 from distutils import util
 from neo.Network.nodemanager import NodeManager
 import logging
+from neo.Network.common import wait_for
 
 
 class CommandConfig(CommandBase):
@@ -201,13 +201,9 @@ class CommandConfigMaxpeers(CommandBase):
             connected_count = len(nodemgr.nodes)
             if current_max < connected_count:
                 to_remove = connected_count - c1
-                r_list = []
-                i = 1
                 for _ in range(to_remove):
-                    r_list.append(nodemgr.nodes[-i])
-                    i += 1
-                for node in r_list:
-                    asyncio.run_coroutine_threadsafe(node.disconnect(), nodemgr.loop)  # need to avoid it being labelled as dead/bad
+                    last_connected_node = nodemgr.nodes[-1]
+                    wait_for(last_connected_node.disconnect())  # need to avoid it being labelled as dead/bad
 
             print(f"Maxpeers set to {c1}")
             return c1
--- a/neo/Prompt/Commands/Config.py
+++ b/neo/Prompt/Commands/Config.py
@@ -1,3 +1,4 @@
+import asyncio
 from neo.Network.common import blocking_prompt as prompt
 from neo.logging import log_manager
 from neo.Prompt.CommandBase import CommandBase, CommandDesc, ParameterDesc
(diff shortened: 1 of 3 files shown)


neo/Prompt/Commands/Config.py
```diff
@@ -196,11 +199,15 @@ def execute(self, arguments):
...
-            if c1 < current_max and c1 < connected_count:
+            if current_max < connected_count:
                 to_remove = connected_count - c1
+                r_list = []
+                i = 1
                 for _ in range(to_remove):
-                    last_connected_node = nodemgr.nodes[-1]
-                    wait_for(last_connected_node.disconnect())  # need to avoid it being labelled as dead/bad
+                    r_list.append(nodemgr.nodes[-i])
+                    i += 1
```
ixje: You previously said:
```
This additional list (r_list) is required because iterating with nodemgr.nodes[-1] called the same node again and again since nodemgr.nodes wasn't updated immediately. I added several print statements for debugging and tested manually in privatenet, which yielded this example:

Minpeers set to 1
Connected Count:  3
To Remove:  2
Removed 127.0.0.1:20335
Removed 127.0.0.1:20335
Maxpeers set to 1
The new method builds the list of nodes to remove first, then removes each node on the list.
```
Can you explain what's wrong with that output?

When I only apply this fix:
```
if current_max < connected_count:
```

and insert a print statement as follows
```
                for _ in range(to_remove):
                    last_connected_node = nodemgr.nodes[-1]
                    print(f"disconnecting {last_connected_node.address}")
                    wait_for(last_connected_node.disconnect())
```

then I get
```
neo> show nodes
Connected: 4 of max 5
Peer 0  /NEO:2.8.0/       127.0.0.1:20335 height:     2016
Peer 1  /NEO:2.8.0/       127.0.0.1:20333 height:     2016
Peer 2  /NEO:2.8.0/       127.0.0.1:20334 height:     2016
Peer 3  /NEO:2.8.0/       127.0.0.1:20336 height:     2016

neo> config maxpeers 1
disconnecting 127.0.0.1:20336
disconnecting 127.0.0.1:20334
disconnecting 127.0.0.1:20333
Maxpeers set to 1
neo> show nodes
Connected: 1 of max 1
Peer 0  /NEO:2.8.0/       127.0.0.1:20335 height:     2016

neo>
```

That seems to do exactly what it needs to do. Disconnect the last 3 nodes and only leave the first one that was connected.

jseagrave21: I will check if the behavior has improved in the latest commit, but you can see that the original behavior tried to remove the same node twice so you were stuck with connected peers > maxpeers.

jseagrave21: Okay, I just checked and it seems to be working the way you previously describe. I think maybe `wait_for` has improved?

jseagrave21:
@ixje request an additional review

neo/Prompt/Commands/Config.py
ixje: `wait_for` has not changed. It is just a utility function that allows run asynchronous code in a blocking fashion. Therefore I couldn't see how it would pick the same node twice because it waits for the node to be disconnected before remove the next one. Did you perhaps run it with your `run_coroutine_threadsafe` instead? because that could cause it as that creates tasks.

jseagrave21: Ahh okay, that might have been the issue. I had originally chosen not to use `wait_for` because of the previously mentioned Runtime error, but it seems it may have caused unintended consequences.

ixje:
50 points

lllwvlvwlll:

Issue #960, opened 2019-06-10
Simple Policy Plugin


saltyskip:
### Current behavior
Python nodes currently do not support the Simple Policy plugin on the network. Transactions that do not meet the requirements will be accepted in the node, but stuck in the mempool.

### Expected behavior
Implementation of Simple Policy plugin for neo python

### How to reproduce

Submit large transaction via RPC to python node. If over 1024 bytes and no fee is attached, the transaction should fail. In NEO python it will be accepted into the mempool

### Your environment

RPC

ixje:
Thanks for reporting. 

Is there already an unambiguous description of the feature? It has been discussed last week but there were different interpretations of its working based on the news.

jseagrave21:
@saltyskip @ixje I can confirm this behavior. When we implemented the `SimplePolicyPlugin` in https://github.com/CityOfZion/neo-python/pull/869, we only implemented for relaying transactions, not if a rawtransaction was sent to the node and then verified. I have since implemented similar logic in the `Verify` function and tested.

hal0x2328:
The SimplePolicy implementation in #964 works, but the transaction fee is still calculated incorrectly. I believe this is because the transaction is built is using the network fee as calculated from `test_invoke` that is based on the unsigned transaction size, but the signed transaction size is larger, leading to a failure during `Transaction.Verify`.

I added an extra debugging line locally to `test_invoke` to print the `wallet_tx.Size()` it was basing the fee off of, and then tried to deploy a contract:

```
Calculating required fee for transaction with size: 11299

-------------------------------------------------------------------------------------------------------------------------------------
Test deploy invoke successful
Total operations executed: 11
Results:
[<neo.Core.State.ContractState.ContractState object at 0x1147a12b0>]
Deploy Invoke TX GAS cost: 490.0
Deploy Invoke TX Fee: 0.10275
-------------------------------------------------------------------------------------------------------------------------------------

Enter your password to deploy this contract on the blockchain
[password]> **********************                                                                                                               
[D 190717 09:06:09 Transaction:596] Verifying transaction: b'abf44004b7176d16f62ec404c29384f71011499a38d8fe6609005e1e5d14d194' (size: 11427 bytes)
[D 190717 09:06:09 Transaction:608] The tx size (11427) exceeds the max free tx size (1024).
    A network fee of 0.10403 GAS is required.
[E 190717 09:06:09 mempool:28] Verifying tx result... failed
[...]

PR #964, opened 2019-06-10
Implement SimplePolicyPlugin in Transaction.Verify

jseagrave21:
- addresses https://github.com/CityOfZion/neo-python/issues/960
  - implements `SimplePolicyPlugin` when verifying a tx in the mempool

referenced logic used in #869 and logic I am using here https://github.com/jseagrave21/neo-python/blob/f75cb5dc58bedbfc12ea1a6afbca5332088f8ca6/neo/Core/TX/RawTransaction.py#L525-L541

`make test` and verification manually on my testnet node

no

- [x] Did you add any tests?


jseagrave21:
> The solution does not seem to take the following rule into account ([ref](https://neo.org/blog/details/4148)):
> 
> > Deploy and invoke smart contract with neo-cli v2.10.2 will cost 0.001 GAS as the starting fee.

This was clarified in #dev-general on Discord by @vncoelho that this was only introduced in neo-cli and not a limitation on Seed Nodes or Concensus Nodes. It was discussed further and removed here: https://github.com/neo-project/neo-cli/pull/378

ixje:
I missed that. Thanks for the link!
Looks good then 👍 👍

vncoelho:
The implementation looks precise and the formula correct.

ixje:
thanks for the extra check @vncoelho

neo/Core/TX/Transaction.py
```diff
@@ -594,6 +595,19 @@ def Verify(self, mempool):
...
         logger.debug("Verifying transaction: %s " % self.Hash.ToBytes())
 
+        # SimplePolicyPlugin
+        if self.Size() > self.MAX_TX_SIZE:
+            logger.debug(f'Maximum transaction size exceeded: {self.Size()} > {self.MAX_TX_SIZE}')
+            return False
+        fee = self.NetworkFee()
+        if self.Size() > settings.MAX_FREE_TX_SIZE and not self.Type == b'\x02':  # Claim Transactions are High Priority
+            req_fee = Fixed8.FromDecimal(settings.FEE_PER_EXTRA_BYTE * (self.Size() - settings.MAX_FREE_TX_SIZE))
+            if req_fee < settings.LOW_PRIORITY_THRESHOLD:
```
vncoelho: This was the trick point. Nice job!

ixje:
100 points

lllwvlvwlll:

PR #970, opened 2019-06-23
Storage changes were not committed immediately

merl111:
During audit I found the following problem:
Block `1455702` contains two `InvocationTransactions`.
`0xe5bca1d8c2b115e2d2ecf074812ae9772753600c9ad31b3c6829c5b61175d7f6` passes successfully on `neo-cli` and on `neo-python`.

`0x22dd1a26d8359f0e468f9e94e36de86c90cafaab77779e92c7b3fe88ac1e4d51` fails on `neo-cli` but still passes on `neo-python`.

The problem is, for each block, a `write_batch` instance is created, due to the usage of a `context manager` the batch gets committed once the block is finishe, because the `write_batch` runs out of scope, means, if the first transactions removes something from the database, it's still available for the second transaction, which was the case here. A `Storage_Get` returned a value, which should have been removed already and therefore the second transaction also finished successfully although it should have failed.

Introduced a nested `write_batch` instance for each transaction.

Audit outcome on block `1455702` matches `neo-cli`. 

- [ ] Did you add any tests?


ixje:
Can you check if this PR is still needed now that we have the caching layer ported?

merl111:
Fixed due to the port of the caching layer.

PR #972, opened 2019-06-28
Improve parameter parsing

jseagrave21:
When manually testing contract invocation and `build_run`, I noticed that parameters entered via the command line were not parsed identically to those entered with the `--i` flag. This is because the `--i` flag used the `gather_params` helper function and the regular command line parsing did not.

I split out the parameter verification into a separate function: `verify_params`
- verifies parameters input via command line and --i flag
- ensures parameter input is identical

manual testing

no

- [ ] Did you add any tests?


jseagrave21:
I am closing this for now to continue working on it. @ixje could you confirm whether or not this is an issue worth trying to solve or should the user understand that the `--i` flag should be used whenever inputting specific params (e.g. integer, neo address as a bytearray, etc.)?

PR #973, opened 2019-06-29
Improve parameter parsing [02]

jseagrave21:
**NOTE:** This PR is a continuation of #972 
When manually testing contract invocation and `build_run`, I noticed that parameters entered via the command line were not parsed identically to those entered with the `--i` flag. This is because the `--i` flag used the `gather_params` helper function and the regular command line parsing did not.

I split out the parameter verification into a separate function: `verify_params` and added a `user_entry` flag to limit verification to user entered params.
- verifies parameters input via command line and --i flag
- ensures parameter input is identical
- does not disrupt existing logic which had previously caused many errors in #972

manual testing

No

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/Commands/Invoke.py
+++ b/neo/Prompt/Commands/Invoke.py
@@ -24,6 +24,7 @@ from neo.SmartContract.ApplicationEngine import ApplicationEngine
 from neo.SmartContract import TriggerType
 from neo.SmartContract.StateMachine import StateMachine
 from neo.SmartContract.ContractParameterContext import ContractParametersContext
+from neo.SmartContract.ContractParameter import ContractParameterType
 from neo.SmartContract.Contract import Contract
 from neo.Core.Cryptography.Helper import scripthash_to_address
 from neo.Core.Cryptography.Crypto import Crypto
@@ -170,12 +171,10 @@ def TestInvokeContract(wallet, args, withdrawal_tx=None, from_addr=None,
     contract = BC.GetContract(args[0])
 
     if contract:
-        #
         params = args[1:] if len(args) > 1 else []
 
         params, neo_to_attach, gas_to_attach = PromptUtils.get_asset_attachments(params)
         params, parse_addresses = PromptUtils.get_parse_addresses(params)
-        params.reverse()
 
         if '--i' in params:
             params = []
@@ -185,6 +184,15 @@ def TestInvokeContract(wallet, args, withdrawal_tx=None, from_addr=None,
                     return None, None, None, None, False
                 params.append(param)
             params.reverse()
+        else:
+            i_args = []
+            for index, iarg in enumerate(contract.Code.ParameterList):
+                param, abort = PromptUtils.verify_params(ContractParameterType(iarg), params[index])
+                if abort:
+                    return None, None, None, None, False
+                i_args.append(param)
+            i_args.reverse()
+            params = i_args
 
         sb = ScriptBuilder()
 
@@ -468,17 +476,23 @@ def test_deploy_and_invoke(deploy_script, invoke_args, wallet,
         invoke_args, neo_to_attach, gas_to_attach = PromptUtils.get_asset_attachments(invoke_args)
         invoke_args, no_parse_addresses = PromptUtils.get_parse_addresses(invoke_args)
 
-        invoke_args.reverse()
-
         if '--i' in invoke_args:
             invoke_args = []
             for index, iarg in enumerate(contract_state.Code.ParameterList):
                 param, abort = PromptUtils.gather_param(index, iarg)
                 if abort:
                     return None, [], 0, None
-                else:
-                    invoke_args.append(param)
+                invoke_args.append(param)
             invoke_args.reverse()
+        else:
+            i_args = []
+            for index, iarg in enumerate(contract_state.Code.ParameterList):
+                param, abort = PromptUtils.verify_params(ContractParameterType(iarg), invoke_args[index])
+                if abort:
+                    return None, [], 0, None
+                i_args.append(param)
+            i_args.reverse()
+            invoke_args = i_args
 
         sb = ScriptBuilder()
 
--- a/neo/Prompt/Commands/Invoke.py
+++ b/neo/Prompt/Commands/Invoke.py
@@ -164,8 +164,8 @@ def InvokeWithTokenVerificationScript(wallet, tx, token, fee=Fixed8.Zero(), invo
     return False
 
 
-def TestInvokeContract(wallet, args, withdrawal_tx=None, from_addr=None,
-                       min_fee=DEFAULT_MIN_FEE, invoke_attrs=None, owners=None):
+def TestInvokeContract(wallet, args, withdrawal_tx=None, from_addr=None, min_fee=DEFAULT_MIN_FEE,
+                       invoke_attrs=None, owners=None, user_entry=False):
     BC = GetBlockchain()
 
     contract = BC.GetContract(args[0])
@@ -184,15 +184,21 @@ def TestInvokeContract(wallet, args, withdrawal_tx=None, from_addr=None,
                     return None, None, None, None, False
                 params.append(param)
             params.reverse()
+        elif user_entry:
+            try:
+                i_args = []
+                for index, iarg in enumerate(contract.Code.ParameterList):
+                    param, abort = PromptUtils.verify_params(ContractParameterType(iarg), params[index])
+                    if abort:
+                        return None, None, None, None, False
+                    i_args.append(param)
+                i_args.reverse()
+                params = i_args
+            except IndexError:
+                print(f"Check params. {len(contract.Code.ParameterList)} params specified and only {len(params)} given.")
+                return None, None, None, None, False
         else:
-            i_args = []
-            for index, iarg in enumerate(contract.Code.ParameterList):
-                param, abort = PromptUtils.verify_params(ContractParameterType(iarg), params[index])
-                if abort:
-                    return None, None, None, None, False
-                i_args.append(param)
-            i_args.reverse()
-            params = i_args
+            params.reverse()
 
         sb = ScriptBuilder()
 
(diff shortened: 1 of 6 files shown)


neo/SmartContract/tests/test_gas_costs.py
```diff
@@ -126,8 +126,8 @@ def test_build_contract_5(self):
 
         tx, result, total_ops, engine = BuildAndRun(arguments, wallet, False)
 
-        expected_cost = Fixed8(1046600000)
-        expected_gas = Fixed8.FromDecimal(1.0)
+        expected_cost = Fixed8.FromDecimal(15.466)
+        expected_gas = Fixed8.FromDecimal(6.0)
```
ixje: Do you know why the GAS costs rises with 5 GAS? That really seems like a lot for (assuming) just an input type parsing difference.

jseagrave21: The difference is the original code `pushed` the `big_str` as bytes and the updated code `pushes` it as a `bytearray`. Here are the input values just prior to being `pushed` onto the stack:

Original:
```
b'abababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababab' Length: 2048
b'6b657931' Length: 8
b'7075745f35' Length: 10
```

Updated:
```
bytearray(b'abababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababababab') Length: 2048
b'6b657931' Length: 8
b'7075745f35' Length:10
```

hal0x2328: Using the development branch, I get a result of 1.0 GAS used when invoking with b’abababab…' and 6.0 GAS used when invoking with b’ghghghgh…' so 5 GAS is the difference between parsing that size of an argument as a hex string vs parsing it as an ascii string. The changes in the test_gas_costs.py would seem to be correct if the parser is no longer arbitrarily converting strings to binary just because they look like hexadecimal values.

jseagrave21:
@ixje please note `make test` passes locally. You can see the failure is being caused by the `snapshot` argument which seems to be from #974.

ixje:
note to self; check `ceil` and gas change impact

neo/Prompt/Commands/Invoke.py
```diff
@@ -331,7 +345,6 @@ def test_invoke(script, wallet, outputs, withdrawal_tx=None,
             # print("Used %s Gas " % engine.GasConsumed().ToString())
 
             consumed = engine.GasConsumed() - Fixed8.FromDecimal(10)
-            consumed = consumed.Ceil()
```
ixje: https://docs.neo.org/docs/en-us/tooldev/advanced/tx_execution.html#b02a262b
> Assign Math.max(0, GasConsumed - 10).Ceiling() to InvocationTransaction.Gas as transaction's system fee. Note that system provides 10 Gas for free in each InvocationTransaction.

jseagrave21: @hal0x2328 could you record your findings for GAS costs from Slack here for discussion? It would appear that either, the current implementation of neo-python (with `Ceil` rounding) is correct per the above documentation, or the documentation is outdated and no longer being utilized in neo-cli.

ixje: The documentation seems to match the code: https://github.com/neo-project/neo/blob/56174b8fc2e31bc0f300e4dab1f61bd8cf81ca6e/neo/Network/P2P/Payloads/InvocationTransaction.cs#L44-L46

jseagrave21: Okay. In that case, I will remove this change.

jseagrave21:
@ixje 
BLUF: I think I have addressed all your comments. Please let me know if I am missing something.

Background: In your first comment, since the string 'bla' is not a `bytearray` or a `bytes` object, `eval` returns a `NameError`. I have included this case in testing and exceptions during `eval` or any other returns other than `bytearray` or `bytes` output will result in an Exception to provide a `bytearray` or `bytes` object.

I address both your second and third comments by incorporating exception handling into the use-entry process.

I was a little unsure of what you meant with `Exception('Please provide a list')`. This exception is now incorporated in the the exception handling which provides good feedback to the user.

Please let me know if I am missing something and thank you for your thorough review.

neo/Prompt/Utils.py
```diff
@@ -322,15 +322,22 @@ def verify_params(ptype, param):
...
     elif ptype == ContractParameterType.Array:
-        res = eval(param)
-        if isinstance(res, list):
-            return res, False
+        try:
+            res = eval(param)
+            if isinstance(res, list):
+                return res, False
+        except Exception:
+            pass
```
jseagrave21: This pass is necessary in case there is no exception thrown but `eval` does not evaluate a `list` object.

ixje:
I think you're trying to resolve it in a too narrow context. Normally, it's good to not catch exceptions too 'broad', to make sure you don't have hidden issues. In this case I pointed at the difference of  `PrompUtils.gather_params()` vs your `verify_params` for that specific reason.

The second issue I pointed out is not resolved
```
neo> sc build_run test-sc.py False False False fe10 02 bla []
Saved output to test-sc.avm
Could not execute command: local variable 'ptype' referenced before assignment
  File "/Users/erik/Documents/code/test-issues/seagrave/venv/bin/np-prompt", line 11, in <module>
    load_entry_point('neo-python', 'console_scripts', 'np-prompt')()
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/bin/prompt.py", line 382, in main
    loop.run_forever()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 523, in run_forever
    self._run_once()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 1758, in _run_once
    handle._run()
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/bin/prompt.py", line 234, in run
    traceback.print_stack()
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/Invoke.py", line 477, in test_deploy_and_invoke
    ptype = ContractParameterType(iarg)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 307, in __call__
    return cls.__new__(cls, value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 555, in __new__
    return cls._missing_(value)
  File "/Library/Frameworks/Python.framework/Versions/3.7/lib/python3.7/enum.py", line 568, in _missing_
    raise ValueError("%r is not a valid %s" % (value, cls.__name__))
ValueError: 254 is not a valid ContractParameterType

During handling of the above exception, another exception occurred:

  File "/Users/erik/Documents/code/test-issues/seagrave/neo/bin/prompt.py", line 220, in run
    cmd.execute(arguments)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/SC.py", line 44, in execute
    return self.execute_sub_command(item, arguments[1:])
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/CommandBase.py", line 75, in execute_sub_command
    return self.__sub_commands[id].execute(arguments)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/SC.py", line 80, in execute
    tx, result, total_ops, engine = BuildAndRun(arguments, PromptData.Wallet)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/BuildNRun.py", line 72, in BuildAndRun
    debug_map=debug_map, invoke_attrs=invoke_attrs, owners=owners, enable_debugger=enable_debugger)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/BuildNRun.py", line 96, in DoRun
    enable_debugger=enable_debugger, user_entry=True)
  File "/Users/erik/Documents/code/test-issues/seagrave/neo/Prompt/Commands/Invoke.py", line 488, in test_deploy_and_invoke
    print("Could not parse param as %s : %s " % (ptype, e))
UnboundLocalError: local variable 'ptype' referenced before assignment
```

I also think we can improve the error message:
```
neo> sc build_run test-sc.py False False False 0710 02 bla hoi
Saved output to test-sc.avm
[I 190916 12:19:15 LevelDBImpl:45] Created DB at /Users/erik/.neopython/Chains/debugstorage
Could not parse param as Array : Please provide a list
```
which param is wrong?

PR #974, opened 2019-07-02
Port caching layer for 2.x

ixje:
The ongoing audit showed issues with differences in data accessed from the caching layer. This ports that part over as the "fastest" solution to getting state consistency between neo-cli and neo-python due to a lack of proper specification describing it and the odd way of handling some internals. 

audit passes and surpasses the previous block we got stuck on.

audit progress + make test

too much to mention

- [ ] Did you add any tests?


neo/Core/TX/Transaction.py
```diff
@@ -683,7 +688,7 @@ def Verify(self, mempool):
     #                    return False
     #
 
-    def GetScriptHashesForVerifying(self):
+    def GetScriptHashesForVerifying(self, snapshot):
```
jseagrave21: @ixje adding `snapshot` as a required input here breaks the current `RawTransaction` class because it requires a db instance. I don't think this parameter should be required, since it not required in the case of a `raw_tx` as seen below.

ixje: I'd say `snapshot` is required for correct default NEO behaviour. However, I think we can make an exception here to accommodate for your raw_tx class. Please PR to make it optional.

jseagrave21: @ixje Please see https://github.com/CityOfZion/neo-python/pull/981

PR #985, opened 2019-07-16
Fixed AccountState Sync Issue

brianlenz:

Explicitly defining `lambda` for account syncing to fix `TypeError: 'AccountState' object is not callable` errors.

Added `lambda`, per @ixje.

Tested before and after. Errors happening before; syncing works after.

No.

- [ ] Did you add any tests?


ixje:
thanks 👍

PR #986, opened 2019-07-16
Auto-Delete Bootstrap File

brianlenz:

Auto-delete the downloaded bootstrap file after running `np-bootstrap`.

* Updated the `np-bootstrap` process so that it will delete the downloaded bootstrap file by default when it's done.
* Added a `-k`/`--keep-bootstrap-file` flag to `np-bootstrap` to allow you to keep the bootstrap file in case you need to re-use the file after download.

Tested it with both configurations.

No.

- [ ] Did you add any tests?


ixje:
👍

PR #989, opened 2019-07-18
Fix simple policy implementation for outgoing transactions

jseagrave21:
Addresses #960; specifically, https://github.com/CityOfZion/neo-python/issues/960#issuecomment-512248474

implement SimplePolicy check when relaying any tx
- results in a cancelled tx and a required fee recommendation if SimplePolicy requirements are not met
- addresses https://github.com/CityOfZion/neo-python/issues/960#issuecomment-512248474 by checking each tx after it is signed

`make test` with unittests

Replaced `TXFeeError` with `TXError` to incorporate a check for violating the tx maximum size

- [x] Did you add any tests?

### DIFF
--- a/neo/Core/TX/Transaction.py
+++ b/neo/Core/TX/Transaction.py
@@ -775,6 +775,6 @@ class ContractTransaction(Transaction):
         self.Type = TransactionType.ContractTransaction
 
 
-class TXFeeError(Exception):
-    """Provide user-friendly feedback for transaction fee errors."""
+class TXError(Exception):
+    """Provide user-friendly feedback for transaction errors."""
     pass
--- a/neo/Core/TX/Transaction.py
+++ b/neo/Core/TX/Transaction.py
@@ -773,8 +773,3 @@ class ContractTransaction(Transaction):
         """
         super(ContractTransaction, self).__init__(*args, **kwargs)
         self.Type = TransactionType.ContractTransaction
-
-
-class TXError(Exception):
-    """Provide user-friendly feedback for transaction errors."""
-    pass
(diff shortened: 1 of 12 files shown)


ixje:
note to self; gave slack feedback regarding late notification if funds are insufficient

neo/Core/TX/Transaction.py
```diff
@@ -775,6 +775,6 @@ def __init__(self, *args, **kwargs):
         self.Type = TransactionType.ContractTransaction
 
 
-class TXFeeError(Exception):
-    """Provide user-friendly feedback for transaction fee errors."""
+class TXError(Exception):
+    """Provide user-friendly feedback for transaction errors."""
```
jseagrave21: It seems like `TXError` is now obsolete. Is there any reason to keep it?

ixje:
I had another look at the code because it just didn't feel right to have `relay()` be responsible for policy validation. One of the issues in resolving this in the best way possible is that certain code parts do too much e.g. https://github.com/CityOfZion/neo-python/blob/fe90f62e123d720d4281c79af0598d9df9e776fb/neo/Prompt/Commands/Invoke.py#L48

does not just invoke a contract, it also sign's the TX and this is a problem for reporting the required fees prior to asking if the TX should be send to the network. Instead of resolving the latter now I want to keep that in mind for the NEO3 build.

I moved the policy validation into its own function and out of the `relay()` code because those 2 should not be tied together. In doing so I could also simplify some of the test logic.

neo/Core/TX/Transaction.py
ixje: I guess not.

ixje:
🎉

PR #990, opened 2019-07-20
Fix Prompt Bug with Sending NEP-5 Tokens

brianlenz:

* Fixed a bug where sending NEP-5 tokens via the Prompt CLI did not properly handle the NEP-5 decimal conversion. Thus, previously, you'd have to enter the amount as a "raw" value without a decimal in order for the proper amount of tokens to be sent.

Added the proper conversion of the user-entered decimal amount to the raw value based on the NEP-5 token's `decimals`. This fix is mirrored off of the `approve` command, which was already handling the conversion properly.

Tested it.

No.

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -57,6 +57,7 @@ All notable changes to this project are documented in this file.
 - Updated ``np-bootstrap`` to delete the downloaded bootstrap file by default `#986 <https://github.com/CityOfZion/neo-python/pull/986>`_
 - Fix ``Remove()`` behaviour for ``Map`` and ``Array`` types to be inline with C#
 - Add support for compressed syscalls
+- Fix NEP-5 token send operation in ``np-prompt`` to properly handle token ``decimals``/scale
 
 
 [0.8.4] 2019-02-14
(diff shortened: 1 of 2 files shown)


ixje:
Sounds ok from the description, but I'd like to make sure understand the issue. 

Say we have a Token with 2 decimals. Previously we could not provide `1.1` as the value on the console but had to use `110` (integral (`1`) + fractional (`10`) concatenated). Is that the right understanding?

brianlenz:
Sorry, @ixje, I should have exhibited with an example which makes it very obvious. Yes, your assessment is correct. Here's an example with NRVE, which has 8 decimals:

### Before Fix
```
neo> wallet token send NRVE AXXX AXXX 1
Will transfer 0.00000001 NRVE from AXXX to AXXX
Transfer fee: 0.0001
-------------------------------------------------------------

Enter your password to send to the network
[Password]>
incorrect password
neo> wallet token send NRVE AXXX AXXX 100000000
 
Will transfer 1.00000000 NRVE from AXXX to AXXX
Transfer fee: 0.0001
-------------------------------------------------------------

Enter your password to send to the network
[Password]>
```

### After Fix
```
neo> wallet token send NRVE AXXX AXXX 1

Will transfer 1.00000000 NRVE from AXXX to AXXX
Transfer fee: 0.0001
-------------------------------------------------------------

Enter your password to send to the network
[Password]>

incorrect password
neo> wallet token send NRVE AXXX AXXX 100000000
could not transfer tokens
```

ixje:
thanks 👍

PR #1002, opened 2019-08-12
Fix CDN failure

corollari:


ixje:
Something in the build failed and when switching to `development` branch I got the above conflicts. Because I wanted to do a release today I had no time to wait for the above to be solved. I manually took your changes and applied them. 

I still want to thank you for the PR 👍 

Will close this as it's already taken care of now.

corollari:
Cool, the point of creating a PR was to make it easier to fix so if it gets in the way I fully support doing it another way.

PR #1003, opened 2019-08-12
Fix max recursion for certain stackitem setups

### LINKED ISSUE #1000: Circular map reference causes Python stack overflow
### Current behavior
(Using development branch, probably applies to master as well).

A circular map reference in a contract will hang the neo-python process and eventually overflow the Python stack.

Here's the line that will be executed over and over increasing the stack usage each time it encounters the circular reference:

https://github.com/CityOfZion/neo-python/blob/b519f55e8ae352f770cfa3290c7916beacfee562/neo/VM/ExecutionEngine.py#L105

This probably impacts circular array references as well.

### Expected behavior

Python should exhibit the same behavior as neo-cli, which is to FAULT as soon as the VM stack size exceeds the maximum allowed.

### How to reproduce

Run neo-python on TestNet and wait until you hit transaction `e30e96c2203e1e8e604b17718edd37ae812324e91804e4af352908793e4e2911` in block `2946221`. You'll know when you hit it because the prompt will become completely unresponsive. (neo-cli returns a FAULT on this transaction after only 1.072 GAS used and then continues on to the next transaction).

or,

try this demo contract:
```
def Main(operation):
  mymap = {}
  mymap['key1'] = "circular reference demo"
  mymap['key2'] = "or, how to kill neo-python with a single invocation"
  mymap['key3'] = "more keys helps speed up the process"
  mymap['key4'] = "more keys helps speed up the process"
  mymap['key5'] = "more keys helps speed up the process"
  mymap['key6'] = "more keys helps speed up the process"
  mymap['key7'] = "more keys helps speed up the process"
  mymap['key8'] = "more keys helps speed up the process"
  mymap['key9'] = "more keys helps speed up the process"
  mymap['circular'] = mymap
```
invoke as:
```
sc build_run poc.py False False False 07 05 test
```

### Your environment
OS: OSX Mojave
neo-python v0.8.5-dev
Python 3.7.2

ixje:
close #1000

implemented "reference" to self checking

audit of mentioned testnet block now passes

no

- [ ] Did you add any tests?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -64,6 +64,7 @@ All notable changes to this project are documented in this file.
 - Update ``CheckDynamicInvoke`` to operate on snapshots
 - Fix ``Contract.Destroy`` not always deleting storage
 - Fix ``Equals()`` of ``ByteArray``
+- Fix max recursion depth exception when counting certain VM StackItems that point to themselves
 
 
 [0.8.4] 2019-02-14
--- a/neo/Core/Block.py
+++ b/neo/Core/Block.py
@@ -191,7 +191,12 @@ class Block(BlockBase, InventoryMixin):
             return False
         if other is self:
             return True
+        if type(other) is not Block:
+            return False
         return self.Hash == other.Hash
+
+    def __eq__(self, other):
+        return self.Equals(other)
 
     @staticmethod
     def FromTrimmedData(byts):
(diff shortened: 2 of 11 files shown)


ixje:
Don't merge until extra audit sync. Atm only validated against the last 25K blocks.

ixje:
Syncing from 0 showed that block `723205` failed the audit because of a wrong equality implementation for `CoinReference` that is now called upon. The above commit fixes this and also addresses possible other cases. Running a longer sync

ixje:
Going to call it a pass. Synced from 0 to 2961000 all blocks verified ok against `neo-cli 2.10.2`.

PR #1010, opened 2019-08-26
Fix for build docker image | wrong pip version

volekerb:
RUN pip3 triggers pip 3.6 version but not 3.7 and dependencies could be installed with an error:
```
Step 9/11 : RUN pip3 install -e .
 ---> Running in be35c386c2b8
Obtaining file:///neo-python
neo-python requires Python '>=3.7' but the running Python is 3.6.8
The command '/bin/sh -c pip3 install -e .' returned a non-zero code: 1
```

- [ ] Did you add any tests?


volekerb:
created correct PR #1011

Issue #1009, opened 2019-08-22
Parser breaks when reading --tx-attr argument with spaces


hal0x2328:
### Current behavior
Using `--tx-attrs` with any spaces inside the attribute JSON (even inside quoted strings) results in the following error:

```
[E 190822 14:50:54 Utils:197] Could not parse json from tx attrs: unexpected EOF while parsing (<string>, line 1) 
```

### Expected behavior

To properly parse tx attributes with spaces in them

### How to reproduce

Try the following invocation on TestNet - the `--tx-attr` value is taken directly from the example shown by the `sc invoke help` command in np-prompt:

```
sc invoke 9aff1e08aea2048a26a3d2ddbb3df495b932b1e7 transfer ['AaPoixMs8SAd7
XtXzJqt1t7MsKwjY6KaiL','Ae8JQr7b9QtCxcQMjJSbxAv685UqWF8CY7',100000000000] --tx-a
ttr=[{"usage": 0x90,"data":"my brief description"}]       
```
The spaces both inside and outside of the quotes will stop the parser from reading the rest of the string, resulting in the EOF error. This works:

```
--tx-attr=[{"usage":0x90,"data":"my_brief_description"}] 
```

- but it's not ideal for text remarks to be unable to have spaces in them. I tried escaping the spaces with one or more backslashes to no avail. The only way to get the attribute accepted is to remove all spaces inside and outside of the quotes.

### Your environment

- OS: OSX Mojave
- neo-python version: 0.8.6-dev
- Python version: 3.7.4

jseagrave21:
@hal0x2328 I am having trouble recreating your error. I changed the attribute unittests in test_send_commands.py to:
```
def test_attributes(self):
        nodemgr = NodeManager()
        nodemgr.reset_for_test()
        nodemgr.nodes = [NeoNode(object, object)]

        with patch('neo.Network.node.NeoNode.relay', return_value=self.async_return(True)):
            with patch('neo.Prompt.Commands.Send.prompt', side_effect=[UserWalletTestCase.wallet_1_pass()]):
                PromptData.Wallet = self.GetWallet1(recreate=True)
                args = ['send', 'gas', self.watch_addr_str, '2', '--tx-attr={"usage": 241,"data":"This is a remark"}']

                res = Wallet.CommandWallet().execute(args)
                for item in res.Attributes:
                    print(item.ToJson())
                self.assertTrue(res)
                self.assertEqual(2, len(
[...]

PR #1013, opened 2019-09-01
Fix parsing spaces

jseagrave21:
Addresses https://github.com/CityOfZion/neo-python/issues/1009 (see `test_attribute_spacing`)

The input parser now merges strings until the count of "[" matches "]" or there is an `IndexError` (e.g. parsing a bytearray).

NOTE:
- results in an extra space when parsing nested lists (i.e. ] ]]), which is still processed correctly
- results in one string for an unmatched bracket, which is still acceptable

(see tests for examples)

unittest and manual testing in privatenet

No

- [x] Did you add any tests?

### DIFF
--- a/neo/Prompt/InputParser.py
+++ b/neo/Prompt/InputParser.py
@@ -11,6 +11,7 @@ class InputParser:
             return command_parts[0], command_parts[1:]
         return None, None
 
+
 def merge_items(command_parts):
     s = 0
     f = 0
--- a/neo/Prompt/InputParser.py
+++ b/neo/Prompt/InputParser.py
@@ -7,12 +7,24 @@ class InputParser:
     def parse_input(self, user_input):
         if len(user_input):
             command_parts = self.parser.parseString(user_input)
-            for n, i in enumerate(command_parts):
-                for x in i:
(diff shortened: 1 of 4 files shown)


ixje:
@hal0x2328 can you confirm this fixes your issue?

neo/Prompt/InputParser.py
```diff
@@ -7,12 +7,25 @@ class InputParser:
...
-                    break
+            command_parts = merge_items(command_parts)
             return command_parts[0], command_parts[1:]
         return None, None
+
+
+def merge_items(command_parts):
+    s = 0
+    f = 0
+    for n, i in enumerate(command_parts):
```
ixje: from a codestyle perspective I found it hard to read `n` as index and `i` as element. It is usually `i` as index, and then something for element.
Can you swap them around ? because I was hurting my head while trying to follow the code because of the order. 

Also please use a bit longer names for `s` and `f`, like `start`/`finish`. They make it easier to read. thanks

jseagrave21: @ixje thank you for the recommendations. I will remember them in future PRs.

PR #1037, opened 2020-08-14
Fix to the syntax warnings caused by comparing literals

ilkericyuz:
There are 5 syntax warnings in 2 files in the current code. Warnings appear when neo-python is installed as a library in a separate project.

`SyntaxWarning: "is not" with a literal. Did you mean "!="?`

Replaced them with `!=` which is the proper way to compare literals.

It doesn't raise syntax warning anymore. 

no

- [ ] Did you add any tests?  n/a


ixje:
Thanks!

